PR #2, opened 2016-12-08
Check for flapping services during WaitForDeployment

zpatrick:


tlake:
Looks good to me! :+1: 🚢

Issue #11, opened 2016-12-13
l0-setup apply overwrites user-provided dockercfg file with empty creds 


jonathanstallings:
l0-setup Version: `v0.8.4_darwin`

Description: `l0-setup apply` overwrites user-provided `dockercfg` file with empty creds `{}`.

Steps to Reproduce: 
1) Run `l0-setup apply <prefix>` without an existing `layer0/instances/<prefix>` directory.
2) Enter the requested AWS credentials.
3) When the prompt stops and advises you to add a valid `dockercfg` file to the `layer0/instances/<prefix>`, do so.
4) Enter `yes` to continue.
5) The `layer0/instances/<prefix>/dockercfg` file will be replaced with contents: `{}`.

Relevant docs:
http://docs.xfra.ims.io/setup/install/#optional-part-3-configure-the-dockercfg-file-to-use-a-private-docker-registry

diemonster:
This is due to how we check for an existing `dockercfg` file: https://github.com/quintilesims/layer0/blob/master/setup/context/commands.go#L408 . If the file is placed after the prompt to create one, it will not be rechecked and will instead be overwritten. 

I'd rather not add _another_ prompt to this section if the file is placed. Will mull this workflow over  a bit and get a patch out shortly.

jonathanstallings:
Could we, perhaps, change to the `l0-setup apply` command to accept a path to a valid `dockercfg` file, which would then be copied to the instance directory after validation?

Then the private docker registry user would only have to run `l0-setup apply` once.

In the other case, if users pass an empty path to dockercfg prompt, l0-setup will create an empty dockercfg file, as is the current behavior.

diemonster:
@jonathanstallings I could get behind that, especially using an optional flag for specifying the `dockercfg` path.

i.e. 
`l0-setup apply --dockercfg /path/filename ...`

@zpatrick any preference here?

PR #16, opened 2016-12-15
l0-setup dockercfg flag update

jonathanstallings:
Addresses Issue #11.

Principal changes include:
- Add `--dockercfg` flag on `l0-setup apply` allowing passing a path to valid`dockercfg` file. This file will be copied to the instance directory.
- Reword user prompt displayed when no `dockercfg` file supplied by user.
- Create default `dockercfg` file before showing user prompt.
- Add simple JSON validation of `dockercfg` file.
- Begin refactoring for easier testability.
- Add simple unit test for `validateDockercfg`.

Further work would be needed to add more test coverage.


setup/context/commands.go
```diff
@@ -5,14 +5,15 @@ import (
...
-	"github.com/quintilesims/layer0/common/aws/provider"
-	"github.com/quintilesims/layer0/common/aws/s3"
-	"github.com/quintilesims/layer0/common/config"
 	"io/ioutil"
 	"net/http"
 	"os"
 	"strings"
 	"time"
+
+	"github.com/quintilesims/layer0/common/aws/provider"
```
diemonster: [nitpick] You're probably using `goimports` rather than `gofmt`, which creates this line break between stdlib packages and 3rd party libraries. Our styleguide thus far has been to use `gofmt`, which does not add this line break.

jonathanstallings: Right. I remember you mentioned that. Could you send me a link to the styleguide?

jonathanstallings: Changed to using `gofmt` instead of `goimports` from now on. Thanks for the reminder!

setup/context/commands_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"fmt"
+	"os"
+	"testing"
+)
+
+type MockFileIO struct {
+	data map[string][]byte
+}
+
+func (mock MockFileIO) ReadFile(path string) ([]byte, error) {
```
diemonster: I would recommend making your method receivers into pointers. This is a pretty good article on why that can be important, aside from consistency with the majority of our other methods: https://nathanleclaire.com/blog/2014/08/09/dont-get-bitten-by-pointer-vs-non-pointer-method-receivers-in-golang/

jonathanstallings: Thanks! I'll check that out.

setup/context/commands.go
```diff
@@ -393,35 +430,60 @@ func Endpoint(c *Context, syntax string, insecure, dev, quiet bool) error {
...
+func checkDockercfg(c *Context, force bool, io FileIO) error {
 	path := fmt.Sprintf("%s/dockercfg", c.InstanceDir)
 
-	// if file doesn't exist
-	if _, err := os.Stat(path); err != nil {
+	// if file does not exist
+	if _, err := io.Stat(path); err != nil {
+
+		// write empty dockercfg file
+		if err := io.WriteFile(path, []byte("{}"), 0666); err != nil {
```
diemonster: [nit] the file permission here is `0666` --- I'd recommend changing it to `0660`, not sure why `everyone` would need to have read/write access to an auth file, esp on a multi-user system (this was like this prior to your PR, btw, missed it the first time!)

setup/context/commands.go
diemonster: There is no styleguide aside from the code b/c `gofmt` takes care of guiding our style 🎩

setup/context/commands.go
```diff
@@ -63,8 +87,21 @@ func Apply(c *Context, force bool) error {
 		}
 	}
 
-	// write empty dockercfg file if it doesn't exist
-	if err := checkDockercfg(c, force); err != nil {
+	var dockercfg string
+	dockercfg_flag, ok := c.Flags["dockercfg"]
```
zpatrick: `dockercfg` should not be a context flag, it should be passed in as a param to `Apply` just like `force` flag is (will go over more in main)

setup/context/commands.go
```diff
@@ -63,8 +87,21 @@ func Apply(c *Context, force bool) error {
...
+
+	if ok && dockercfg_flag != nil && *dockercfg_flag != "" {
+		fmt.Printf("Detected dockercfg: `%s`\n", *dockercfg_flag)
+		dockercfg = *dockercfg_flag
```
zpatrick: go uses camel case, not underscore: `dockercfgFlag`

setup/context/commands.go
```diff
@@ -393,35 +430,60 @@ func Endpoint(c *Context, syntax string, insecure, dev, quiet bool) error {
...
 			}
 		}
 
-		// write empty dockercfg file
-		if err := ioutil.WriteFile(path, []byte("{}"), 0666); err != nil {
-			return fmt.Errorf("Failed to write 'dockercfg': %v", err)
-		}
+	}
+
+	if err := validateDockercfg(path, io); err != nil {
```
zpatrick: [nit] shorten this block to: `return validateDockercfg(path, io)`

setup/context/commands_test.go
```diff
@@ -0,0 +1,69 @@
+package context
```
zpatrick: I'm pretty hesitant to give the :+1:  on these unit tests and introducing the whole `FileIO` interface for a couple of reasons:

* This tool isn't setup to be unit tested. In my opinion, adding more interfaces, classes, and mocks on top of the current architecture isn't a step in the right direction - it just adds more complexity and maintenance costs to an already complex tool. If we want unit tests, we should do it right or not at all. 
* Creating your own mock class (`MockIO`) is a pit we fell into when first starting go. The "better" way of generating mocks is is via a tool called [gomock](github.com/golang/mock). That being said, regardless of how the mock class is generated/updated, the larger issue I have with it is the additional maintenance cost. 
* This is echoing the first point a bit - but these unit tests don't really _add_ anything. They essentially just check if `json.Marshal` works. If having unit tests don't give you confidence in your code - why have them at all? Don't get me wrong - I'm not trying to say that Jonathan writes bad unit tests - what I **am** saying is that the current architecture is so bad that it's damn near impossible for anyone to write a useful unit test. 

TLDR: I think it would be better to drop the unit tests and associated interfaces/mocks/etc from this PR in favor of re-architecting this tool to be more unit testable.

setup/main.go
```diff
@@ -74,6 +75,10 @@ func main() {
 
 	app.Command("apply", "Create/Update a Layer0", func(cmd *cli.Cmd) {
```
zpatrick: `dockercfg` is not a flag for the `context` object, but for the `apply` command. This should look like:
```
app.Command("apply", "Create/Update a Layer0", func(cmd *cli.Cmd) {
		flags := loadFlags(cmd, []string{"access_key", "secret_key", "region"})
		force := cmd.BoolOpt("force", false, "Set this flag to skip prompting on a missing dockercfg file")
		dockercfg := cmd.StringOpt("dockercfg", "", "Path to valid dockercfg file")

		vpc := cmd.StringOpt("vpc", "", "VPC id to target.  Will create new VPC if blank.")
		flags["vpc_id"] = vpc

		command := func(c *context.Context) error {
			return context.Apply(c, *force, *dockercfg)
		}

		flagsCommand(cmd, command, flags)
	})
```

setup/context/commands_test.go
diemonster: I had a feeling you would say as much about `gomock`. I think @jonathanstallings work here is valuable in learning _why_ we use Gomock, and in that sense, this was not a pointless exercise. As for the value of the tests themselves and the "all or nothing" approach, I'm inclined to agree with you.

Devil's advocate argument re: `MockIO` is that for new developers coming into this project, the need for Gomock is not entirely obvious. I'd like us to keep that in mind when people attempt to add tests in the future (if they even try to add tests!)

setup/context/commands.go
jonathanstallings: Of course. Should I change the case on `vpc_flag` etc. from which I modeled this from as well?
https://github.com/jonathanstallings/layer0/blob/c8d151db6033d76ac354821be1940df9cab087e3/setup/context/commands.go#L60

setup/context/commands.go
jonathanstallings: Yup!

setup/context/commands.go
jonathanstallings: Got it. Thanks, zpatrick!

setup/main.go
jonathanstallings: Thanks! I have made this change.

setup/context/commands_test.go
jonathanstallings: Thanks, both. I appreciate the discussion here. Part of why I opted to create the mock class directly was because of my unfamiliarity with gomock, for sure. It also served as a bit of a learning exercise for working with Go. I will certainly look into using the best tools for the project.

I have to differ with the percieved value of the unit tests. Although it is true that the _current_ implementation is essentially just wrapping `json.Marshal`, that could certainly change; e.g., more robust validation against certain schemas could be added or it could be tweaked another way. The unit tests are there as sentinels to ensure that behavior works as expected when things are changed.

Regardless, I agree with the larger points expressed above. The unit tests have been dropped, and the mock class has been removed.

jonathanstallings:
Thank you, @diemonster and @zpatrick for your thorough and useful feedback. I understand that the `l0-setup` tool may be re-architected, and thus this PR may not have relevance long term.

In any case, it was a great way for me to learn a bit more about Go and layer0, which is a project I would like to support in the future.

Issue #4, opened 2016-12-08
Remove Certificates as entities


zpatrick:
* Remove certificates as entities, and remove all code relating to certificate management. 
* In `loadbalancer create`,  take the `--certificate` param as the name of an existing aws certificate.

PR #20, opened 2017-01-06
Data Refactor & Certificate Removal

zpatrick:
This contains the following changes:
* Refactors the data layer to be more readable and maintainable
* Updates logic layer to use the new data layer
* Update logic layer tests to be cleaner and more readable
* Removes certificates as entities
* Changes loadbalancer commands in cli to take existing certificate names

This  fixes https://github.com/quintilesims/layer0/issues/4


api/logic/load_balancer_logic.go
```diff
@@ -123,70 +123,51 @@ func (this *L0LoadBalancerLogic) UpdateLoadBalancer(loadBalancerID string, ports
...
-	}
-
-	tags, err := this.TagData.GetTags(filter)
+	tags, err := this.TagStore.SelectByQuery("load_balancer", "")
 	if err != nil {
 		return false, err
 	}
 
-	return len(tags) > 0, nil
+	ewts := tags.GroupByEntity().
```
diemonster: Nice, this is a lot more readable. What is the meaning of the `ewts` name, though?

diemonster:
For consistency's sake, are you planning on changing all existing method receiver variables from `this` to the Go style?

diemonster:
Regarding certificates, I'm fine with removing them but I think we need to discuss what the alternate path is going to be for people that use them. We never had the best support for certificates obviously, but they did technically sort of work ;)

api/handlers/tag_handler.go
```diff
@@ -27,69 +28,143 @@ func (this *TagHandler) Routes() *restful.WebService {
...
+	// filter by entityID and/or entityType
+	// if neither are specified, this acts as a SelectAll
+	tags, err := t.TagStore.SelectByQuery(entityType, entityID)
 	if err != nil {
 		ReturnError(response, err)
 		return
 	}
 
-	response.WriteAsJson(result)
+	// filter the non-special params the by tag.Name and tag.Value (e.g. environment_id, version)
```
diemonster: [nit] typo in comment

api/handlers/tag_handler.go
```diff
@@ -27,69 +28,143 @@ func (this *TagHandler) Routes() *restful.WebService {
...
 
-	result, err := this.TagData.GetTags(keys)
+	var entityType string
+	var entityID string
+	var fuzz string
+	var latestVersion bool
+
+	// break out special filter params so we don't filter
+	// them by tag.Key and tag.Value
+	if val, ok := params["type"]; ok {
```
diemonster: This logic makes a lot more sense to me. Consider a helper function for removing the special params?

api/handlers/tag_handler_test.go
```diff
@@ -1,98 +1,363 @@
...
+	handler := NewTagHandler(store)
+
+	cases := []HandlerTestCase{
+		{
+			Name: "2",
+			Request: &TestRequest{
+				Query: "version=2",
+			},
+			Run: func(r *testutils.Reporter, _ interface{}, req *restful.Request, resp *restful.Response, read Readf) {
+				handler.FindTags(req, resp)
```
zpatrick: [nit] The "call method and marshal" workflow could be made into a single helper function since all handler methods are `func(req, resp)`:
```
Run: func(r *testutils.Reporter, target interface{}, calllAndMarshal Readf) {
    var tags []models.EntityWithTags
    callAndMarshal(handler.FindTags, &tags)

    r.AssertEqual(len(tags), 2)
    ...
}
```

zpatrick:
> For consistency's sake, are you planning on changing all existing method receiver variables from this to the Go style?

Yeah, but not all at once. Basically just do it whenever working on a particular class.

zpatrick:
> Regarding certificates, I'm fine with removing them but I think we need to discuss what the alternate path is going to be for people that use them.

Those changes are here. You can use existing IAM certificates in `loadbalancer create | addport`. 
People still have a "default` certificate with the self-signed api cert instead of the `imshealthlabs` one we used to have.

zpatrick:
@diemonster I've added updates based on your feedback

common/db/tag_store/mysql.go
```diff
@@ -0,0 +1,175 @@
...
+	return m.query("SELECT * FROM tags")
+}
+
+func (m *MysqlTagStore) SelectByQuery(entityType, entityID string) (models.Tags, error) {
+	if entityType == "" && entityID == "" {
+		return m.SelectAll()
+	}
+
+	clauses := []string{}
+	args := []interface{}{}
```
diemonster: is `interface` necessary here?

common/db/tag_store/mysql_test.go
```diff
@@ -0,0 +1,208 @@
...
+		t.Fatal(err)
+	}
+
+	testutils.AssertEqual(t, len(tags), len(expected))
+	for _, tag := range tags {
+		testutils.AssertInSlice(t, tag, expected)
+	}
+}
+
+func TestMysqlTagStoreInsert(t *testing.T) {
```
diemonster: these tests are nice

common/models/entities_with_tags.go
```diff
@@ -0,0 +1,43 @@
+package models
+
+type EntitiesWithTags []*EntityWithTags
+
+type ewtFilter func(e EntityWithTags) bool
+
+func (e EntitiesWithTags) RemoveIf(f ewtFilter) EntitiesWithTags {
```
diemonster: [dumb question?] why is there logic in this package?

common/db/tag_store/mysql.go
zpatrick: it could be a slice of strings, but since it gets passed to `m.query(query, args...)` (which takes `...interface{}`), it saves us the step of converting `[]string` to `[]interface{}`

common/models/entities_with_tags.go
zpatrick: To me, it just seems like the most logical place to put it. Also, I don't want to have to import 2 packages whenever I'd like to operate on tags.

diemonster: Fair enough. My thought was more around if someone external was to consume this package individually, would the logic contained here be helpful? If it's purely specific to our implementation surrounding this package, I think we should move the logic into `common`.

common/db/tag_store/mysql.go
diemonster: ok makes sense

common/db/job_store/mysql.go
```diff
@@ -0,0 +1,221 @@
...
+	jobs, err := m.query("SELECT * FROM jobs where job_id=?", id)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(jobs) == 0 {
+		return nil, fmt.Errorf("Job with id '%s' not found", id)
+	}
+
+	return jobs[0], nil
```
diemonster: how do we know `[0]` is the `ID` we want? Do we need to do any additional error checking if `len(jobs) >= 2`?

common/db/job_store/mysql_test.go
```diff
@@ -0,0 +1,175 @@
...
+			JobType:     int64(types.DeleteLoadBalancerJob),
+			Request:     "request4",
+			TimeCreated: time.Now(),
+			LastUpdated: time.Now(),
+			Meta:        map[string]string{"k4": "v4"},
+		},
+	}
+}
+
+func NewTestJobStore(t *testing.T) *MysqlJobStore {
```
diemonster: 🔥 🔥 🔥

common/startup/startup.go
```diff
@@ -86,74 +80,44 @@ func GetBackend(credProvider provider.CredProvider, region, mysqlConnection, adm
...
 
-func getTagData(mysqlConnection string) (data.TagData, error) {
-	var getTagDataStore func() (data.TagDataStore, error)
-	if mysqlConnection != "" {
-		os.Setenv("LAYER0_MYSQL_CONNECTION", mysqlConnection)
-		getTagDataStore = func() (data.TagDataStore, error) { return data.NewTagMySQLDataStore() }
-	} else {
-		getTagDataStore = func() (data.TagDataStore, error) { return data.NewTagSQLiteDataStore() }
-	}
+func getTagStore() (tag_store.TagStore, error) {
```
diemonster: [nit] this is calling a constructor rather than just `get`ting the `TagStore`. Is `getTagStore` still an appropriate name given what it's actually doing?

tests/test.sh
```diff
@@ -1,9 +1,11 @@
...
 
-for test in */ ; do
+dirs=(admin deploy environment job load_balancer service task)
+
+for dir in "${dirs[@]}"; do
     echo
-    echo "Testing ${test::-1}"
-    echo "=========================="
-    bats "$test/test.bats" || :
+    echo "Testing $dir"
```
diemonster: 👍

common/testutils/db.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"flag"
+	"fmt"
+	"github.com/quintilesims/layer0/common/db"
+)
+
+var (
+	username = flag.String("username", "layer0", "username for the test db")
+	password = flag.String("password", "nohaxplz", "password for the test db")
```
diemonster: I'd suggest changing the user/pass combination to something more generic

common/startup/startup.go
zpatrick: it's building + initializing the `tagStore` ... maybe something like `getNewTagStore()`?

common/db/job_store/mysql.go
zpatrick: This runs on the assumption that `job_id` is the primary key for the `jobs` table. 
Given that, the number of results can only ever be 0 or 1.

common/models/entities_with_tags.go
zpatrick: >  if someone external was to consume this package individually, would the logic contained here be helpful?

Yes, definitely. It's similar to adding python-like list semantics to a go slice, e.g. you can do things like `tags.First()` .

I get why being rigid on the rule that `/models` should only ever contain just structs has its merits. And if we should hardline on that, then ok let's move these to common 

But to put in my final thoughts -  I had the same feelings initially that I should put the logic stuff in `/common`. However, when you try to use it practice, it gets really annoying needing to to do this back-and-forth conversion between the logic and model packages. Functions end up looking like:
```
func getNameTag(type, id string) (*models.Tag, error) {
    tags, err := tagStore.Select(type, id)
    if err != nil { ... }

    // convert []models.Tag to common.Tags
    tagLogic := common.Tags(tags)

     // find by name
    nameTag := tagLogic.WithKey("name").First()

    // convert common.Tag back to models.Tag
    return models.Tag(nameTag), nil
}
```

vs. 

```
func getNameTag(type, id string) (*models.Tag, error) {
    tags, err := tagStore.Select(type, id)
    if err != nil { ... }

    return tags.WithKey("name").First(), nil
}
```

Also, I feel like our apprehension for putting any sort of logic into `/models` may be a result of us using the old `layer0-models` repository with swaggergen.

api/handlers/job_handler.go
```diff
@@ -30,29 +30,29 @@ func (this *JobHandler) Routes() *restful.WebService {
 
 	service.Route(service.GET("/").
 		Filter(basicAuthenticate).
-		To(this.ListJobs).
+		To(this.SelectAll).
```
tlake: 👍

api/handlers/job_handler.go
```diff
@@ -30,29 +30,29 @@ func (this *JobHandler) Routes() *restful.WebService {
...
 		Doc("Stop and remove a job").
 		Param(id).
 		Returns(http.StatusNoContent, "Deleted", nil))
 
 	return service
 }
 
-func (this *JobHandler) ListJobs(request *restful.Request, response *restful.Response) {
-	jobs, err := this.JobLogic.ListJobs()
+func (this *JobHandler) SelectAll(request *restful.Request, response *restful.Response) {
```
tlake: [nit] Since we're touching JobHandler methods, would we want to update `this` to be more Golangish in this PR?

api/logic/environment_logic.go
```diff
@@ -113,21 +109,13 @@ func (this *L0EnvironmentLogic) UpdateEnvironment(environmentID string, minClust
...
 		return err
 	}
 
-	for _, tag := range rangeTags(tags) {
-		if tag.Key == "name" {
-			model.EnvironmentName = tag.Value
-			break
-		}
+	if tag := tags.WithKey("name").First(); tag != nil {
+		model.EnvironmentName = tag.Value
```
tlake: These refacs are p nice 👍

common/models/entities_with_tags.go
diemonster: If this was a `models` package that was intended to be used by other language consumers, then I think being hardline would make more sense. But it isn't, and the benefits of the current semantics for `getNameTag` outweigh rigidity (i think)

api/handlers/job_handler.go
zpatrick: yup, will change

common/startup/startup.go
zpatrick: changed to `getNewTagStore`

common/testutils/db.go
zpatrick: What are you thinking? I don't want to use something like `root` since that comes baked into mysql.

api/handlers/job_handler.go
zpatrick: updated

common/testutils/db.go
diemonster: u: `layer0` p: `password`

zpatrick:
@diemonster I believe all blocking changes have been fixed

common/db/tag_store/mysql_test.go
```diff
@@ -0,0 +1,208 @@
...
+			Expected:   testTags[6:8],
+		},
+		{
+			EntityType: "deploy",
+			EntityID:   "d1",
+			Expected:   testTags[10:12],
+		},
+	}
+
+	for i, c := range cases {
```
tlake: Oh man, I love this parametrization. :+1::+1::+1:

common/models/entities_with_tags.go
tlake: I think I'm with zp on this one - it feels similar to some backend design practices I've picked up in Python. Example:
```
class ExampleModel(db.Model):
    id = db.Column(db.Integer, primary_key=True)
    name = db.Column(db.String(256))
    created = db.Column(db.DateTime, nullable=False, default=datetime.utcnow())

    @classmethod
    def new(cls, name, session=None):
        if session is None:
            session = db.session
        instance = cls(name=name)
        session.add(instance)
        session.commit()
        return instance

    @classmethod
    def get_by_name(cls, name, session=None):
        if session is None:
            session = db.session
        try:
            instance = session.query(cls).filter_by(string_id=string_id).one()
        except NoResultFound:
            instance = None
        return instance
```

This way, each model know things about how it should go about creating/deleting/querying itself. If you've got multiple modules all accessing the same model, it prevents you from needing to define that logic in multiple places. Of course, you folks already get around that by using the `common/` dir, but it just feels cleaner to me for model logic to be with the models, instead of divorced and defined somewhere completely different.

</twocents>

Issue #27, opened 2017-01-09
l0 deploy list doesn't show deploys beyond version 9


tribaljack:
From Jenny:

`l0 deploy list` shows only up to version 9.

Repro:
Create more than 9 iterations of a Deploy.
See deploy list:

```
$ ./l0 deploy create dockerrun-hound.aws.json hound
DEPLOY ID  DEPLOY NAME  VERSION  
hound.12   hound        12       

$ ./l0 deploy list
DEPLOY ID  DEPLOY NAME  VERSION  
hound.9    hound        9        
job.9      job          9      
```

PR #28, opened 2017-01-11
Convert deploy versions to ints when comparing

zpatrick:
Fixes https://github.com/quintilesims/layer0/issues/27


api/handlers/tag_handler_test.go
```diff
@@ -173,7 +182,51 @@ func TestFindTags_byVersion(t *testing.T) {
...
-				r.AssertEqual(tags[0].EntityID, "d2")
+				r.AssertEqual(tags[0].EntityID, "d.2")
+			},
+		},
+		{
+			Name: "7",
+			Request: &TestRequest{
+				Query: "version=7",
+			},
+			Run: func(r *testutils.Reporter, _ interface{}, req *restful.Request, resp *restful.Response, read Readf) {
```
tlake: [nit] Maybe these tests could be further parametrized? Programmatically generate both the `[]*models.Tag{}`s on line 155 and the `[]HandlerTestCase{}`s here? Excepting the last test case here where we're making sure that 84 doesn't exist, only a few numbers in some strings are changing between tests.

common/testutils/reporter.go
```diff
@@ -18,49 +18,49 @@ func NewReporter(t *testing.T, name string) *Reporter {
 	}
 }
 
-func (this *Reporter) Error(err error) {
-	namedFormat := this.formatName(err.Error())
-	this.T.Errorf(namedFormat)
+func (r *Reporter) Error(err error) {
```
tlake: Removing more `this`s be like:

api/handlers/tag_handler_test.go
zpatrick: :+1: good call

zpatrick: I've added the `callAndRead` function to remove some of the duplicate code in these cases.

tlake:
changes look good

PR #35, opened 2017-01-12
Add summary models and improve performance for deploy list

zpatrick:
This adds the new <Entity<Summary models for bulk list calls. It also contains the updated Deploy List call that returns the new DeploySummary object. I also noticed our Deploy Get call doesn't actually fill in the Dockerrun field, so I added a function to populate that field.


common/models/load_balancer_summary.go
```diff
@@ -0,0 +1,8 @@
+package models
+
+type LoadBalancerSummary struct {
```
diemonster: I realize these structs are a subset of the structs they're based on, but overall, what are we gaining by having them?

diemonster: I guess, to answer my own question, there are benefits in clarity throughout the code. That's one good reason. Another would be not having to pass around non-critical data relating to List operations, although I wonder how much performance / memory gains we really get from that...

api/backend/ecs/deploy_manager_test.go
```diff
@@ -29,37 +29,49 @@ func (this *MockECSDeployManager) Deploy() *ECSDeployManager {
...
+
 func TestGetDeploy(t *testing.T) {
 	testCases := []testutils.TestCase{
 		testutils.TestCase{
 			Name: "Should call ecs.DescribeTaskDefinition with proper params",
 			Setup: func(reporter *testutils.Reporter, ctrl *gomock.Controller) interface{} {
 				mockDeploy := NewMockECSDeployManager(ctrl)
 
-				deployID := id.L0DeployID("some_id").ECSDeployID()
+				deployID := id.L0DeployID("some_id.1").ECSDeployID()
```
diemonster: [ultra nit] the period strikes me as a little odd

api/backend/ecs/deploy_manager.go
```diff
@@ -132,3 +138,31 @@ func marshalDeploy(body []byte) (*deploy, error) {
 
 	return &d, nil
 }
+
+func extractDockerrun(taskDef *ecs.TaskDefinition) ([]byte, error) {
```
diemonster: 👍

cli/command/deploy_command.go
```diff
@@ -115,8 +115,17 @@ func (d *DeployCommand) printDeploys(deploys []*models.Deploy) error {
...
 
-func filterDeploys(deploys []*models.Deploy) []*models.Deploy {
-	catalog := map[string]*models.Deploy{}
+func (d *DeployCommand) printDeploySummaries(deploys []*models.DeploySummary) error {
+	entities := []entity.Entity{}
+	for _, s := range deploys {
+		entities = append(entities, entity.NewDeploySummary(s))
+	}
+
+	return d.Printer.PrintEntities(entities)
```
diemonster: would it be overkill to add pagination for `all` returns?

common/models/load_balancer_summary.go
zpatrick: Main benefits in my eyes are:
* API contracts - keeping our current models for `GET /<entity>` and saying "oh, by the way, fields x, y, and z will always be blank for this call" is just plain lazy and confusing.
* Performance - this will vary by entity, but removing fields like `Dockerrun` for Deploys and `Deployments` for Services will speed things up significantly.

api/backend/ecs/deploy_manager_test.go
zpatrick: thats the layer0 id format for deploys: `<deploy name>.<version>`

cli/command/deploy_command.go
zpatrick: I think so. We don't have any consumers of the API (cli, terraform) that display things in pages.

PR #36, opened 2017-01-12
change dir structure for smoketests

zpatrick:
* Places `.bats` test files into same directory
* Adds a makefile instead of `test.sh`
* Adds skeleton dir for future integration tests


tests/smoke/Makefile
```diff
@@ -0,0 +1,10 @@
+test:
```
diemonster: lol why didn't we do this from the beginning?

PR #38, opened 2017-01-16
Fix JobList/Get, remove entity layer in cli

zpatrick:
* The `GetJob` and `ListJobs` commands were accidentally replaced by `SelectAll` and `SelectByID`. Those are reverted back to `GetJob` and `ListJobs`, respectively. 

* This removes the `entity` and `table` packages from the CLI.
These packages were both created to provide a generic way to print an entity. Given that this required 2 levels of abstraction, and that the `JSONPrinter` bypassed the generic type, we clearly  implemented this incorrectly. 

If we simply remove the generic printing, the code actually becomes more readable on the whole. For example, `command.ListServices` looks like:
```go
services, _ := s.Client.ListServices()
return s.Printer.PrintServices(services)
```

instead of what we have now:
```go
services, _ := s.Client.ListServices()
// convert []models.Service to []entity.Entity
var entities []entity.Entity
for _, s := range services{
    entities = append(entities, entity.NewService(s))
}
return s.Printer.PrintEntities(entities)
```

The text printer becomes longer, but doesn't hide it's implementation between layers of abstraction:
```go
func (t *TextPrinter) PrintEnvironmentSummaries(environments ...*models.EnvironmentSummary) error {
        rows := []string{"ENVIRONMENT ID | ENVIRONMENT NAME"}
        for _, e := range environments {
                row := fmt.Sprintf("%s | %s", e.EnvironmentID, e.EnvironmentName)
                rows = append(rows, row)
        }

        fmt.Println(columnize.SimpleFormat(rows))
        return nil
```

Overall, I think this makes the CLI code more readable and much less redundant without the `entity` package


zpatrick:
I'd like to get this merged before continuing the `List<Entity>` updates in order to reduce redundant code.

cli/command/command.go
```diff
@@ -112,15 +110,11 @@ func (cm *Command) get(c *cli.Context, entityType string, getEntity func(string)
...
-		if err != nil {
+		if err := getEntity(id); err != nil {
 			return err
 		}
-
-		entities = append(entities, entity)
 	}
 
-	return cm.Printer.PrintEntities(entities)
+	return nil
```
tlake: So we've stripped out the `cm.Printer.PrintEntities(entities)` line, but I'm not seeing a replacement. How does the obtained entity get displayed to the user?

cli/command/environment_command.go
```diff
@@ -98,22 +97,30 @@ func (e *EnvironmentCommand) Create(c *cli.Context) error {
...
 }
 
 func (e *EnvironmentCommand) Delete(c *cli.Context) error {
 	return e.deleteWithJob(c, "environment", e.Client.DeleteEnvironment)
 }
 
 func (e *EnvironmentCommand) Get(c *cli.Context) error {
-	return e.get(c, "environment", func(id string) (entity.Entity, error) {
+	environments := []*models.Environment{}
+	getEnvironmentf := func(id string) error {
```
tlake: I like this internal functions pattern.

cli/client/job_test.go
```diff
@@ -36,7 +35,7 @@ func TestSelectByID(t *testing.T) {
 	client, server := newClientAndServer(handler)
 	defer server.Close()
 
-	job, err := client.SelectByID("id")
+	job, err := client.GetJob("id")
```
tlake: Sorry I didn't catch these Select changes last iteration - I saw them, but didn't realize they weren't intentional.

cli/printer/interface.go
```diff
@@ -1,16 +1,20 @@
...
+	PrintDeploys(deploys ...*models.Deploy) error
+	PrintDeploySummaries(deploys ...*models.DeploySummary) error
+	PrintEnvironments(environments ...*models.Environment) error
+	PrintJobs(jobs ...*models.Job) error
+	PrintLoadBalancers(loadBalancers ...*models.LoadBalancer) error
+	PrintLogs(logs ...*models.LogFile) error
+	PrintServices(services ...*models.Service) error
+	PrintTasks(tasks ...*models.Task) error
+	Printf(format string, tokens ...interface{})
+	Fatalf(code int64, format string, tokens ...interface{})
```
tlake: https://twitter.com/eevee/status/820579150939328512 :+1:

cli/printer/json.go
```diff
@@ -3,73 +3,82 @@ package printer
...
-// don't use spinner for json output
-func (this *JSONPrinter) StartSpinner(string) {}
-func (this *JSONPrinter) StopSpinner()        {}
+func (j *JSONPrinter) StartSpinner(string) {}
+func (j *JSONPrinter) StopSpinner()        {}
 
-func (this *JSONPrinter) PrintEntity(e entity.Entity) error {
-	return this.PrintEntities([]entity.Entity{e})
+func (j *JSONPrinter) Printf(format string, tokens ...interface{}) {
+	message := struct{
```
tlake: Oh, cool, anonymous structs. Learnin' new stuff thanks to zpatrick!

cli/client/job_test.go
zpatrick: no worries, my fault for `sed`-ing code

api/logic/load_balancer_logic.go
```diff
@@ -168,7 +168,7 @@ func (this *L0LoadBalancerLogic) populateModel(model *models.LoadBalancer) error
 	}
 
 	for _, tag := range tags.WithKey("load_balancer_id").WithValue(model.LoadBalancerID) {
-		model.Services = append(model.Services, tag.EntityID)
+		model.ServiceID = tag.EntityID
 	}
```
zpatrick: This should also lookup `model.ServiceName`

PR #39, opened 2017-01-17
Bulk List for Environments

zpatrick:
Improves performance in the API, prints the summary model in the CLI


api/backend/ecs/environment_manager.go
```diff
@@ -49,19 +49,17 @@ func (this *ECSEnvironmentManager) ListEnvironments() ([]*models.Environment, er
 		return nil, err
 	}
 
-	models := []*models.Environment{}
-	for _, cluster := range clusters {
+	environments := make([]*models.Environment, len(clusters))
```
diemonster: this is an across the board comment, but I much prefer using context-specific names over `models` as you do here (and throughout) 👍

Issue #30, opened 2017-01-11
Create test framework


zpatrick:
Wanted to open up a discussion on what we would like our test framework to look like, how the tests should run, what goals we should have, etc. etc.

### High Level Goals
* Instead of unit-test level of precision (e.g. did x security group get created?), we test behaviors that a customer would care about - are N copies of my service running? Are there a reasonable number of ec2 instances in my environment? Is my service reachable? Are my services being killed off when they become unhealthy? Can I run 300 tasks in a row? Can I list all 500 of my deploys?
* Tests should run async in discrete environments. 
* Tests should cleanup after themselves.
* There should be as few moving parts as possible - moving parts being programs and services that are required to execute the tests, and the number of steps required to run the test from start to finish.
* We should abstract away resource creation and cleanup as much as possible. Each test shouldn't have 50+ lines of code for creating and deleting each environment, service, etc.
* Don't reinvent the wheel in regards to local-machine<->layer0-api communication (if we can avoid it). For example, bash can use the `l0` binary. Go can use the `cli/client` package. However, if we used something like python, we would essentially have to re-write the `cli/client` package in python.

### Architecture
_Obviously this is still up in the air_

We use Terraform to create and destroy resources for each test case. 
Each test case has its own directory containing terraform files required for the test, e.g.
```
layer0/tests/integration/
    integration_test.go
    cases/
        case_one/
            layer0.tf
        case_two/
            layer0.tf
            aws.tf
```

The framework would take the directory name in its constructor and have some helper functions for executing terraform commands:
``` go
func TestCaseOne(t *testing.T) {
    context := NewIntegrationTestContext("case_one")
    // run terraform destroy when the test is done
    defer context.Destroy()

    // run terraform apply to create everything for the test
    context.Apply()

    // now you are ready to do some testing
    ...
}
```

#### Pros
* Don't need to write/maintain a whole bunch of new code dealing with resource creation/deletion for tests
* Reading/Creating/Updating resources for tests becomes extremely simple
* Idempotent create/delete means we can easily implement retries
* Test cleanup is done with `terraform destroy`
* You can add any AWS resources you need for your test into a terraform file

#### Cons
* You can't choose when resources are created; it happens all at once. However, the test case itself _could_ create resources during the test itself, it would just need to be responsible for destroying those resources before it completes. 
* As we've seen with l0-setup, making sure that the required terraform variables are being used correctly is a pretty error-prone process. 

### Test Service(s)
[...]

PR #51, opened 2017-01-25
System test framework

zpatrick:
Todo
* ~~Make a `-d/--debug` flag, change some statements to print on `debug` instead of `verbose`~~
* ~~Fill out README~~
* ~~Cleanup the [sts](https://github.com/quintilesims/sts) repo~~
* ~~Add comments at the top of tests specifying what gets created via terraform~~
* ~~Automatically run `terraform get` to sync up terraform modules~~

closes https://github.com/quintilesims/layer0/issues/30


tests/system/deadServiceRecreated_test.go
```diff
@@ -0,0 +1,26 @@
+package system
+
+import (
+	"testing"
+	"time"
+)
+
+func TestDeadServiceRecreated(t *testing.T) {
```
tlake: :+1: These tests are pretty sweet

tests/system/cases/dead_service_recreated/.terraform/modules/2aeccc0db5d6ee507c3067b347953328
```diff
@@ -0,0 +1 @@
+/home/ec2-user/go/src/github.com/quintilesims/layer0/tests/system/cases/modules/sts
```
zpatrick: done

tests/system/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+In addition to wrapping terraform commands, the `SystemTestContext` object contains some wrappers around the `cli/client` package, e.g.:
+```
+env := c.GetEnvironment("environment_name")
+svc := c.GetService("environment_name", "service_name")
+lb := c.GetLoadBalancer("environment_name", "load_balancer_name")
+```
+
+
+### Running a Test
```
diemonster: This whole doc is great :1st_place_medal: 

My only comment for here would be to define what we consider a 'system test' to actually be. This is probably best illustrated by whatever we have in `/cases` but there is still ambiguity around what a system test versus end-to-end test versus integration test. From what i'm seeing, we fall somewhere in the realm of system test because we're testing module functionality and not necessarily _workflow_.

tests/system/README.md
```diff
@@ -0,0 +1,85 @@
...
+Following the example above, the contents of `mySimpleExample_test.go` should look like:
+```go
+package system
+
+import (
+        "testing"
+)
+
+// Test Resources:
+// This test creates an environment named 'sme' that has a
```
diemonster: where does the `sme` name get passed from? The `my_simple_example` file? If so, would be worth clarifying that.

tests/system/README.md
```diff
@@ -0,0 +1,85 @@
...
+}
+```
+
+Each test case should have its own file with a matching name. 
+The test function has some comments about what resources are created by terraform.
+The `startSystemTest` function tells the test to run in parallel, creates a new `framework.SystemTestContext` object using the specified directory (`"cases/my_simple_example"` in this example), and runs `terraform apply`.
+The `defer c.Destroy()` statement ensures any resources created by terraform at the start of the test will be destroyed once the test has finished.
+
+### The SystemTestContext object
+In addition to wrapping terraform commands, the `SystemTestContext` object contains some wrappers around the `cli/client` package:
```
diemonster: are these wrappers only for `get`s?

diemonster:
somewhat tangential to this PR but not: what else does one need to know about https://github.com/quintilesims/sts in order to write these tests?

tests/system/main_test.go
```diff
@@ -0,0 +1,92 @@
...
+		Dir:    dir,
+		DryRun: *dry,
+		Vars:   vars,
+	})
+
+	c.Apply()
+	return c
+}
+
+func waitFor(t *testing.T, name string, timeout time.Duration, conditionSatisfied func() bool) {
```
diemonster: I might be misunderstanding the pattern, but this function seems integral to the `system` packages' API. As such, seeing it unexported caught me a little off-guard. It already seems generalized enough to be useable outside of this package, but maybe there is no use for doing that?

tests/system/README.md
zpatrick: fixed

tests/system/README.md
zpatrick: for now - it's simple enough to add new functions as we need them.j

tests/system/main_test.go
zpatrick: The `system` package doesn't export anything. It is generic enough to move to the testutils package, but hadn't done that since nothing outside of the `system` package uses it.

zpatrick:
> somewhat tangential to this PR but not: what else (if anything) does one need to know about https://github.com/quintilesims/sts in order to write these tests?

This is covered in the readme, and further info can be found by going to that repo.

tests/system/deadServiceRecreated_test.go
```diff
@@ -0,0 +1,29 @@
+package system
+
+import (
+	"testing"
+	"time"
+)
+
+// Test Resources:
+// This test creates an environment named 'dsr' that has a
```
zpatrick: yeah, it's just an acronym of the test case name. This was added into the readme

Issue #34, opened 2017-01-12
Environment delete error due to waiting for tasks to stop


tribaljack:
deleted environment once,
that seemed to hang:

```
l0-job
------
2016-12-08 01:49:02 [Main] INFO : Loading job '4e4fb3df25de'
2016-12-08 01:49:02 [Main] INFO : Running step 'Delete Dependencies'
2016-12-08 01:49:02 [Main] INFO : Running Action: DeleteEnvironmentServices
2016-12-08 01:49:02 [Main] INFO : Running Action: DeleteEnvironmentLoadBalancers
2016-12-08 01:49:03 [Main] INFO : Running Action: DeleteService on 'TheChune7a13'
2016-12-08 01:49:03 [Main] DEBUG: Waiting for service to stop
2016-12-08 01:49:03 [Main] DEBUG: Wait Stop Service l0-prod084-TheChune7a13 iteration 1 of %!s(int=10)
2016-12-08 01:49:07 [Main] DEBUG: Wait Stop Service l0-prod084-TheChune7a13 iteration 2 of %!s(int=10)

```
deleted again
that failed:
```
 ~  l0 job logs 4e4fb3df25de
l0-job
------
2016-12-08 01:49:02 [Main] INFO : Loading job '4e4fb3df25de'
2016-12-08 01:49:02 [Main] INFO : Running step 'Delete Dependencies'
2016-12-08 01:49:02 [Main] INFO : Running Action: DeleteEnvironmentServices
2016-12-08 01:49:02 [Main] INFO : Running Action: DeleteEnvironmentLoadBalancers
2016-12-08 01:49:03 [Main] INFO : Running Action: DeleteService on 'TheChune7a13'
2016-12-08 01:49:03 [Main] DEBUG: Waiting for service to stop
2016-12-08 01:49:03 [Main] DEBUG: Wait Stop Service l0-prod084-TheChune7a13 iteration 1 of %!s(int=10)
2016-12-08 01:49:07 [Main] DEBUG: Wait Stop Service l0-prod084-TheChune7a13 iteration 2 of %!s(int=10)
2016-12-08 01:49:11 [Main] DEBUG: Wait Stop Service l0-prod084-TheChune7a13 iteration 3 of %!s(int=10)
2016-12-08 01:49:15 [Main] DEBUG: Wait Stop Service l0-prod084-TheChune7a13 iteration 4 of %!s(int=10)
[...]

Issue #10, opened 2016-12-09
Delete Service frequently times out 


zpatrick:
We've seen this in the past when Docker containers don't respond to kill commands. 
Currently, when we deleted a service, we just scale the DesiredCount to 0 and then call `ecs.DeleteService`. 

I'm guessing that if we call `ecs.StopTask` for each of the running tasks in the Service, this may result in a more "forceful" delete, and could potentially fix the timeouts we've been seeing. Checkout the docs for StopTask here: https://github.com/aws/aws-sdk-go/blob/master/service/ecs/api.go#L7385

PR #52, opened 2017-01-25
Force service delete

zpatrick:
This calls `ecs.StopTask` on each task a service is running before deleting the service. This also removes the `waitUntilServiceStopped` function, as it appears ecs no longer requires the number of running tasks to be 0 before being able to call `ecs.DeleteService`. I've also bumped up the wait times for jobs, as I noticed relatively small environments are bumping up to about 60% of max allowed time. 

Iv'e tested this against an environment with a large number of services running (10), but this still needs to be QA'd against a service that doesn't respond to kill commands before merging. 

This fixes https://github.com/quintilesims/layer0/issues/34 and https://github.com/quintilesims/layer0/issues/10

### DIFF
--- a/api/backend/ecs/service_manager.go
+++ b/api/backend/ecs/service_manager.go
@@ -120,12 +120,28 @@ func (this *ECSServiceManager) DeleteService(environmentID, serviceID string) er
 	ecsServiceID := id.L0ServiceID(serviceID).ECSServiceID()
 	desiredCount := int64(0)
 
-	if err := this.ECS.UpdateService(ecsEnvironmentID.String(), ecsServiceID.String(), nil, &desiredCount); err != nil {
-		return err
+	service, err := this.GetService(environmentID, serviceID)
+        if err != nil {
+                return  err
+        }
+
+        taskARNs := []*string{}
+        for _, deployment := range service.Deployments {
+                arns, err := getTaskARNs(this.ECS, ecsEnvironmentID, stringp(deployment.DeploymentID))
+                if err != nil {
+                        return err
+                }
+
+                taskARNs = append(taskARNs, arns...)
+        }
+
+	for _, arn := range taskARNs{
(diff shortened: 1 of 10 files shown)


zpatrick:
Update: I've tested this against the Chunnel's frontend service (which they've had issues deleting), and was able to delete the service just fine.

api/backend/ecs/service_manager.go
```diff
@@ -120,12 +120,28 @@ func (this *ECSServiceManager) DeleteService(environmentID, serviceID string) er
...
+        }
+
+	for _, arn := range taskARNs{
+		if err := this.ECS.StopTask(ecsEnvironmentID.String(), "Service deleted by User", *arn); err != nil{
+			log.Warnf("Stop Task for Service '%s' had error: %v", serviceID, err)
+		}
 	}
 
-	log.Debugf("Waiting for service to stop")
-	if err := this.waitUntilServiceStopped(ecsEnvironmentID, ecsServiceID); err != nil {
```
diemonster: can you think of an event where forcefully stopping a service would be undesirable in comparison to the current behavior? i.e. the difference between killing a process and waiting for it to gracefully shutdown versus `kill -9`? I suppose this necessitates us having a better understanding of how `StopTask` works...

diemonster: To expand on my point above, my concern is mostly relating to services that "do things" when they receive a graceful shutdown interrupt, i.e. write final log entries, set permissions or locks on external resources, etc. 

To that end, I wonder if it would be worth exposing two forms of `delete`: the old one that uses `waitUntilServiceStopped` and this new one that is more forceful.

zpatrick: documented here: https://github.com/aws/aws-sdk-go/blob/master/service/ecs/api.go#L2325

diemonster: nice, ok. This is pretty reasonable then.

PR #53, opened 2017-01-26
WIP: Load Balancer Healthchecks (L0)

### LINKED ISSUE #13: Configure load balancer healthchecks
Allow users to specify health checks for layer0 load balancers on the cli and in the `terraform-layer0-provider`. The terraform format should follow the existing `health_check` block for terraform: https://www.terraform.io/docs/providers/aws/r/elb.html

Possible UX for the cli:
`l0 loadbalancer create --healthcheck `
`l0 loadbalancer sethealthcheck`

tlake:
There are a few things that still need working out, but here's the feature. I'm sure there are things that could be shored up and embetterified.

Still needs to pull in latest from `develop`.

The terraform aspect still needs to happen.

Running into an error when running smoketests: creating a lb with a cert fails with an error message of `invalid character 'r' looking for beginning of value`. Running the command with `-d` debug flag points at `backend/ecs/load_balancer_manager.go:612`, which is inside of the `getSubnetsAndAvailZones()` function. Cursory examination seems to indicate that `availZones` is an empty string slice at the point that it's used for `range`, captain.

References #13

### DIFF
--- a/api/backend/ecs/load_balancer_manager.go
+++ b/api/backend/ecs/load_balancer_manager.go
@@ -36,8 +36,8 @@ func NewECSLoadBalancerManager(ec2 ec2.Provider, elb elb.Provider, iam iam.Provi
 	}
 }
 
-func (this *ECSLoadBalancerManager) ListLoadBalancers() ([]*models.LoadBalancer, error) {
-	loadBalancerDescriptions, err := this.ELB.DescribeLoadBalancers()
+func (e *ECSLoadBalancerManager) ListLoadBalancers() ([]*models.LoadBalancer, error) {
+	loadBalancerDescriptions, err := e.ELB.DescribeLoadBalancers()
 	if err != nil {
 		return nil, err
 	}
@@ -57,10 +57,10 @@ func (this *ECSLoadBalancerManager) ListLoadBalancers() ([]*models.LoadBalancer,
 	return loadBalancers, nil
 }
 
-func (this *ECSLoadBalancerManager) GetLoadBalancer(loadBalancerID string) (*models.LoadBalancer, error) {
+func (e *ECSLoadBalancerManager) GetLoadBalancer(loadBalancerID string) (*models.LoadBalancer, error) {
 	ecsLoadBalancerID := id.L0LoadBalancerID(loadBalancerID).ECSLoadBalancerID()
 
-	loadBalancer, err := this.ELB.DescribeLoadBalancer(ecsLoadBalancerID.String())
+	loadBalancer, err := e.ELB.DescribeLoadBalancer(ecsLoadBalancerID.String())
 	if err != nil {
 		if ContainsErrCode(err, "LoadBalancerNotFound") {
 			err := fmt.Errorf("LoadBalancer with id '%s' does not exist", loadBalancerID)
@@ -70,47 +70,47 @@ func (this *ECSLoadBalancerManager) GetLoadBalancer(loadBalancerID string) (*mod
 		return nil, err
 	}
 
-	// todo: does describe laodbalancer return nil or erro?
-	return this.populateModel(loadBalancer), nil
+	// todo: does describe loadbalancer return nil or erro?
+	return e.populateModel(loadBalancer), nil
 }
 
-func (this *ECSLoadBalancerManager) DeleteLoadBalancer(loadBalancerID string) error {
+func (e *ECSLoadBalancerManager) DeleteLoadBalancer(loadBalancerID string) error {
 	ecsLoadBalancerID := id.L0LoadBalancerID(loadBalancerID).ECSLoadBalancerID()
 	roleName := ecsLoadBalancerID.RoleName()
 
-	policyList, err := this.IAM.ListRolePolicies(roleName)
+	policyList, err := e.IAM.ListRolePolicies(roleName)
 	if err != nil && !ContainsErrCode(err, "NoSuchEntity") {
 		return err
 	}
 
 	for _, name := range policyList {
 		policy := stringOrEmpty(name)
-		if err := this.IAM.DeleteRolePolicy(roleName, policy); err != nil {
+		if err := e.IAM.DeleteRolePolicy(roleName, policy); err != nil {
 			return err
 		}
 	}
 
-	if err := this.waitUntilRolePoliciesDeleted(roleName); err != nil {
+	if err := e.waitUntilRolePoliciesDeleted(roleName); err != nil {
 		return err
 	}
 
-	if err := this.IAM.DeleteRole(roleName); err != nil {
+	if err := e.IAM.DeleteRole(roleName); err != nil {
 		if !ContainsErrCode(err, "NoSuchEntity") {
 			return err
 		}
 	}
 
-	if err := this.waitUntilRoleDeleted(roleName); err != nil {
+	if err := e.waitUntilRoleDeleted(roleName); err != nil {
 		return err
 	}
 
-	if err := this.ELB.DeleteLoadBalancer(ecsLoadBalancerID.String()); err != nil {
+	if err := e.ELB.DeleteLoadBalancer(ecsLoadBalancerID.String()); err != nil {
 		if !ContainsErrCode(err, "NoSuchEntity") {
 			return err
 		}
 	}
 
-	securityGroup, err := this.EC2.DescribeSecurityGroup(ecsLoadBalancerID.SecurityGroupName())
+	securityGroup, err := e.EC2.DescribeSecurityGroup(ecsLoadBalancerID.SecurityGroupName())
 	if err != nil {
 		return err
 	}
@@ -119,7 +119,7 @@ func (this *ECSLoadBalancerManager) DeleteLoadBalancer(loadBalancerID string) er
 	// todo: using waiters seems pretty verbose - should find a way to clean this up
 	if securityGroup != nil {
 		check := func() (bool, error) {
-			if err := this.EC2.DeleteSecurityGroup(securityGroup); err == nil {
+			if err := e.EC2.DeleteSecurityGroup(securityGroup); err == nil {
 				return true, nil
 			}
 
@@ -130,7 +130,7 @@ func (this *ECSLoadBalancerManager) DeleteLoadBalancer(loadBalancerID string) er
 			Name:    fmt.Sprintf("SecurityGroup delete for '%s'", securityGroup),
 			Retries: 30,
 			Delay:   time.Second * 10,
-			Clock:   this.Clock,
+			Clock:   e.Clock,
 			Check:   check,
 		}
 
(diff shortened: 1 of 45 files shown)


cli/command/load_balancer_command_test.go
```diff
@@ -273,3 +286,100 @@ func TestListLoadBalancers(t *testing.T) {
...
+
+	tc.Client.EXPECT().
+		GetLoadBalancer("id").
+		Return(&models.LoadBalancer{}, nil)
+
+	flags := Flags{
+		"set-target": "TCP",
+	}
+
+	contexts := map[string]*cli.Context{
```
zpatrick: this needs more scenarios:
missing 1 arg, missing 2 args, non-integer values for the integer flags, more cases around a bad `set-target` flag, etc.

api/backend/ecs/load_balancer_manager.go
```diff
@@ -301,26 +317,53 @@ func (this *ECSLoadBalancerManager) createLoadBalancer(
...
 		return err
 	}
 
 	return nil
 }
 
-func (this *ECSLoadBalancerManager) UpdateLoadBalancer(loadBalancerID string, ports []models.Port) (*models.LoadBalancer, error) {
-	model, err := this.GetLoadBalancer(loadBalancerID)
+func (e *ECSLoadBalancerManager) UpdateLoadBalancerHealthCheck(loadBalancerID string, healthCheck models.HealthCheck) (*models.LoadBalancer, error) {
+	model, err := e.GetLoadBalancer(loadBalancerID)
```
zpatrick: [nit] Switching the order of this function a bit will improve performance (in most error scenarios) :
```go
func (e *ECSLoadBalancerManager) UpdateLoadBalancerHealthCheck(loadBalancerID string, healthCheck models.HealthCheck) (*models.LoadBalancer, error) {
    ecsLoadBalancerID := id.L0LoadBalancerID(loadBalancerID).ECSLoadBalancerID()
    if err := e.updateHealthCheck(ecsLoadBalancerID, healthCheck); err != nil {
        return nil, err
    }

    return e.GetLoadBalancer(loadBalancerID)
}

```

cli/printer/text.go
```diff
@@ -171,6 +171,32 @@ func (t *TextPrinter) PrintLoadBalancers(loadBalancers ...*models.LoadBalancer)
...
+func (t *TextPrinter) PrintLoadBalancerHealthCheck(loadBalancer *models.LoadBalancer) error {
+	getEnvironment := func(l *models.LoadBalancer) string {
+		if l.EnvironmentName != "" {
+			return l.EnvironmentName
+		}
+
+		return l.EnvironmentID
+	}
+
+	rows := []string{"LOADBALANCER ID | LOADBALANCER NAME | ENVIRONMENT | TARGET | INTERVAL | TIMEOUT | HEALTHY THRESHOLD | UNHEALTHY THRESHOLD "}
```
tlake: It is a bit long, but I also feel like we should show something to the user, so they can confirm that the command worked. Maybe it's just enough to not return an error.

However, when you run `loadbalancer healthcheck lb-name` without any flags, it will currently print out the current health check for that load balancer, which seems useful to me.

Is there a different format we could use to display healthcheck information to the user? Should we just drop that line of reasoning altogether?

tests/smoke/load_balancer.bats
```diff
@@ -50,24 +50,40 @@ certificate_name="l0-$LAYER0_PREFIX-api"
     l0 loadbalancer delete --wait loadbalancer2
 }
 
-@test "loadbalancer create --port 80:80/http test loadbalancer3" {
-    l0 loadbalancer create --port 80:80/http test loadbalancer3
+@test "loadbalancer create --healthcheck-target 'TCP:80' --healthcheck-interval 30 --healthcheck-timeout 5 --healthcheck-healthy-threshold 2 --healthcheck-unhealthy-threshold 2 loadbalancer3" {
```
tlake: Truth.

api/logic/load_balancer_logic_test.go
```diff
@@ -226,3 +237,32 @@ func TestUpdateLoadBalancer(t *testing.T) {
...
+		UpdateLoadBalancerHealthCheck("lb_id", healthCheck).
+		Return(&models.LoadBalancer{HealthCheck: healthCheck}, nil)
+
+	loadBalancerLogic := NewL0LoadBalancerLogic(testLogic.Logic())
+	received, err := loadBalancerLogic.UpdateLoadBalancerHealthCheck("lb_id", healthCheck)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	expected := &models.LoadBalancer{
```
tlake: ✅

cli/command/load_balancer_command_test.go
```diff
@@ -273,3 +286,100 @@ func TestListLoadBalancers(t *testing.T) {
...
+
+	tc.Resolver.EXPECT().
+		Resolve("load_balancer", "env").
+		Return([]string{"id"}, nil)
+
+	tc.Client.EXPECT().
+		GetLoadBalancer("id").
+		Return(&models.LoadBalancer{}, nil)
+
+	// We're testing a call is NOT made to the client for updating the health check
```
tlake: ✅

cli/command/load_balancer_command.go
```diff
@@ -221,6 +281,78 @@ func (l *LoadBalancerCommand) Get(c *cli.Context) error {
...
+	}
+
+	loadBalancer, err := l.Client.GetLoadBalancer(id)
+	if err != nil {
+		return err
+	}
+
+	updateIsRequired := false
+	if target := c.String("set-target"); target != "" {
+		if split := strings.Split(target, ":"); len(split) != 2 {
```
tlake: It does look like AWS gives some pretty decent errors - I might just leave that error handling up to them.

```
$ l0 loadbalancer healthcheck --set-target abbabba:9999999 test-lb-1
AWS Error: healthcheck target must begin with one of http, tcp, https, ssl (code 'ValidationError')

$ l0 loadbalancer healthcheck --set-target http:9999999 test-lb-1
AWS Error: healthcheck http target must specify a port followed by a path that begins with a slash. e.g. http:80/ping/this/path (code 'ValidationError')
```

cli/command/load_balancer_command.go
```diff
@@ -51,6 +51,31 @@ func (l *LoadBalancerCommand) GetCommand() cli.Command {
 						Name:  "private",
 						Usage: "if specified, creates a private load balancer (default is public)",
 					},
+					cli.StringFlag{
+						Name:  "healthcheck-target",
+						Value: "TCP:80",
+						Usage: "health check target in format 'PROTOCOL:PORT'",
```
tlake: ✅

cli/command/load_balancer_command.go
```diff
@@ -221,6 +281,78 @@ func (l *LoadBalancerCommand) Get(c *cli.Context) error {
 	return l.Printer.PrintLoadBalancers(loadBalancers...)
 }
 
+func (l *LoadBalancerCommand) HealthCheck(c *cli.Context) error {
+	args, err := extractArgs(c.Args(), "NAME")
+	if err != nil {
+		return err
+	}
+
+	id, err := l.resolveSingleID("load_balancer", args["NAME"])
```
tlake: ✅

cli/command/load_balancer_command.go
```diff
@@ -221,6 +281,78 @@ func (l *LoadBalancerCommand) Get(c *cli.Context) error {
...
+		if split := strings.Split(target, ":"); len(split) != 2 {
+			return NewUsageError("Health Check Target format is: PROTOCOL:PORT")
+		}
+
+		updateIsRequired = true
+		loadBalancer.HealthCheck.Target = target
+	}
+
+	if interval := c.String("set-interval"); interval != "" {
+		if i, err := strconv.Atoi(interval); err != nil {
```
tlake: ✅

cli/command/load_balancer_command_test.go
tlake: Good call on the non-int values; updating that.

The way the command is set up, though, the user doesn't need to specify all of the flags (I assume that's what you mean by missing args?) - just the ones they want to update. Calling `loadbalancer healthcheck --set-interval 5 name-of-lb` shouldn't fail even though none of the other flags are defined.

Also, since AWS has decent `target` validation errors, it seems to me that I don't need to test for it in this function. Sound right?

api/backend/ecs/load_balancer_manager.go
tlake: ✅

cli/printer/text.go
zpatrick: If anything, I think we'd show just `TARGET`. If we do, it would be nice to take something out to prevent that table from getting too long

cli/command/load_balancer_command.go
```diff
@@ -221,6 +281,100 @@ func (l *LoadBalancerCommand) Get(c *cli.Context) error {
...
+			loadBalancer.HealthCheck.UnhealthyThreshold = healthCheck.UnhealthyThreshold
+		}
+
+		loadBalancer, err = l.Client.UpdateLoadBalancerHealthCheck(id, loadBalancer.HealthCheck)
+		if err != nil {
+			return err
+		}
+	}
+
+	return l.Printer.PrintLoadBalancerHealthCheck(loadBalancer)
```
zpatrick: :+1: these changes look good

cli/command/load_balancer_command_test.go
```diff
@@ -273,3 +286,86 @@ func TestListLoadBalancers(t *testing.T) {
...
+	}
+}
+
+func TestHealthCheck_userInputErrors(t *testing.T) {
+	tc, ctrl := newTestCommand(t)
+	defer ctrl.Finish()
+	command := NewLoadBalancerCommand(tc.Command())
+
+	contexts := map[string]*cli.Context{
+		"Non-int '--set-interval' flag":            getCLIContext(t, Args{"env", "name"}, Flags{"set-interval": "two"}),
```
zpatrick: Almost there:
* This command only takes one arg, so these calls should be `getCLIContext(t, Args{"name"}, Flags{...`
* There should be a case for a missing `NAME` arg: `getCLIContext(t, nil, Flags{...`

api/backend/ecs/load_balancer_manager.go
```diff
@@ -301,26 +317,53 @@ func (this *ECSLoadBalancerManager) createLoadBalancer(
...
 
-func (this *ECSLoadBalancerManager) UpdateLoadBalancer(loadBalancerID string, ports []models.Port) (*models.LoadBalancer, error) {
-	model, err := this.GetLoadBalancer(loadBalancerID)
+func (e *ECSLoadBalancerManager) UpdateLoadBalancerHealthCheck(loadBalancerID string, healthCheck models.HealthCheck) (*models.LoadBalancer, error) {
+	ecsLoadBalancerID := id.L0LoadBalancerID(loadBalancerID).ECSLoadBalancerID()
+	if err := e.updateHealthCheck(ecsLoadBalancerID, healthCheck); err != nil {
+		return nil, err
+	}
+
+	model, err := e.GetLoadBalancer(loadBalancerID)
```
zpatrick: [nit] This block is redundant since the healthcheck is already populated in `populateModel`. Should be:
```
func (e *ECSLoadBalancerManager) UpdateLoadBalancerHealthCheck(loadBalancerID string, healthCheck models.HealthCheck) (*models.LoadBalancer, error) {
    ecsLoadBalancerID := id.L0LoadBalancerID(loadBalancerID).ECSLoadBalancerID()
    if err := e.updateHealthCheck(ecsLoadBalancerID, healthCheck); err != nil {
        return nil, err
    }

    return e.GetLoadBalancer(loadBalancerID)
}
```

tlake: ✅

cli/command/load_balancer_command_test.go
tlake: ✅

tlake:
Should just have updating with `develop` remaining

tlake:
I think that should do it

zpatrick:
:+1: LGTM

PR #64, opened 2017-01-28
Auto smoketests Testing #2

diemonster:


zpatrick:
@bivers: something is wonky with this branch. Looks like it was based off of a super old version of develop.

Issue #55, opened 2017-01-27
l0-setup only checks for dockercfg


diemonster:
Versions of docker since April 2015 automatically use a new file format, 'config.json':
http://stackoverflow.com/questions/35519072/docker-config-json-vs-dockercfg

We need to patch l0-setup so that it can use either `dockercfg` or `config.json`

jonathanstallings:
Will do. Question: If neither type of config file (`dockercfg` or `config.json`) is provided by user, which type do we want to create as a default?

zpatrick:
We use the "dockercfg Authentication Format" described [here](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/private-auth.html). 
Note that those docs are actually wrong on this part:
```
$ cat ~/.docker/config.json
{"https://index.docker.io/v1/":{"auth":"zq212MzEXAMPLE7o6T25Dk0i","email":"email@example.com"}}
```

:point_up: That's in `.dockercfg` format, not a `config.json` format. 
The `config.json` format has an`"auths"` field, it would look like:
```
{
        "auths": {
                "https://index.docker.io/v1/": {
                        "auth": "zq212MzEXAMPLE7o6T25Dk0i",
                        "email": "email@example.com"
                }
        }
}
```

So we would want to transform anything in `config.json` format to `.dockercfg` format. 

**...However** 
I don't think anyone has actually tested what happens if you just pass in a `config.json`. There's a chance it may work as is, in which case we could just close this issue. Probably worth checking out.

diemonster:
If you just copy a `config.json` in place of a `dockercfg`, everything seems to work just fine

jonathanstallings:
Thanks, both. Given the above, I've made a very small change that advises the user that they may pass a `config.json` format file to the `--dockercfg` flag.

zpatrick:
@diemonster did you try with a new environment that pulled from a private repo? Running `l0-setup apply` and using an environment that already existed wouldn't have pulled the new docker config file.

diemonster:
@zpatrick I came across this issue specifically when needing `push` access to the quintilesims/l0-api Dockerhub repo. So while it was a public repo, I required creds to push as part of CI. As such, my workflow when discovering the issue was: 
- `docker login -u username -p password`
- `l0-setup apply smoketest`
- spawn a new subshell or a new container
- `l0-setup restore smoketest`

At this point, I get a notification that the `dockercfg` file was not restored.

jonathanstallings:
[...]

PR #66, opened 2017-01-30
WIP: Advise that user may pass config.json format to dockercfg flag (#55)

jonathanstallings:
Brandon determined that using a `config.json` format as a `dockercfg` file seemed to work without issue. Therefore, I am simply advising users that they may pass this format to the `--dockercfg` flag on apply.

TODO: Check for additional work based on feedback on issue #55


zpatrick:
This does not work. See: https://github.com/quintilesims/layer0/issues/55

PR #71, opened 2017-02-01
add end-to-end smoketest

diemonster:
These smoketests run only when a PR is made to the `master` branch _or_ you are attempting to merge straight to `master`. 

Smoketests are currently failing w code in `develop`  but the workflow passes, including tearing down the `smoketest` Layer0 instance:
https://travis-ci.org/quintilesims/layer0/builds/197439405


tests/smoke/Makefile
```diff
@@ -1,4 +1,12 @@
-test:
+SHELL:=/bin/bash
+
+deps:
+	git clone https://github.com/sstephenson/bats.git
+	mkdir ~/bin || :
```
diemonster: On Travis CI, we have no promises that the `travis` user has permissions to write to `/usr/local/...`. Even when installing via `sudo`  I ran into this issue, and related permissions problems around it. In the end I opted to use a custom bin directory, which is portable to all linux distros anyway.

scripts/ci-smoketest.sh
```diff
@@ -0,0 +1,30 @@
...
+set -x
+
+# Ignore if master branch is not target
+if [ ! "$TRAVIS_BRANCH" == "master" ]; then
+  echo "[INFO] Skipping Smoketest";
+  exit 0
+fi
+
+# Tag repository
+git tag "$LAYER0_PREFIX"
```
diemonster: Two reasons: 
1) too many image hashes in our Dockerhub --- do we really need a separate hash for every smoketest?
2) when I was previously running this all in a Docker container, it turns out Travis does not preserve tag data, so our method of `git describe` was return nothing.

api/Makefile
```diff
@@ -1,6 +1,10 @@
+SHELL:=/bin/bash
 L0_VERSION?=$(shell git describe --tags)
 API_DOCKER_IMAGE=quintilesims/l0-api:$(L0_VERSION)
 
+deps:
+	@docker login -u $(DOCKERHUB_USERNAME) -p $(DOCKERHUB_PASSWORD)
```
diemonster: No. The current workflow here is not running in a container, anyway. Having it in `deps` here will allow this to work in either a docker or non-docker context

setup/Makefile
```diff
@@ -16,16 +21,34 @@ download-terraform: build-directories
...
-        done
+  done
 
 build: download-terraform
 	CGO_ENABLED=0 GOOS=linux   GOARCH=amd64 go build -ldflags "-s -X main.Version=$(L0_VERSION)" -a -o build/linux/l0-setup .
 	CGO_ENABLED=0 GOOS=darwin  GOARCH=amd64 go build -ldflags "-s -X main.Version=$(L0_VERSION)" -a -o build/darwin/l0-setup .
 	CGO_ENABLED=0 GOOS=windows GOARCH=amd64 go build -ldflags "-s -X main.Version=$(L0_VERSION)" -a -o build/windows/l0-setup.exe .
 
+install-smoketest: deps
+		for os in linux; do \
```
diemonster: yeah this could be simplified.... lazy cut'n'paste ;)

setup/Makefile
```diff
@@ -16,16 +21,34 @@ download-terraform: build-directories
...
+                mkdir -p terra ; \
+                cd terra ; \
+                aws s3 cp s3://xfra-layer0/terraform/$$os/terraform_$(TERRAFORM_VERSION)_$$os.zip . ; \
+                unzip terraform_$(TERRAFORM_VERSION)_$$os.zip ; \
+                rm terraform_$(TERRAFORM_VERSION)_$$os.zip ; \
+                cd - ; \
+                cp ./terra/terraform ~/bin/ ; \
+    done
+	  CGO_ENABLED=0 GOOS=linux   GOARCH=amd64 go build -ldflags "-s -X main.Version=$(L0_VERSION)" -a -o ./l0-setup .
```
diemonster: could work with `cli`, but I had problems w `l0-setup` and its runtime context. When building the `l0-setup` binary, it needed to be run alongside the files in `/setup`. Maybe you have a trick to get around that.

cli/Makefile
```diff
@@ -1,13 +1,20 @@
...
 release: build
 
 test:
-	go test ./...
+		go test ./...
 
-.PHONY: build release test
+install-smoketest:
+		CGO_ENABLED=0 GOOS=linux   GOARCH=amd64 go build -ldflags "-s -X main.Version=$(L0_VERSION)" -a -o build/linux/l0 .
+		mkdir ~/bin || :
```
diemonster: Yeah this would probably work fine for the `l0` binary

scripts/ci-smoketest.sh
zpatrick: 1 - yes, if we wanted to have multiple smoketests run at once from different PRs. I don't think dockerhub has a limit on image tags?
2. That's no good, that will be a blocker if/when we wanted to run `make release`  in travis. Perhaps we need to create an issue to see if there's anyway we can make travis keep `.git` data?

api/Makefile
zpatrick: ah i see

setup/Makefile
zpatrick: ah right, nevermind then.

tests/smoke/Makefile
zpatrick: I see, makes sense. 

Not that this is a "dirty" solution or anything,  but this makes me think we should start carving out a common pattern for "setting things up for CI" vs "setting things up locally" in our Makefiles. Hopefilly, its very clear for future maintainers "hey, this is slightly odd because it's running in CI, not necessarily your local environment"

scripts/ci-smoketest.sh
diemonster: re: Travis keeping git data, there are multiple bugs out for it. Some people have tried preventing Travis' "shallow clone" by setting the `depth` to 99999 and/or all sorts of other weird workarounds. More time could be spent exploring this, but the bottom line is that the behavior is non-standard. 

One idea I had was to update a `git tag` variable as part of Travis' global env vars when we run a release. Another would be to set the tag in the `travis.yml` file prior to a release. Essentially it's hardcoded for now, and I feel that we could refactor this part when we add the release automation.

tests/smoke/Makefile
zpatrick: Since the nuances are going to change based on whatever the CI engine is, perhaps we have a pattern
where any command prefixed with `travis-` does the extra work required to getting it running in Travis CI:
```
make deps   =   get/install dependencies locally
make travis-deps  = get/install dependencies for travis ci
make smoketests = run smoketests locally, assuming you have all deps in place
make travis-smoketests = get/install deps for smoketests, run them, cleanup afterwards
```

scripts/ci-smoketest.sh
zpatrick: I see, well it's not a blocker for now so let's deal with it when we get there.

tests/smoke/Makefile
diemonster: Agreed. Setting a custom bin dir in `~/` is not super weird, but it's definitely not a common as using `/usr/local/...`. The makefiles take care of setting the paths appropriately, though. 

I really had to do this as a compromise as running things in a docker context didn't work due to how `l0` takes config. I think if we improved CI workflow for `l0` and `l0-setup`, 99% of the weirdness of this CI workflow would be alleviated.

zpatrick: sure, this is a non-blocker. I'll make an issue for it

tests/smoke/Makefile
```diff
@@ -1,4 +1,10 @@
-test:
+SHELL:=/bin/bash
+
+deps:
+	git clone https://github.com/sstephenson/bats.git
+	sudo ./bats/install.sh /usr/local
+
+test: deps
```
zpatrick: `deps` shouldn't be a child of `test`

diemonster: fixed

PR #76, opened 2017-02-02
Update README for local API setup

jonathanstallings:
* Necessary to run main.go from within setup directory
* Need to unset auth token from API shell as well


zpatrick:
:+1:  looks good

PR #84, opened 2017-02-06
Convert config.json to dockercfg format as needed (#55)

jonathanstallings:
* Allow l0-setup to read docker config from `config.json` or `dockercfg`
* Write `dockercfg` format to instance directory
* Allow `--docker-config-dir` flag to specify directory to read config from
* Attempt to read from global docker config dir if not found at given config directory
* Fall back to writing "empty" dockercfg file (`{}`) if needed
* Preserve `--force` flag, which disables alert on `empty` dockercfg

Would love any feedback you have, especially regarding preferences for where to organize the code additions such as `/setup/context/dockerauth.go`.

### DIFF
--- a/setup/context/commands.go
+++ b/setup/context/commands.go
@@ -5,12 +5,14 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
+	"github.com/docker/docker/pkg/homedir"
 	"github.com/quintilesims/layer0/common/aws/provider"
 	"github.com/quintilesims/layer0/common/aws/s3"
 	"github.com/quintilesims/layer0/common/config"
 	"io/ioutil"
 	"net/http"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 )
@@ -27,7 +29,7 @@ type TFModule struct {
 	Resources map[string]interface{} `json:"resources"`
 }
 
-func Apply(c *Context, force bool, dockercfg string) error {
+func Apply(c *Context, force bool, dockerCfgDir string) error {
 	if err := c.Load(false); err != nil {
 		return err
 	}
@@ -63,17 +65,9 @@ func Apply(c *Context, force bool, dockercfg string) error {
 		}
 	}
 
-	if dockercfg != "" {
-		fmt.Printf("Detected dockercfg: `%s`\n", dockercfg)
-		instancePath := fmt.Sprintf("%s/dockercfg", c.InstanceDir)
-
-		if err := CopyFile(dockercfg, instancePath); err != nil {
-			return fmt.Errorf("Failed to copy dockercfg: %v.", err)
-		}
-	}
-
-	// validate dockercfg; if file does not exist, write empty dockercfg file
-	if err := checkDockercfg(c, force); err != nil {
+	// Check for docker config file (config.json or dockercfg) and copy it to
+	// instance directory in `dockercfg` format.
+	if err = checkDockercfg(c, dockerCfgDir, force); err != nil {
 		return err
 	}
 
@@ -408,7 +402,7 @@ func Endpoint(c *Context, syntax string, insecure, dev, quiet bool) error {
 }
 
 func Plan(c *Context, args []string) error {
-	if err := checkDockercfg(c, false); err != nil {
+	if err := checkDockercfg(c, c.InstanceDir, false); err != nil {
 		return err
 	}
 
@@ -416,21 +410,47 @@ func Plan(c *Context, args []string) error {
 	return Terraform(c, args)
 }
 
-func checkDockercfg(c *Context, force bool) error {
-	path := fmt.Sprintf("%s/dockercfg", c.InstanceDir)
+// checkDockercfg looks for docker config file (config.json or dockercfg) from
+// given directory and saves in `dockercfg` format in instance directory.
+func checkDockercfg(c *Context, dockerCfgDir string, force bool) error {
+	if dockerCfgDir == "" {
+		dockerCfgDir = c.InstanceDir
+	} else {
+		fmt.Printf("Checking for docker config file at: `%s`\n", dockerCfgDir)
+	}
 
-	// if file does not exist
-	if _, err := os.Stat(path); err != nil {
+	// Load docker config file from given dockerCfgDir
+	configFile, err := LoadDockerConfig(dockerCfgDir)
+	if err != nil {
+		fmt.Printf("Could not load docker config from `%v`.\n", dockerCfgDir)
 
-		// write empty dockercfg file
-		if err := ioutil.WriteFile(path, []byte("{}"), 0660); err != nil {
-			return fmt.Errorf("Failed to write 'dockercfg': %v", err)
+		// Try global docker config directory
+		dockerCfgDir = filepath.Join(homedir.Get(), ".docker")
+		fmt.Printf("Attempting to load global docker config from `%v`.\n", dockerCfgDir)
+
+		configFile, err = LoadDockerConfig(dockerCfgDir)
+		if err != nil {
+			// Fallback: configFile will yield "empty" dockercfg ({})
+			configFile.Empty = true
+			fmt.Println("Unable to load global docker config.")
+		} else {
+			fmt.Println("Successfully loaded global docker config.")
 		}
+	}
 
-		// alert user
+	if err := configFile.Save(c.InstanceDir); err != nil {
+		return fmt.Errorf("Error saving dockercfg: %v", err)
+	}
+
+	if configFile.Empty == true {
+		fmt.Println("Created default 'dockercfg' file.")
+
 		if !force {
-			text := fmt.Sprintf("NOTICE: No 'dockercfg' file present at %s. \n", path)
-			text += "Created default 'dockercfg' file. \n"
+			path := fmt.Sprintf("%s/dockercfg", c.InstanceDir)
+
+			text := fmt.Sprintf("\nNOTICE: The 'dockercfg' file present at `%s` \n", path)
+			text += "does not contain any authentication information."
+			text += "\n"
 			text += "\n"
 			text += "If you require private registry authentication, please edit this file \n"
 			text += "with the required credentials before proceeding. \n"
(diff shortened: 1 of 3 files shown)


jonathanstallings:
Notes:
Some guidance found by reference to code within the docker repository around time when `config.json` support was added.
See: https://github.com/docker/docker/blob/18c9b6c6455f116ae59cde8544413b3d7d294a5e/registry/auth.go

Sections of `dockerauth.go` may be simplified and attribution to reference code may be added as needed.

setup/context/dockerauth.go
```diff
@@ -0,0 +1,116 @@
...
+	OLD_CONFIGFILE = "dockercfg"
+)
+
+// AuthConfig contains authorization information for connecting to a Registry.
+type AuthConfig struct {
+	Username string `json:"username,omitempty"`
+	Password string `json:"password,omitempty"`
+	Auth     string `json:"auth,omitempty"`
+
+	// Email is an optional value associated with the username.
```
zpatrick: we don't need these comments

setup/context/dockerauth.go
```diff
@@ -0,0 +1,116 @@
...
+type DockerConfigFile struct {
+	AuthConfigs map[string]AuthConfig `json:"auths"`
+	HTTPHeaders map[string]string     `json:"HttpHeaders,omitempty"`
+	filename    string                // Note: not serialized - for internal use only
+	Empty       bool                  // Note: not serialized - for internal use only
+}
+
+// LoadDockerConfig loads docker config files of config.json or dockercfg formats.
+func LoadDockerConfig(configDir string) (*DockerConfigFile, error) {
+	if configDir == "" {
```
zpatrick: remove param checking/validation - that's a python-esq pattern you typically don't do w/ go

setup/context/dockerauth.go
```diff
@@ -0,0 +1,116 @@
...
+	RegistryToken string `json:"registrytoken,omitempty"`
+}
+
+// DockerConfigFile contains `config.json` file info.
+// `dockercfg` files use just the `AuthConfigs` field.
+type DockerConfigFile struct {
+	AuthConfigs map[string]AuthConfig `json:"auths"`
+	HTTPHeaders map[string]string     `json:"HttpHeaders,omitempty"`
+	filename    string                // Note: not serialized - for internal use only
+	Empty       bool                  // Note: not serialized - for internal use only
```
zpatrick: don't need to comment on internal use, it's a given from lowercase fields

zpatrick: if `Empty` is private why is it capitalized?

setup/context/dockerauth.go
```diff
@@ -0,0 +1,116 @@
...
+	configFile := DockerConfigFile{
+		AuthConfigs: make(map[string]AuthConfig),
+		filename:    filepath.Join(configDir, CONFIGFILE),
+	}
+
+	// Try config.json file first
+	if _, err := os.Stat(configFile.filename); err == nil {
+		file, err := os.Open(configFile.filename)
+		if err != nil {
+			return &configFile, err
```
zpatrick: `return nil, err`

setup/context/dockerauth.go
```diff
@@ -0,0 +1,116 @@
...
+		return nil, fmt.Errorf("No configdir supplied.")
+	}
+
+	configFile := DockerConfigFile{
+		AuthConfigs: make(map[string]AuthConfig),
+		filename:    filepath.Join(configDir, CONFIGFILE),
+	}
+
+	// Try config.json file first
+	if _, err := os.Stat(configFile.filename); err == nil {
```
zpatrick: always check `err != nil`, and don't created nested blocks if you can help it. This should be:
```
if _, err := os.Stat(configFile.filename); err != nil {
    return nil, err
}

file, err := os.Open(configFile.filename)
...

```

setup/context/dockerauth.go
```diff
@@ -0,0 +1,116 @@
...
+	data, err := json.MarshalIndent(configFile.AuthConfigs, "", "\t")
+	if err != nil {
+		return err
+	}
+
+	if err := os.MkdirAll(filepath.Dir(saveDir), 0600); err != nil {
+		return err
+	}
+
+	err = ioutil.WriteFile(filepath.Join(saveDir, OLD_CONFIGFILE), data, 0600)
```
zpatrick: [nit] `if err := iotuil.WriteFile(...); err != nil{ ... }`

setup/main.go
```diff
@@ -76,15 +76,15 @@ func main() {
 	app.Command("apply", "Create/Update a Layer0", func(cmd *cli.Cmd) {
 		flags := loadFlags(cmd, []string{"access_key", "secret_key", "region"})
 
-		dockercfg := cmd.StringOpt("dockercfg", "", "Path to valid dockercfg file")
+		dockerCfgDir := cmd.StringOpt("docker-config-dir", "", "Path to directory containing valid dockercfg or config.json file")
```
zpatrick: I think this should stay as a path to a file:
```
dockercfg := cmd.StringOpt("dockercfg", "", "Path to valid dockercfg or config.json file")
```

setup/context/commands.go
```diff
@@ -63,17 +65,9 @@ func Apply(c *Context, force bool, dockercfg string) error {
...
-		if err := CopyFile(dockercfg, instancePath); err != nil {
-			return fmt.Errorf("Failed to copy dockercfg: %v.", err)
-		}
-	}
-
-	// validate dockercfg; if file does not exist, write empty dockercfg file
-	if err := checkDockercfg(c, force); err != nil {
+	// Check for docker config file (config.json or dockercfg) and copy it to
+	// instance directory in `dockercfg` format.
+	if err = checkDockercfg(c, dockerCfgDir, force); err != nil {
```
zpatrick: [nit] don't need to propagate `force` behavior across too many function if we can avoid it:
```
if !force {
    if err := checkDockercfg(c, dockerCfgDir); err != nil{
        return err
    }
}
```

setup/context/commands.go
```diff
@@ -408,29 +402,55 @@ func Endpoint(c *Context, syntax string, insecure, dev, quiet bool) error {
...
 	args = append([]string{"plan"}, args...)
 	return Terraform(c, args)
 }
 
-func checkDockercfg(c *Context, force bool) error {
-	path := fmt.Sprintf("%s/dockercfg", c.InstanceDir)
+// checkDockercfg looks for docker config file (config.json or dockercfg) from
+// given directory and saves in `dockercfg` format in instance directory.
+func checkDockercfg(c *Context, dockerCfgDir string, force bool) error {
+	if dockerCfgDir == "" {
```
zpatrick: we don't use default params like this in go, this logic should exist in the caller

setup/context/commands.go
```diff
@@ -408,29 +402,55 @@ func Endpoint(c *Context, syntax string, insecure, dev, quiet bool) error {
...
+		dockerCfgDir = c.InstanceDir
+	} else {
+		fmt.Printf("Checking for docker config file at: `%s`\n", dockerCfgDir)
+	}
+
+	// Load docker config file from given dockerCfgDir
+	configFile, err := LoadDockerConfig(dockerCfgDir)
```
zpatrick: For the sake of keeping things simple, I don't think we need this "discovery" behavior. 
Specifying a file path via `--dockercfg=path/to/file` is all we need to implement.

setup/context/dockerauth.go
jonathanstallings: Removed.

setup/context/dockerauth.go
jonathanstallings: Comments removed.

jonathanstallings: Good catch. Changed.

setup/context/dockerauth.go
jonathanstallings: 👍

setup/context/commands.go
jonathanstallings: I was under the impression that the `--force` flag was only intended to skip prompting on missing `dockercfg` file. However, the "empty" `dockercfg` file still needs to be created if missing, correct?

This suggestion would skip creating that file if missing.

setup/context/commands.go
jonathanstallings: I have seen this pattern in go source code, and I was modeling from it. Definitely will change if you prefer it differently, as I want to set this up the proper way. Thanks for the tip.

Moving default logic to caller.

jonathanstallings:
I'd be very happy to pair on this @zpatrick. Thanks for the offer!

setup/context/dockerauth.go
jonathanstallings: Removed.

setup/context/commands.go
jonathanstallings: Removed discovery behavior and reverted to specifying file path explicitly.

setup/context/commands.go
diemonster: @jonathanstallings I think zpatrick is just saying that passing `force` as a param/behavior around to other functions, even if they are helpers, is not ideal.

setup/main.go
jonathanstallings: Changed to file path as requested.

setup/context/dockerauth.go
jonathanstallings: Changed to `err != nil` pattern. Thanks!

setup/context/dockerauth.go
jonathanstallings: Using the the existing pattern facilitates saving an "empty" config. However, I can certainly change this.

diemonster:
@zpatrick @jonathanstallings 

Re: ripping out the "discovery behavior". I'm a bit torn on this, but it's a workflow issue more general to all of our CLI UX. 

We consistently expect the user to know about specific flags to circumvent default behavior, but in some cases (like this) our default behavior is not the most common use case. Beyond that, we push the burden of trial and error onto the user and force them to muck through our tools until they get the perfect combination of arguments. This is frustrating for many users, and I'm worried it will eventually bite us if Layer0 is adopted beyond our close confines. 

Zpatrick's suggestion is in line with how the rest of our CLI UX works, so in that sense, I agree w his comment _right now_. But thinking from a bigger picture, we need to consider adding actual logic in our CLI tools to prevent users from giving up too quickly. `l0-setup` for all of it's deficiencies, is the introduction to our entire toolset for new people, and we need to put a friendlier foot forward.

zpatrick:
closing for different branch

PR #94, opened 2017-02-09
fix for terraform plugin build failure

tlake:
The plugin now does a proper comparison between the health check object received from `expandHealthCheck` and an empty initialized health check object. If they're equal (i.e. the health check field was completely unspecified in the `layer0.tf` file), the model is populated with our default values.

### DIFF
--- a/plugins/terraform/resource_load_balancer.go
+++ b/plugins/terraform/resource_load_balancer.go
@@ -112,8 +112,8 @@ func resourceLayer0LoadBalancerCreate(d *schema.ResourceData, meta interface{}) 
 	ports := expandPorts(d.Get("port").(*schema.Set).List())
 	healthCheck := expandHealthCheck(d.Get("health_check"))
 
-	if !healthCheck {
-		healthCheck = models.HealthCheck{
+	if healthCheck == nil {
+		healthCheck = &models.HealthCheck{
 			Target:             "TCP:80",
 			Interval:           30,
 			Timeout:            5,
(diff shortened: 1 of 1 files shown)


plugins/terraform/resource_load_balancer.go
```diff
@@ -112,7 +112,8 @@ func resourceLayer0LoadBalancerCreate(d *schema.ResourceData, meta interface{})
 	ports := expandPorts(d.Get("port").(*schema.Set).List())
 	healthCheck := expandHealthCheck(d.Get("health_check"))
 
-	if !healthCheck {
+	blankCheck := models.HealthCheck{}
```
diemonster: correct me if i'm wrong @zpatrick but I think the suggestion was something like
```
blankCheck := &models.HealthCheck{}
if healthCheck == *blankCheck {
```

tlake: I don't think I'm catching the difference between this recommended
```
blankCheck := &models.HealthCheck{}
if healthCheck == *blankCheck {
```
and my own
```
blankCheck := models.HealthCheck{}
if healthCheck == blankCheck {
```

plugins/terraform/resource_load_balancer.go
```diff
@@ -112,7 +112,7 @@ func resourceLayer0LoadBalancerCreate(d *schema.ResourceData, meta interface{})
 	ports := expandPorts(d.Get("port").(*schema.Set).List())
 	healthCheck := expandHealthCheck(d.Get("health_check"))
 
-	if !healthCheck {
+	if healthCheck == (models.HealthCheck{}) {
```
zpatrick: This doesn't seem very canonical - could you update `expandHealthCheck` to return  a `*models.HealthCheck` and check `if healthcheck == nil` here?

tlake: ✅

scripts/tlakemakefile
```diff
@@ -0,0 +1,20 @@
+mocks:
```
zpatrick: forgot a .gitignore perhaps?

tlake: Ah shoot :p ✅

zpatrick:
Going to move this into `develop` instead of `hotfix->master`. Reason being - we want these changes to be in `v0.9-rc2` and/or `v0.9` release

PR #95, opened 2017-02-09
add text printer unit tests

zpatrick:


cli/printer/text_test.go
```diff
@@ -0,0 +1,397 @@
+package printer
+
+import (
+	"github.com/quintilesims/layer0/common/models"
+	"github.com/quintilesims/layer0/common/types"
+	"time"
+)
+
+// testing stdout: https://blog.golang.org/examples
```
tlake: Oh, this is interesting. Not sure I feel superb about comments being used as test assertions, but it works, and it doesn't have anything to do with you.

PR #96, opened 2017-02-10
Allow placementConstraints in task definition file (#83)

### LINKED ISSUE #83: Allow placementConstraints to be added in the task definition file
Adding this would open up a whole world of task placement strategies. Although I realize it may not really fit with Layer0, it would just be exposing functionality that already exists in AWS ECS and provide additional flexibility for outlying use cases.

http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_TaskDefinitionPlacementConstraint.html

tlake:
Closes #83

- Placement constraints may now be added to task definition
(Dockerrun.aws.json) files. See
http://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-placement-constraints.html
for more information.

- Updated `aws` dependencies as a by-product of updating the `aws-go-sdk`
library, which is necessary for the use of placement constraints.


common/aws/ecs/ecs.go
```diff
@@ -344,12 +348,18 @@ func (this *ECS) RegisterTaskDefinition(family string, roleARN string, networkMo
 		networkModep = aws.String(networkMode)
 	}
 
+	awsPlacementConstraints := make([]*ecs.TaskDefinitionPlacementConstraint, 0)
```
zpatrick: [nit] this is more of a personal preference thing, but I've created quite a few bugs when mixing `make(object, count)` with `append`. I usually stay away from `make` unless I know I'm going to use `slice[i] = object` and not `append`.

PR #103, opened 2017-02-16
Release v0.9.0-rc2

tlake:


tlake:
Do we know why our builds keep having trouble with not deleting old vpcs?

tlake:
Alright, cool - all tests passed, but without actually having identified the reason for their previous failures. I ended up just manually deleting the couple of resources that seemed to continue existing after previous runs' destroy steps.

PR #118, opened 2017-03-08
Tests for the terraform plugin (#97)

### LINKED ISSUE #97: tests for the terraform plugin

tlake:
Closes #97


tlake:
_Notes:_

- These tests are so far acceptance tests, and they create and destroy actual resources on AWS. They also take a while to run so perhaps they should only be run within the release pipeline?

- These acceptance tests seem fragile at the moment - if the tests are interrupted, artefacts and resources are left dangling on AWS. While developing locally, this problem is fixed by running `l0 environment delete <testing_env_name>`, but this could be a pain point if something goes wrong inside our pipeline.

zpatrick:
> These tests are so far acceptance tests, and they create and destroy actual resources on AWS. They also take a while to run so perhaps they should only be run within the release pipeline?


We only need to test that the correct requests are being sent to the layer0 api. Is there a way to mock out the client?

tlake:
> We only need to test that the correct requests are being sent to the layer0 api. Is there a way to mock out the client?

I think so - that was my original plan, before we discovered the patterns hashicorp used for their plugins. I can reorient in that direction.

plugins/terraform/provider.go
```diff
@@ -1,15 +1,15 @@
...
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/terraform"
 	"github.com/quintilesims/layer0/cli/client"
+	"time"
 )
 
 var defaultTimeout = time.Minute * 15
 
-func Provider() *schema.Provider {
+func Provider() terraform.ResourceProvider {
```
zpatrick: Is this change required? Wouldn't this cause the `TestProvider` function to fail at `if err := Provider().(*schema.Provider)`?

plugins/terraform/resource_environment_test.go
```diff
@@ -0,0 +1,140 @@
+package main
+
+import (
+	"github.com/golang/mock/gomock"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/quintilesims/layer0/common/models"
+	"testing"
+)
+
+func TestEnvironmentCreate_basic(t *testing.T) {
```
zpatrick: [nit] let's be explicit that we are testing for `_defaults` here

plugins/terraform/resource_environment_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+func TestEnvironmentUpdate(t *testing.T) {
+	ctrl, mockClient, provider := setupUnitTest(t)
+	defer ctrl.Finish()
+
+	gomock.InOrder(
+		mockClient.EXPECT().
+			CreateEnvironment("test-env", "m3.medium", 0, []byte("")).
+			Return(&models.Environment{EnvironmentID: "eid"}, nil),
+		// Return(&createdEnvironment, nil),
```
zpatrick: remove commented code

plugins/terraform/resource_environment_test.go
```diff
@@ -0,0 +1,140 @@
...
+		// Return(&updatedEnvironment, nil),
+
+		mockClient.EXPECT().
+			GetEnvironment("eid").
+			Return(&models.Environment{EnvironmentID: "eid"}, nil),
+		// Return(&updatedEnvironment, nil),
+	)
+
+	environmentResource := provider.ResourcesMap["layer0_environment"]
+	d1 := schema.TestResourceDataRaw(t, environmentResource.Schema, map[string]interface{}{
```
zpatrick: [nit] I assume you are doing the "create" steps first so that the `d.HasChange()` functions work correctly. There's probably a more explicit way to do this (maybe using [`d.Set()`](https://github.com/hashicorp/terraform/blob/master/helper/schema/resource_data.go#L150)?).

plugins/terraform/resource_load_balancer_test.go
```diff
@@ -0,0 +1,250 @@
+package main
+
+import (
+	"github.com/golang/mock/gomock"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/quintilesims/layer0/common/models"
+	"testing"
+)
+
+func TestLoadBalancerCreate_basic(t *testing.T) {
```
zpatrick: `_basic` -> `_defaults`

plugins/terraform/resource_load_balancer_test.go
```diff
@@ -0,0 +1,250 @@
...
+	defer ctrl.Finish()
+
+	loadBalancerResource := provider.ResourcesMap["layer0_load_balancer"]
+	d := schema.TestResourceDataRaw(t, loadBalancerResource.Schema, map[string]interface{}{
+		"name":        "test-lb",
+		"environment": "test-env",
+	})
+
+	mockClient.EXPECT().
+		CreateLoadBalancer("test-lb", "test-env", models.HealthCheck{"TCP:80", 30, 5, 2, 2}, []models.Port{}, true).
```
zpatrick: Will the layer0 api fail this call if we send an empty `[]models.Port{}`? I don't recall the layer0 api setting a default.

plugins/terraform/resource_load_balancer_test.go
```diff
@@ -0,0 +1,250 @@
...
+	ports := []models.Port{
+		{
+			HostPort:      80,
+			ContainerPort: 80,
+			Protocol:      "http",
+		},
+		{
+			HostPort:      8080,
+			ContainerPort: 8001,
+			Protocol:      "http",
```
zpatrick: one of these ports should be `https` with a `certificate` field set.

plugins/terraform/resource_load_balancer_test.go
```diff
@@ -0,0 +1,250 @@
...
+		},
+	}
+
+	loadBalancerResource := provider.ResourcesMap["layer0_load_balancer"]
+	d := schema.TestResourceDataRaw(t, loadBalancerResource.Schema, map[string]interface{}{
+		"name":        "test-lb",
+		"environment": "test-env",
+		"port": flattenPorts([]models.Port{
```
zpatrick: can you do `"port": flattenPorts(ports),`?

plugins/terraform/resource_load_balancer_test.go
```diff
@@ -0,0 +1,250 @@
...
+	ctrl, mockClient, provider := setupUnitTest(t)
+	defer ctrl.Finish()
+
+	loadBalancerResource := provider.ResourcesMap["layer0_load_balancer"]
+	d := schema.TestResourceDataRaw(t, loadBalancerResource.Schema, map[string]interface{}{
+		"name":        "test-lb",
+		"environment": "test-env",
+	})
+
+	d.Set("health_check", flattenHealthCheck(models.HealthCheck{
```
zpatrick: [nit] can this not me added along with the other raw data in line 86?

plugins/terraform/resource_service_test.go
```diff
@@ -0,0 +1,159 @@
...
+		mockClient.EXPECT().
+			CreateService("test-svc", "test-env", "test-dep", "test-lb").
+			Return(&models.Service{ServiceID: "sid"}, nil),
+
+		mockClient.EXPECT().
+			ScaleService("sid", 2).
+			Return(&models.Service{ServiceID: "sid"}, nil),
+
+		mockClient.EXPECT().
+			WaitForDeployment("sid", time.Minute*15).
```
zpatrick: [nit] let's use `gomock.Any()` for timeout

plugins/terraform/provider.go
tlake: It actually fails at that line (19) and at 13 before it when it's `*schema.Provider`:
```
invalid type assertion: Provider().(*schema.Provider) (non-interface type *schema.Provider on left)
```
If I strip out the type assertions, it looks like things work.

I guess it's because `schema.Provider` (_hashicorp/terraform/helper/schema/provider.go_) is type struct, whereas `terraform.ResourceProvider` (_hashicorp/terraform/terraform/resource_provider.go_) is type interface?

Do we have a preference?

plugins/terraform/resource_environment_test.go
tlake: Good call. ✅

plugins/terraform/resource_environment_test.go
tlake: ✅

plugins/terraform/resource_load_balancer_test.go
tlake: ✅

plugins/terraform/resource_load_balancer_test.go
tlake: ✅

plugins/terraform/resource_load_balancer_test.go
tlake: Yep! Just noticed that myself, too. :) ✅

plugins/terraform/resource_load_balancer_test.go
tlake: Totally can ✅

plugins/terraform/resource_service_test.go
tlake: I like it. ✅

plugins/terraform/provider.go
zpatrick: I'd say do whatever the builtin terraform providers do. Pending that, best practice in go when building constructors is to return the object literal, not an interface - so I'd expect `func Provider() *schema.Provider`

tlake: Yeah, it looks like they use `terraform.ResourceProvider`.

plugins/terraform/resource_environment_test.go
tlake: Yeah, I was having some issues finding the best way to trigger a change.

At one point, I had something like this:

```
	environmentResource := provider.ResourcesMap["layer0_environment"]
	d := schema.TestResourceDataRaw(t, environmentResource.Schema, map[string]interface{}{
		"name": "test-env",
	})

	if err := environmentResource.Create(d1, mockClient); err != nil {
		t.Fatal(err)
	}

	fmt.Printf("has change: %t\n", d1.HasChange("min_count"))
	// false, as expected

	d.Set("min_count", 3)
	d.SetId("eid")

	fmt.Printf("has change: %t\n", d1.HasChange("min_count"))
	// false, which is NOT expected

	if err := environmentResource.Update(d, mockClient); err != nil {
		t.Fatal(err)
	}
```

It seemed to me that setting up two different `TestResourceDataRaw` was the way to trigger `HasChange` updates, but it's entirely possible that I've overlooked or just not found the actual right way to do it.

plugins/terraform/resource_load_balancer_test.go
tlake: Oooh, good call - I can update this with a specified port, but bigger-picture: should I be making CLI-based failure tests? They seem a little useless to me, since the client's mocked out.

 - send off a config that's wrong
 - tell the mockClient to return something saying that it's wrong
 - assert that... I've told the mockClient to say that something's wrong...?

zpatrick: I guess better question is: do we have anything in the terraform schema or functions that check if at least 1 port is set? Not sure if setting "required: true" works for these complex types.

On a larger note

plugins/terraform/resource_environment_test.go
zpatrick: No worries, we can change it later if we figure out a cleaner way to set the state changes (if there even is one)

zpatrick:
Looks good to me!

plugins/terraform/resource_load_balancer_test.go
tlake: Locally, when I run `terraform apply` on a `.tf` file that specifies a load balancer without ports, it errors for me with
```
Errors:

  * layer0_load_balancer.health-lb: "port": required field is not set
```

It seems like terraform is handling the `required: true` properly.

PR #124, opened 2017-03-14
add initial Windows documentation / support

diemonster:


sesh-kebab:
Looks good. The recommended size in the Amazon walkthrough for volumes seems to be 200Gb. But 128Gb if probably fine too I guess.

diemonster:
@sesh-kebab Ahhh good call, if they recommend 200gb I will change.

sesh-kebab:
Also, I think it would be better to pass in the userdata when creating the environment to avoid having to manually copy the cluster name over. I.e.,

```
<powershell>
## The string 'windows' should be replaced with your cluster name

# Set agent env variables for the Machine context (durable)
$clusterName = {{ .ECSEnvironmentID }}
Write-Host Cluster name set as: $clusterName -foreground green

[Environment]::SetEnvironmentVariable("ECS_CLUSTER", $clusterName, "Machine")
```

zpatrick:
Would it be better to add a `windows` example to https://github.com/quintilesims/layer0-examples?

diemonster:
@zpatrick not a bad idea, it does seem strange to put this where it currently sits. I'll move the stuff and close this PR.

Issue #104, opened 2017-02-21
Make the scheduler stateless


zpatrick:
The scheduler is currently a stateful component of the api. This has led to bugs with users who scale the api > 1. We need to update this component to make the api a stateless service. 

The "scheduler" is a relatively high-level concept - more concretely, it's the relationship between the the cluster scalar, task scheduler, and right sizer. Luckily, we can use many existing components to make the scheduler stateless. 

I think the following architecture changes can provide us with a stateless scheduler with the minimum amount of refactoring:

## Tasks 
Remove the task scheduler and run tasks as jobs. In the same way the ecs service scheduler works, the "Run Task" job would simply try and retry running the task until there is adequate resources in the environment. 

## Scalar
This would be a combination of the right sizer (which scales clusters down) and the cluster scalar (which scales clusters up). Simply put, the scalar runs a single, stateless function that is very similar to `OptimizeCluster` in the current right sizer:
* Gather the resources in the environment - the number of **desired** (not current!) instances in the autoscaling group. 
* Gather the required resources in the environment - the number of running services and tasks + the number of pending services and tasks.
* Make logical decisions on whether to scale the autoscaling group up, down, or do nothing. 
* We would need a simple locking mechanism to enforce that the Scalar function never runs in parallel.

diemonster:
Regarding schedule improvements, would it be worth investigating this first? https://blox.github.io/

PR #145, opened 2017-03-22
Scheduler refactor

zpatrick:
closes https://github.com/quintilesims/layer0/issues/104


api/scheduler/environment_scaler_test.go
```diff
@@ -0,0 +1,346 @@
...
+}
+
+type EnvironmentScalerUnitTest struct {
+	ExpectedScale     int
+	MemoryPerProvider bytesize.Bytesize
+	ResourceProviders []*resource.ResourceProvider
+	ResourceConsumers []resource.ResourceConsumer
+}
+
+func (m *MockProviderManager) AddNewProvider(environmentID string) (*resource.ResourceProvider, error) {
```
zpatrick: should switch the order of these declarations

api/scheduler/environment_scaler_test.go
```diff
@@ -0,0 +1,346 @@
...
+			resource.NewResourceProvider("", true, bytesize.MB*3, nil),
+		},
+		ResourceConsumers: []resource.ResourceConsumer{
+			// these 4 consumers can be placed in the current cluster
+			{Memory: bytesize.MB},
+			{Memory: bytesize.MB},
+			{Memory: bytesize.MB},
+			{Memory: bytesize.MB * 2},
+			// these 2 consumers will require a new provider
+			{Memory: 2 * bytesize.MB},
```
zpatrick: should be consistent and use `bytesize.MB *2`

api/scheduler/environment_scaler_test.go
```diff
@@ -0,0 +1,346 @@
...
+		MemoryPerProvider: bytesize.MB * 2,
+		ResourceProviders: []*resource.ResourceProvider{
+			resource.NewResourceProvider("", true, bytesize.MB, []int{80}),
+			resource.NewResourceProvider("", true, bytesize.MB*2, []int{80}),
+		},
+		ResourceConsumers: []resource.ResourceConsumer{
+			// this consumer will require a new provider for ports
+			{Memory: bytesize.MB, Ports: []int{80}},
+			// this consumer wil require a new provider for memory
+			{Memory: bytesize.MB * 2},
```
zpatrick: I don't think this is correct, it may be able to be placed in the 2nd provider

diemonster:
This is awesome but un-reviewable. Can you please break it into chunks? Otherwise we will need to sit with it together and go through the whole thing b/c it's just too big to navigate otherwise.

api/handlers/task_handler.go
```diff
@@ -119,13 +122,13 @@ func (this *TaskHandler) CreateTask(request *restful.Request, response *restful.
 		return
 	}
 
-	task, err := this.TaskLogic.CreateTask(req)
+	job, err := this.JobLogic.CreateJob(types.CreateTaskJob, req)
```
diemonster: Major change is that we're creating a job executed by the `runner` instead of executing the task right then and there

runner/job/task.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"encoding/json"
+	log "github.com/Sirupsen/logrus"
+	"github.com/quintilesims/layer0/api/backend/ecs"
+	"github.com/quintilesims/layer0/common/models"
+	"time"
+)
+
+var DeleteTaskSteps = []Step{
```
diemonster: this was missing previously, when we deleted an `environment` before, `tasks` were not also cleared

api/backend/ecs/resource_manager.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	inUse := pint64(instance.PendingTasksCount)+pint64(instance.RunningTasksCount) > 0
+	provider := resource.NewResourceProvider(instanceID, inUse, availableMemory, usedPorts)
+
+	r.logger.Debugf("Environment '%s' generated provider: %#v\n", ecsEnvironmentID, provider)
+	return provider, true
+}
+
+func (r *ECSResourceManager) AddNewProvider(environmentID string) (*resource.ResourceProvider, error) {
```
diemonster: some clarification on this would be helpful

api/scheduler/environment_scaler.go
```diff
@@ -0,0 +1,130 @@
...
+		}
+
+		if !hasRoom {
+			newProvider, err := providerManager.AddNewProvider(environmentID)
+			if err != nil {
+				return nil, err
+			}
+
+			if !newProvider.HasResourcesFor(consumer) {
+				err := fmt.Errorf("Resource '%s' cannot fit into an empty provider!", consumer.ID)
```
diemonster: let's change this error message to be something more meaningful for the end-user... i.e. This Environment Instance Type is Too Small....

api/scheduler/environment_scaler.go
```diff
@@ -0,0 +1,130 @@
...
+	// check if we need to scale down
+	unusedProviders := []*resource.ResourceProvider{}
+	for i := 0; i < len(providers); i++ {
+		if !providers[i].IsInUse() {
+			unusedProviders = append(unusedProviders, providers[i])
+		}
+	}
+
+	desiredScale := len(providers) - len(unusedProviders)
+	actualScale, err := providerManager.ScaleTo(environmentID, desiredScale, unusedProviders)
```
diemonster: `unusedProviders` is a list that has the empty resource(s) that can be scaled down

api/backend/ecs/resource_manager.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+func (r *ECSResourceManager) getResourceProvider(ecsEnvironmentID id.ECSEnvironmentID, instance *ecs.ContainerInstance) (*resource.ResourceProvider, bool) {
+	instanceID := pstring(instance.Ec2InstanceId)
+	if pstring(instance.Status) != "ACTIVE" {
+		r.logger.Warnf("Instance %d is not in the 'ACTIVE' state", instanceID)
+		return nil, false
+	}
+
+	// todo: what if instance.AgentConnected == false?
```
diemonster: Let's log that the provider is not available and let it sit (in the case of Splunk or other long-running services)

api/backend/ecs/resource_manager.go
diemonster: calcNewProvider?

common/aws/ec2/ec2.go
```diff
@@ -43,47 +44,48 @@ type EC2Internal interface {
...
-	"i2.xlarge":   nil,
-	"i2.2xlarge":  nil,
-	"i2.4xlarge":  nil,
-	"i2.8xlarge":  nil,
-	"d2.xlarge":   nil,
-	"d2.2xlarge":  nil,
-	"d2.4xlarge":  nil,
-	"d2.8xlarge":  nil,
+var InstanceSizes = map[string]bytesize.Bytesize{
+	"t2.nano":     0.5 * bytesize.GiB,
```
diemonster: this is interesting... what's the reason we need to track the (memory?) size of each instance?

zpatrick: `CalculateNewProvider` function for the resource manager

common/models/resource_provider.go
```diff
@@ -0,0 +1,8 @@
+package models
+
+type ResourceProvider struct {
+	ID              string `json:"id"`
+	InUse           bool   `json:"in_use"`
+	UsedPorts       []int  `json:"used_ports"`
```
diemonster: just noticed this is a slice of `int`s... we are not tracking protocol type here, are we? Unless I'm missing something, this would disallow the same port if used by UDP and TCP, which is the case for Consul for example (8500 tcp and udp are needed)

runner/job/context.go
```diff
@@ -2,71 +2,46 @@ package job
...
+func NewJobContext(jobID string, lgc *logic.Logic, request string) *JobContext {
+	return &JobContext{
 		jobID:             jobID,
-		logic:             lgc,
-		loadBalancerLogic: logic.NewL0LoadBalancerLogic(*lgc),
-		serviceLogic:      logic.NewL0ServiceLogic(*lgc),
-		environmentLogic:  logic.NewL0EnvironmentLogic(*lgc),
 		request:           request,
-		mutex:             &sync.Mutex{},
+		Logic:             lgc,
```
diemonster: still meant to be `lgc`?

api/main.go
```diff
@@ -132,19 +137,42 @@ func main() {
...
+	go runEnvironmentScaler(environmentLogic)
 
 	logrus.Infof("Starting Job Janitor")
 	jobJanitor.Run()
 
 	logrus.Print("Service on localhost" + port)
 	logrus.Fatal(http.ListenAndServe(port, nil))
 }
+
+func runEnvironmentScaler(environmentLogic *logic.L0EnvironmentLogic) {
```
diemonster: essentially the new Right Sizer, sleeps for 5 mins in a separate goroutine and then scales things accordingly

api/backend/ecs/resource_manager.go
```diff
@@ -0,0 +1,210 @@
...
+		return nil, false
+	}
+
+	if !pbool(instance.AgentConnected) {
+		r.logger.Errorf("Instance %d agent is disconnected")
+		return nil, false
+	}
+
+	// this is non-intuitive, but the ports being used by tasks are kept in
+	// instance.ReminaingResources, not instance.RegisteredResources
```
diemonster: typo

api/logic/environment_resource_getter_test.go
```diff
@@ -0,0 +1,373 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	return string(bytes)
+}
+
+var deployWithOneContainer []byte = []byte(`
+{
+  "containerDefinitions": [
```
diemonster: maybe add a test case that specifies `udp` as a protocol

runner/job/context.go
zpatrick: yep

Issue #114, opened 2017-03-03
Make system test framework a standalone product


zpatrick:
This tool could be useful for scopes outside of layer0. 
If we gave it its own repo and refactored it to be a little more generic, it could be a very useful open source tool for system testing.

Issue #105, opened 2017-02-21
Performance Tests


tlake:
- [ ] Stress Tests
- [ ] DOS Tests

zpatrick:
I'm noticing that layer0 doesn't handle task stress testing very well. Would be useful to verify we can run ~100 tasks at once.

PR #146, opened 2017-03-23
Use tftest repo

zpatrick:
This refactors the system tests to use the standalone tftest repo: https://github.com/quintilesims/tftest.
It also adds the following:

* Merge service scale up/down into a single test
* Add a `clients` package, which wraps the layer0 and sts clients that just calls `t.Fatal` when an error occurs
* Use terraform outputs to get required variables

Closes https://github.com/quintilesims/layer0/issues/114
Closes https://github.com/quintilesims/layer0/issues/105


tests/system/taskPerformance_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/Sirupsen/logrus"
+	"github.com/quintilesims/layer0/common/testutils"
+	"testing"
+	"time"
+)
+
+// Test Resources:
+// This test creates an environment named 'tp'
+// and a deploy named 'apline'
+func TestTaskPerformance(t *testing.T) {
```
tlake: I like this test a lot. :+1:

common/testutils/functions.go
```diff
@@ -41,3 +42,13 @@ func AssertInSlice(t *testing.T, expected, slice interface{}) {
 		slice,
 		reflect.TypeOf(slice))
 }
+
+func WaitFor(t *testing.T, interval, timeout time.Duration, conditionSatisfied func() bool) {
```
tlake: This is a cool little pattern.

Issue #152, opened 2017-03-24
automate windows environment creation in Layer0 -- API


tribaljack:
Semantic:
> l0 environment create --os=windows <env name>

Do we need to worry about minimum ec2 instance type for windows? given the larger base image?

andycmaj:
so here's the part in userdata.ps1 that ended up working for private docker auth:

```
...

# Configure docker auth
# TODO: temporary hack
$dockercfgContent = '{"d.ims.io":{"auth":"<***PRIVATE_REG_AUTH_TOKEN***>","email":""}}';
[Environment]::SetEnvironmentVariable("ECS_ENGINE_AUTH_DATA", $dockercfgContent, "Machine")
[Environment]::SetEnvironmentVariable("ECS_ENGINE_AUTH_TYPE", "dockercfg", "Machine")

# Output all environment variables
Get-ChildItem -Path Env:* | Sort-Object Name

...
```

Few notes about the actual solution:

* AWS Powershell tools are already installed on the AMI, so you can do `Copy-S3Object -BucketName "{{ .S3Bucket }}" -Key "bootstrap/dockercfg" -LocalFile "dockercfg"`

diemonster:
@andycmaj nice! I'll add this to the hacky instructions for now. thanks!

Issue #158, opened 2017-03-27
l0 environment create --ami


tribaljack:
enable users to create environment using an AMI, which should significantly reduce the time to create an environment during development process, no need to download 8GB windows image with each test pass.

PR #167, opened 2017-03-28
Include fields for operating system and ami id in environment create on API

zpatrick:
closes https://github.com/quintilesims/layer0/issues/152
(partially closes) https://github.com/quintilesims/layer0/issues/158


api/logic/environment_logic.go
```diff
@@ -85,7 +86,17 @@ func (e *L0EnvironmentLogic) CreateEnvironment(req models.CreateEnvironmentReque
...
-	environment, err := e.Backend.CreateEnvironment(req.EnvironmentName, req.InstanceSize, req.MinClusterCount, req.UserDataTemplate)
+	if req.OperatingSystem == "" {
+		return nil, errors.Newf(errors.MissingParameter, "OperatingSystem is required")
+	}
+
+	environment, err := e.Backend.CreateEnvironment(
+		req.EnvironmentName,
+		req.InstanceSize,
+		req.OperatingSystem,
+		req.AMIID,
```
diemonster: [nit] I guess the guideline here (https://msdn.microsoft.com/en-us/library/141e06ef(v=vs.71).aspx) would be to use `AmiID`. although that looks even weirder

common/config/config.go
```diff
@@ -123,22 +126,16 @@ func SetCLIVersion(version string) {
 	cliVersion = version
 }
 
-var serviceAMIs = map[string]string{
-	"us-west-2": "ami-6cb9ac0d",
```
diemonster: are we no longer hardcoding default AMIs?

api/backend/ecs/environment_manager.go
```diff
@@ -374,7 +397,75 @@ func renderUserData(ecsEnvironmentID id.ECSEnvironmentID, userData []byte) (stri
...
+Read-S3Object -BucketName {{ .S3Bucket }} -Key bootstrap/dockercfg -File dockercfg.json
+$dockercfgContent = [IO.File]::ReadAllText("dockercfg.json")
+[Environment]::SetEnvironmentVariable("ECS_ENGINE_AUTH_DATA", $dockercfgContent, "Machine")
+[Environment]::SetEnvironmentVariable("ECS_ENGINE_AUTH_TYPE", "dockercfg", "Machine")
+
+# Output all environment variables
+Get-ChildItem -Path Env:* | Sort-Object Name
+
+# Output all environment variables
+Get-ChildItem -Path Env:* | Sort-Object Name
```
zpatrick: fixed

common/config/config.go
zpatrick: this was actually unused. it duplicated what was done in l0-setup

Issue #156, opened 2017-03-27
automate windows environment creation in Layer0 -- Terraform Plug-ins


tribaljack:
Semantics:

```
resource "layer0_environment" "dev" {
  name = "dev"
  os   = "windows"
  size = "m3.medium"
}
```

PR #175, opened 2017-03-30
Add --ami option to cli.  Add os and ami options to terraform plugin

zpatrick:
Closes https://github.com/quintilesims/layer0/issues/158
Closes https://github.com/quintilesims/layer0/issues/156


cli/command/environment_command.go
```diff
@@ -46,6 +46,10 @@ func (e *EnvironmentCommand) GetCommand() cli.Command {
 						Value: "linux",
 						Usage: "specifies windows or linux container environment cluster",
 					},
+					cli.StringFlag{
+						Name:  "ami",
+						Usage: "specifies a custom AMI to use in the environment",
```
tlake: Maybe we should specify that the user needs to input an AMI ID, in case the ambiguity inspires somebody to use an AMI name?

docs-src/docs/reference/cli.md
```diff
@@ -213,6 +213,14 @@ Use the **create** subcommand to create an additional Layer0 environment (_envir
     <div class="divCellNoWrap">--user-data</div>
     <div class="divCell">The user data template to use for the environment's autoscaling group.</div>
   </div>
+  <div class="divRow">
+    <div class="divCellNoWrap">--os</div>
+    <div class="divCell">The operating system used in the environment. Options are "linux" or "windows" (default: linux).</div>
+  </div>
+  <div class="divRow">
+    <div class="divCellNoWrap">--ami</div>
+    <div class="divCell">The AMI to use in the environment. If not specified, Layer0 will use the default AMI for the specified operating system.</div>
```
tlake: Again, maybe we should be explicit about AMI ID vs name.

docs-src/docs/reference/terraform-plugin.md
```diff
@@ -141,6 +143,10 @@ Available instance sizes can be found [here](https://aws.amazon.com/ec2/instance
 * `min_count` - (Optional, Default: 0) The minimum number of instances allowed in the environment
 * `user-data` - (Optional) The user data template to use for the environment's autoscaling group.
 See the [cli reference](/reference/cli/#environment) for the default template.
+* `os` - (Optional, Default: "linux") Specifies the type of operating system used in the environment.
+Options are "linux" or "windows".
+* `ami` - (Optional) Specifies the AMI to use in the environment.
```
tlake: Once more re: AMI ID vs name - I know down on line 160 it specifies "ID" but it probably wouldn't hurt to say so in a few places, and I don't think it's mentioned at all in the CLI reference.

cli/command/environment_command.go
diemonster: agreed w tlake, we should put an example of the AMI ID format, too just to be on the safe side

tests/system/windowsService_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/quintilesims/layer0/common/testutils"
+	"net/http"
+	"testing"
+	"time"
+)
+
+// Test Resources:
+// This test creates an environment named 'ws' that has a
+// Windows service named 'windows' running in iis
+func TestWindowsService(t *testing.T) {
```
diemonster: Nice, loving this pattern overall. Super readable.

tlake:
@zpatrick The tests seem to be failing because `sort.Slice` is undefined. I encountered this in a different branch which isn't going to be merged anymore, but I think we need to update the `.travis.yml` file to use the proper Go version.

```
...
language: go

go:
  - 1.8
...
```

zpatrick:
@tlake: updated .travis.yml. I also noticed I forgot to add `OS` column to cli output, so that is added in.

tlake:
Lookin' great and ready to 🚢  to me.

Issue #117, opened 2017-03-08
Migrate Layer0 Binaries to Github


diemonster:
We currently host released binaries in S3, but that isn't necessary with Github. Let's do the following:

- migrate our older releases (just v0.8.4+) to Github
- update the release script to push binaries to Github.

https://help.github.com/articles/creating-releases/

tlake:
We should also figure out a good solution to creating binaries/releases and pushing tags that allows for some good automation. As a minor example, needing to manually update version numbers for the badges in the README is annoying and liable to be forgotten.

Related to #183.

PR #181, opened 2017-03-30
Incorporate README Badges (#178)

### LINKED ISSUE #178: Incorporate README Badges
In a branch that's now-defunct, I had made changes to add some nifty badges to our README. Since that branch is not going to be merged, those changes should be made on their own.

https://github.com/quintilesims/layer0/compare/documentation-guides
- Changes in README.md
- Changes in RELEASE.md

tlake:
Closes #178

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,10 @@
 # ![logo](https://quintilesims.github.io/layer0/static/logo_rs.png)
 
+[![Build Status](https://travis-ci.org/quintilesims/layer0.svg?branch=master)](https://travis-ci.org/quintilesims/layer0)  [![Version](https://img.shields.io/badge/version-0.9.0-brightgreen.svg)](https://github.com/quintilesims/layer0/releases/tag/v0.9.0)  [![Documentation](https://img.shields.io/badge/docs-latest-blue.svg)](https://quintilesims.github.io/layer0/)
+
+---
+
+
 Layer0 is a framework that helps you deploy your docker container to the cloud with minimal fuss. Using a simple command line interface (CLI), you can manage the entire life cycle of your application without having to focus on infrastructure.
 
 To get started with Layer0, please see [the Layer0 User's Guide](https://quintilesims.github.io/layer0/).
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,5 +1,10 @@
 # ![logo](https://quintilesims.github.io/layer0/static/logo_rs.png)
 
+[![Build Status](https://travis-ci.org/quintilesims/layer0.svg?branch=master)](https://travis-ci.org/quintilesims/layer0)  [![Version](https://img.shields.io/badge/version-0.9.0-brightgreen.svg)](https://github.com/quintilesims/layer0/releases/tag/v0.9.0)  [![Documentation](https://img.shields.io/badge/docs-latest-blue.svg)](https://quintilesims.github.io/layer0/)
```
diemonster: Is there any way we could avoid having to hardcore the version here?

tlake: Huh, I don't know - it didn't occur to me to look into it. Do we have a pattern for templating `.md`s anywhere else in Layer0? I know we do some stuff with `L0_VERSION?=$(shell git describe --tags)` in our Makefiles.

diemonster: as per our discussion: let's add a new tag with every release called `latest` which we can use here

tlake: Will do - I'll pull in #183

tlake: Further reflection has prompted me to postpone the creation of a `latest` tag. Pushing two tags for the same release would likely wreak havoc wherever our scripts call `git describe --tags`. There's a Stack Overflow question [here](http://stackoverflow.com/questions/8089002/git-describe-with-two-tags-on-the-same-commit) that seems to imply that `git describe --tags` will take the newest tag, which means that in theory we could be sure to create the version-specific tag **after** the `latest` tag, but that seems fragile and janky.

So for now, we'll keep our cool badges and require a little bit of manual attention during the release process (it's documented in RELEASE.md) until we figure something better out.

- Removing #183 from this PR
- Referencing #117 for the future

diemonster: 👍

PR #190, opened 2017-04-03
159 bulk task fix

### LINKED ISSUE #159: l0 stress test failure for task create in bulk
failing our perf tests now

zpatrick:
After many attempts using ecs's builtin `copy` field, I've determined it's unreliable at best, and most likely broken. I switched over to managing the number of copies by creating them one at a time in the `CreateTask` job. This implementation has worked much more reliably and passes the system test. 

As an added benefit, users are no longer restricted to a maximum of 10 copies in `l0 task create` (which was a limit from ECS). They can now create as many copies as they want. 

closes #159

### DIFF
--- a/api/backend/ecs/task_manager.go
+++ b/api/backend/ecs/task_manager.go
@@ -2,7 +2,6 @@ package ecsbackend
 
 import (
 	"fmt"
-	log "github.com/Sirupsen/logrus"
 	"github.com/quintilesims/layer0/api/backend"
 	"github.com/quintilesims/layer0/api/backend/ecs/id"
 	"github.com/quintilesims/layer0/common/aws/cloudwatchlogs"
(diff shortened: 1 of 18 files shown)


diemonster:
This seems like a much better implementation for now. It's simple, it works, and debugging would be way easier. My only worry is API call rates with very large task amounts --- not a problem for the time being, but I'm leaving this here in case we hit the issue down the road.

api/backend/ecs/task_manager.go
```diff
@@ -153,47 +151,16 @@ func (this *ECSTaskManager) CreateTask(
...
-		// don't return error here since we need to return the partial failure
-		model, err := modelFromTasks(tasks)
-		if err != nil {
-			log.Errorf("Failed to get models from tasks in %s: %v", taskID, err)
-		}
-
-		return model, partialFailure
-	}
-
+	tasks, failed, err := this.ECS.RunTask(ecsEnvironmentID.String(), ecsDeployID.TaskDefinition(), 1, stringp(ecsTaskID.String()), ecsOverrides)
```
diemonster: possible culprit? https://github.com/aws/amazon-ecs-agent/issues/166

Issue #194, opened 2017-04-04
private load balancers need to limit traffic from environments


zpatrick:
Right now, the private load balancer security groups allow traffic from anywhere (since we rely on the internal ELB to block out public traffic). 
However, this allows for traffic from anywhere within the VPC to hit the load balancer - e.g. traffic from other environments. 

The private load balancers should _only_ use their environment's security group. 
This will force it to only accept traffic originating from it's environment.

PR #199, opened 2017-04-06
remove sg from private lbs

zpatrick:
closes https://github.com/quintilesims/layer0/issues/194

### DIFF
--- a/api/backend/ecs/load_balancer_manager.go
+++ b/api/backend/ecs/load_balancer_manager.go
@@ -70,7 +70,6 @@ func (e *ECSLoadBalancerManager) GetLoadBalancer(loadBalancerID string) (*models
 		return nil, err
 	}
 
-	// todo: does describe loadbalancer return nil or erro?
 	return e.populateModel(loadBalancer), nil
 }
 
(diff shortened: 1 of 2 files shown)


api/backend/ecs/load_balancer_manager.go
```diff
@@ -304,17 +303,23 @@ func (e *ECSLoadBalancerManager) createLoadBalancer(
 		return err
 	}
 
-	securityGroup, err := e.upsertSecurityGroup(ecsLoadBalancerID, ports)
-	if err != nil {
-		return err
+	// only public load balancers get an additional security group
```
diemonster: are there any side effects here for private load balancers when envs are linked?

zpatrick: Not that I've seen so far. I ran the system tests to verify:
- Public->Private communication without environment linking fails
- Public->Private communication with linking succeeds 

I also spun up a consul cluster behind a private lb and it was able to communicate to itself as expected.

diemonster: > I also spun up a consul cluster behind a private lb and it was able to communicate to itself as expected.

the main case I'm worried about here is a consul cluster in it's own environment, with a separate environment's service(s) trying to hit the consul cluster's private load balancer. If the private load balancer doesn't get the 'Access all' SG, I'm unclear on how other envs will be able to hit it.

zpatrick: that's what was verified with: 
```
* Public->Private communication without environment linking fails
* Public->Private communication with linking succeeds
```

zpatrick: and it does get the 'Access all' SG from the environment

diemonster: Awesome, ok. 👍

Issue #197, opened 2017-04-06
Set "binpack" placement strategy as default


zpatrick:
By default, we should render deploys with the "binpack" strategy if one isn't already set (similar to how we render logging). Docs here: http://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-placement-strategies.html

diemonster:
What is the current default strategy if the param isn't set? I'm not seeing any indicate of what that would be from the ECS docs.

zpatrick:
described here: http://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-placement.html

PR #201, opened 2017-04-06
add binpack strategy to start task/service

zpatrick:
Was able to run system tests to verify changes didn't break anything. 
Added some retry logic to env link test since it can sometimes fail if the private service is still coming up. 

closes https://github.com/quintilesims/layer0/issues/197

### DIFF
--- a/common/aws/ecs/ecs.go
+++ b/common/aws/ecs/ecs.go
@@ -392,6 +392,12 @@ func (this *ECS) RunTask(cluster, taskDefinition string, count int64, startedBy 
 		Count:          aws.Int64(count),
 		StartedBy:      startedBy,
 		Overrides:      taskOverride,
+		PlacementStrategy: []*ecs.PlacementStrategy{
+			{
+				Type:  aws.String(ecs.PlacementStrategyTypeBinpack),
+				Field: aws.String("memory"),
+			},
+		},
 	}
 
 	connection, err := this.Connect()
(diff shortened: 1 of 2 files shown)


common/aws/ecs/ecs.go
```diff
@@ -392,6 +392,12 @@ func (this *ECS) RunTask(cluster, taskDefinition string, count int64, startedBy
 		Count:          aws.Int64(count),
 		StartedBy:      startedBy,
 		Overrides:      taskOverride,
+		PlacementStrategy: []*ecs.PlacementStrategy{
```
diemonster: can this still be overwritten in a task definition to use a different placement strategy?

zpatrick: no, I thought it was something you put in a task definition, but it turns out is only a param on these aws calls.

diemonster: Huh... that's actually pretty weird. I think this is a somewhat unique "problem" since there are very few params parameters that an end user would potentially care about that _aren't_ represented in a task definition. 

To be clear, I feel like the documentation for this feature is incomplete; I'm still not totally sure what "default" is, even if we can make some educated guesses. But this is where we could make Layer0 a marked improvement over ECS in that we can expose these settings within our own configuration.

Anyway, here's ideas along those lines:
 
- We add some API params that allow clients to set the placementStrategy, either through CLI flags or terraform config
- We write some heuristics that make the placementStategy decision for the user 

Anyway, I'm OK with accepting this PR as is, but if I do, I think we should make another issue that allows configuring the placementStrategy.

zpatrick: We will definitely need to separate that out given it's a huge feature request. The scheduler is only setup for binpack right now, and would scale very differently if someone wanted to use the `spread` strategy effectively.

diemonster: 👍

PR #209, opened 2017-04-18
Don't return nil on env link errors

zpatrick:


zpatrick:
Don't merge until the CLI stuff has merged

PR #224, opened 2017-04-24
Add note to clear up confusion about load balancer ports and health checks

### LINKED ISSUE #142: Loadbalancer healthcheck params are confusing
Our current CLI impl of loadbalancer healthchecks are proving confusing for users -- @Cryowatt, @andycmaj, and @sesh-kebab have all missed our docs and/or have gotten bit by the 'new' behavior, which suggests it needs some refinement.

A suggested approach:

- We revert to the old `loadbalancer` healthcheck behavior of picking a supplied port from `--port` and use that for the default healthcheck. We then warn the user that a default has been picked automatically.

tlake:
Closes #142

Additionally, current ongoing work on the guides overhaul will be sure to keep this in mind.

### DIFF
--- a/docs-src/docs/reference/cli.md
+++ b/docs-src/docs/reference/cli.md
@@ -659,8 +659,10 @@ Use the **create** subcommand to create a new load balancer.
 </div>
 
 !!! Note "Ports and Health Checks"
-    When both the `--port` and the `--healthcheck-target` options are omitted, Layer0 configures the load balancer with some default values: `80:80/tcp` for ports and `TCP:80` for healthcheck target.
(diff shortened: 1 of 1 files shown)


docs-src/docs/reference/cli.md
```diff
@@ -658,6 +658,11 @@ Use the **create** subcommand to create a new load balancer.
   </div>
 </div>
 
+!!! Note "Ports and Health Checks"
+    When both the `--port` and the `--healthcheck-target` options are omitted, Layer0 configures the load balancer with some default values: `80:80/tcp` for ports and `TCP:80` for healthcheck target.
+    These default values together create a load balancer configured with a simple but functioning health check, opening up a set of ports that allows traffic to the target of the healthcheck.
```
diemonster: Consider adding a bit at the end of this line that spells out what each element of `80:80/tcp` means.

PR #226, opened 2017-04-26
fix public doc rendering

diemonster:
Docs were not updated with rc2 release; added output from `mkdocs build`


docs/guides/consul/index.html
```diff
@@ -530,8 +530,9 @@ <h2 id="part-3-create-the-service">Part 3: Create the service<a class="headerlin
     <li class="command"><strong>l0 service get consulsvc</strong></li>
   </ul><br />
   When the service has finished provisioning, you will see the following output:</p>
-<pre class="code"><code>SERVICE ID  SERVICE NAME  ENVIRONMENT  LOADBALANCER  DEPLOYS      SCALE
-1consul     consulsvc     demo         consullb      consul:1     1/1</code></pre>
+<div class="code"><pre><span></span>SERVICE ID  SERVICE NAME  ENVIRONMENT  LOADBALANCER  DEPLOYS      SCALE
```
sesh-kebab: weird that empty `<span>`s are being generated.

diemonster: I updated to a newer version of mkdocs and the Material plugin that we use, so that might account for such things... i dunno :-D

Issue #229, opened 2017-05-03
Dynamo read quota hit for tags table


andycmaj:
Seen in L0 api logs during heavy load.

<img width="1369" alt="screencapture at wed may 3 13 16 13 pdt 2017" src="https://cloud.githubusercontent.com/assets/97470/25679605/bcb0bdd2-3002-11e7-946b-0a58186242cf.png">

Note that these throttled requests are SCANs, so there's a good chance these calls could be optimized.

api logs at the time:
<img width="1421" alt="screencapture at wed may 3 15 58 12 pdt 2017" src="https://cloud.githubusercontent.com/assets/97470/25684995/5c4ca58e-3019-11e7-8646-cf123a9c239f.png">

This may or may not have been related to being throttled by ECS at the time, but still could be problematic to expect bursty READ behavior for tags.

andycmaj:
also note that these scans didn't return any more items during this bursty period, they just happened more often
<img width="1036" alt="screencapture at wed may 3 16 01 08 pdt 2017" src="https://cloud.githubusercontent.com/assets/97470/25685078/d0837e50-3019-11e7-8cf0-407b8faa1aa7.png">

andycmaj:
this looks to be happening under heavy load again in our current isolated aws account only running v0.10.0

```
2017-05-04 00:53:34 [Main] INFO : request POST /task (202) 1m51.105456354s

17:53:34
2017-05-04 00:53:34 [Environment Scaler] DEBUG: Scaling environment 'testThe497ef' in 10s

17:53:39
2017-05-04 00:53:39 [Environment Scaler] ERROR: There was an error scaling environment testThe497ef: ProvisionedThroughputExceededException: The level of configured provisioned throughput for the table was exceeded. Consider increasing your provisioning level with the UpdateTable API.

17:53:39
status code: 400, request id: 035MFIJOJF03331JRPQO54PR0VVV4KQNSO5AEMVJF66Q9ASUAAJG

17:53:44
2017-05-04 00:53:44 [Environment Scaler] DEBUG: Scaling environment 'testThe497ef' now

17:53:44
2017-05-04 00:53:44 [ECS Resource Manager] DEBUG: Environment 'l0-test010r2-testThe497ef' generated provider: &resource.ResourceProvider{ID:"i-04c2ea1b0d937c89c", inUse:true, usedPorts:[]int{22, 2376, 2375, 80, 51678, 51679}, availableMemory:3.681550336e+09}

17:53:44
[...]

PR #230, opened 2017-05-04
Optimize dynamodb

zpatrick:
This sets consistent reads **only** after write operations. This should reduce the throughput by a significant amount. 

TODO:
* We will need to try and repro the "job stuck in the pending state" issue shown in https://github.com/quintilesims/layer0/issues/229. To get into this state, the job must have been successfully created by the api (which means the job entry was inserted into dynamodb), but when we tried to query for the job, the entry didn't exist.


andycmaj:
looks like `Job` struct is probably a <4k read but here's some info just wanted to make sure you know about:

> One read capacity unit represents one strongly consistent read per second, or two eventually consistent reads per second, for items up to 4 KB in size. If you need to read an item that is larger than 4 KB, DynamoDB will need to consume additional read capacity units. The total number of read capacity units required depends on the item size, and whether you want an eventually consistent or strongly consistent read.

so switching to eventual consistency for polling job status is probably only going to reduce your read usage by 50%, as opposed to order of magnitude. 
According to 
https://cloud.githubusercontent.com/assets/97470/25687488/76fb26a0-302b-11e7-805a-28b5b7aa425a.png, our spike went over the quota by almost 5X.

andycmaj:
(it could be that you've optimized as much as possible and that we just need to up our dynamo read quota for our prod L0... that's a fine answer too)

andycmaj:
here's a table-scan-items-returned chart during a run a few minutes ago during which we were seeing throttled dynamo reads on Tag go crazy:
screencapture at thu may 4 00 23 58 pdt 2017

here's the resulting throttled request count and quota usage charts for this window:
screencapture at thu may 4 00 25 10 pdt 2017

here are the 12-hour graphs for scan item returned + read rates + throttled request count:
screencapture at thu may 4 00 28 01 pdt 2017
screencapture at thu may 4 00 27 51 pdt 2017
screencapture at thu may 4 00 27 16 pdt 2017

seems like we might have a problem returning too many results from tags table queries.
we've had problems like this in the past, and you might need to re-think indexing/querying scheme for the Tags table

common/db/tag_store/dynamo.go
```diff
@@ -39,24 +46,33 @@ func (d *DynamoTagStore) Clear() error {
...
+	defer d.setConsistentRead(true)
 	return d.table.Delete("TagID", tagID).Run()
 }
 
 func (d *DynamoTagStore) SelectAll() (models.Tags, error) {
+	defer d.setConsistentRead(false)
+
 	tags := models.Tags{}
-	if err := d.table.Scan().Consistent(true).All(&tags); err != nil {
+	if err := d.table.Scan().Consistent(d.consistentRead).All(&tags); err != nil {
```
andycmaj: yeah even with eventually consistent read here, that only halves throughput usage.
our tags table size is getting up to around 10k items with just 12 hours of light usage.

isn't there some way you can query for tags by key type or entity type + environment (i don't see environment id in there right now, but it certainly could be a useful partition or range key.

common/db/tag_store/dynamo.go
```diff
@@ -39,24 +46,33 @@ func (d *DynamoTagStore) Clear() error {
...
 		return nil, err
 	}
 
 	return tags, nil
 }
 
 func (d *DynamoTagStore) SelectByQuery(entityType, entityID string) (models.Tags, error) {
+	defer d.setConsistentRead(false)
+
 	scan := d.table.Scan()
```
andycmaj: yeah if yuo could narrow down items in this Scan by using Environment id as a partition key and querying on environment instead of scanning here, i think it would make a huge difference. 

(granted this is just based on my general understanding of the dynamo concepts of scan and query. i'm not sure how the GO api behaves here when you ask for Scan().All())

andycmaj: but i think this graph pretty much indicates that this isn't going to be scalable at all:
screencapture at thu may 4 01 19 01 pdt 2017

diemonster: We should use something like [ElastiCache](https://aws.amazon.com/elasticache/) to cache read requests to Dynamo. I agree that using Dynamo in the way we are currently is not scalable.

common/db/job_store/dynamo.go
```diff
@@ -73,9 +69,34 @@ func (d *DynamoJobStore) UpdateJobStatus(jobID string, status types.JobStatus) e
...
+
 	if err := d.table.Update("JobID", jobID).Set("Meta", meta).Run(); err != nil {
 		return err
 	}
 
 	return nil
 }
+
+func (d *DynamoJobStore) SelectAll() ([]*models.Job, error) {
+	defer d.setConsistentRead(false)
```
danieljurek: In general I'd do everything possible to avoid fetching all items from a DynamoDB table unless I knew that DynamoDB table had fewer rows than the provisioned capacity.

common/db/job_store/dynamo.go
```diff
@@ -73,9 +69,34 @@ func (d *DynamoJobStore) UpdateJobStatus(jobID string, status types.JobStatus) e
...
+
+	return jobs, nil
+}
+
+func (d *DynamoJobStore) SelectByID(jobID string) (*models.Job, error) {
+	defer d.setConsistentRead(false)
+
+	var job *models.Job
+	if err := d.table.Get("JobID", jobID).Consistent(d.consistentRead).One(&job); err != nil {
+		return nil, err
```
danieljurek: Since `JobID` is a unique identifier on this table the result should either be one item or no items. DynamoDB is very good at fetching a single item on an indexed field. 

To simplify your application design I would always use a consistent read when fetching a single item by its unique identifier.

common/db/tag_store/dynamo.go
```diff
@@ -71,7 +87,7 @@ func (d *DynamoTagStore) SelectByQuery(entityType, entityID string) (models.Tags
 	}
```
danieljurek: The switch statement reads like there are many different ways to query this table. You can create indexes to help simplify some of these, however queries like those on `EntityType` itself looks like they have the potential to return too many results and overrun your table capacity.

api/logic/job_janitor.go
```diff
@@ -9,7 +9,7 @@ import (
 
 const (
 	JOB_LIFETIME           = time.Hour * 24
-	JANITOR_SLEEP_DURATION = time.Hour * 1
+	JANITOR_SLEEP_DURATION = time.Minute * 10
```
zpatrick: note we bump up the job janitor to try and cleanup bursts of activity. every 10 minutes may be high, we could probably set it to 30m

zpatrick:
I've tested the updated code against a tag table that was pre-populated with:
* 50 environment tags
* 250 service tags
* 250 load_balancer tags
* 5k task tags
* 50k deploy tags

The test ran for 3 hours. It would sleep between 1 and 10 seconds between each iteration, and on each iteration it would make query for each entity type. No read thresholds were hit. The queries averaged the following:
* Environment < 50ms
* Service  50-100ms
* Load Balancer 50-100ms
* Task: 200ms-1.1s
* Deploy: 8-10s

The tag table had a read threshold of `50`, write of `5`. The deploy queries are slow, but this type of query only happens on a `l0 deploy list` request (which should happen very infrequently, if ever). Even so, if this query needs to be higher performance, we can double the read throughput from `50` to `100`, and the price per month only goes from around $7 to $12.  

I still need to test against a lot of writes - probably something that simulates a burst workload that chunnels would see (maybe 1k tasks in 5 minutes?). Would like to hear from @andycmaj what an average workload would look like for them to ballpark these tests.

andycmaj:
@zpatrick I think 1k tasks in 5 minutes would get us there with current workload. 
With a single account, we were attempting about 10 tasks per-minute, but that throughput should be multiplied by number of concurrent users (accounts), so let's say multiply that by 10 or 20 since we'll be adding more consumers, and un-bottlenecking will allow us to create tasks faster.

200 per minute wouldn't be an unreasonable throughput for when we're at full scale with some of the bigger accounts.

PR #241, opened 2017-05-30
Task logs refactor

sesh-kebab:
Attempting to address several issues:

- empty logs generates error message
- retrieving logs seems slow

### DIFF
--- a/api/backend/ecs/utilities.go
+++ b/api/backend/ecs/utilities.go
@@ -1,6 +1,8 @@
 package ecsbackend
 
 import (
+	"strings"
+
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	awsecs "github.com/aws/aws-sdk-go/service/ecs"
 	"github.com/quintilesims/layer0/api/backend/ecs/id"
@@ -8,7 +10,6 @@ import (
 	"github.com/quintilesims/layer0/common/aws/ecs"
 	"github.com/quintilesims/layer0/common/config"
 	"github.com/quintilesims/layer0/common/models"
-	"strings"
 )
 
 const MAX_TASK_IDS = 100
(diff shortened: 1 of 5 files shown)


api/backend/ecs/utilities.go
```diff
@@ -1,14 +1,15 @@
 package ecsbackend
 
 import (
+	"strings"
```
diemonster: [nit] there are two auto-formatters for Go code: `gofmt` and `goimports`. The later will alphabetize and separate stdlib and 3rd party libs as seen here, but we've generally been using `gofmt` prior to this. If we want to make the switch to `goimports`, let's make that call, but until then, use `gofmt`.

sesh-kebab:
@diemonster the reason why the logs are delayed when requesting via l0 api is because we are specifying a `FirstEventTimestamp` & `LastEventTimestamp` when making the api call `GetLogEvents`.

We are getting both of the timestamps from an earlier call to `DescribeLogStreams`. The issue lies her; DescribeLogStreams returns a last event timestamp that is eventually consistent. This is the reason it appears to take more time to retrieve the same events from the L0 api where as they are instantly available from AWS console.

> lastEventTimestamp -> (long)
the time of the most recent log event in the log stream in CloudWatch Logs. This number is expressed as the number of milliseconds since Jan 1, 1970 00:00:00 UTC. lastEventTime updates on an eventual consistency basis. It typically updates in less than an hour from ingestion, but may take longer in some rare situations.

To elaborate further:
```
streams[] = describeLogStreams(logGroupId)
for (stream in stream) {
  getLogEvents(
    logGroupId, 
    stream.name, 
    stream.firstEventTimestamp, 
    stream.lastEventTimestamp)  //eventually consistent field
}
```

I propose removing filtering additionally by FirstEventTimestamp and LastEventTimestam. From what I can tell, it doesn't serve any purpose as we don't allow the user to specify the a time range; the time range is essentially always what is returned from the stream. This requires changing the `Provider` interface in `layer0/common/aws/cloudwatchlogs.go`. Thoughts?

diemonster:
@seshi Interesting, makes perfect sense. I _think_ we previously fixed on supplying the `LastEventTimestamp` due to some edge cases we found when not supplying it, but my memory of that decision is a bit hazy.

For a long running service, not having the ability to filter First and Last timestamps could become unwieldy very quickly. I think the ideal option in that case would be to have optional params for First and Last timestamps. So either we get that fixed as part of this, or we spin off another issue to finish that part up for later... your call.

sesh-kebab:
> So either we get that fixed as part of this, or we spin off another issue to finish that part up for later... your call.

Agree that it is definitely essential for a long running service but I would like to go ahead and spin that off as a separate issue.

PR #267, opened 2017-06-15
Fix documentation oversights

### LINKED ISSUE #251: Add 'introduction to Terraform' in docs
We somewhat glance over the following regarding our usage of Terraform:

- Why Terraform? 
  - Why did we chose Terraform over Cloud Formation?
- What does Terraform do?
- What are the advantages of using Terraform versus the `l0` cli?
- Where do you download / install terraform?

I'm thinking a dedicated page in `/references` would be appropriate, along with a link somewhere in our introduction to said page.

### LINKED ISSUE #254: Fix documentation oversights
There are a few things in our docs that need ironing out/clarifying/etc. since overhauling our guides and L0-setup. This issue will detail them as we find them so we can address them all together.

**/setup/install**

- [x] Include Terraform requirement to be able to use `l0-setup` (call out Terraform v0.9.4+ as explicit pre-req in /setup/install)
    - [x] (also link to _about Terraform_ section as suggested in #251)
- [x] We mention `l0-setup init <prefix>` and `apply`, but are not explicit about `l0-setup apply <prefix>`

**/reference/terraform-plugin**

- ~Fix link to custom `terraform-provider-layer0` in /reference/terraform-plugin page~
    - _not a link; will be more clear when font color differentiates links and code better_
- ~Terraform docs under /reference/terraform-plugin/,  the `provider` link to terraform's providers/index.html doesn't help much.~
    - _leaving item unchanged - there seems to be enough information_
- [x] Terraform docs under /reference/terraform-plugin/, under Prerequisites, the terraform binary found under /bin directory of L0 release is no longer valid.  Perhaps remove the entire prereq section, and go directly to installation.  Make it clear the terraform-provider-layer0 is already downloaded as part of the layer0 release, and to install, copy it under the same directory as the terraform.exe.

**/guides/walkthrough/\***

- ~Make the link to the guides in repo from above the file list table in /walkthrough/deployment-1 to https://github.com/quintilesims/guides/tree/master/walkthrough/, because there is another guestbook inside the Guides, and it can get confusing.~
    - _opting to not make this change for now, as I believe that the current presentation is less confusing_
- [x] In /walkthrough/deployment-1/, mention the benefits of having deploy as an entity.  something along the lines that deploys, once created, can be applied to any service across environments.  maybe mention the versioning of deploys, too, in the listing of deploy command options near bottom of deploy section.

**other**

- [x] update color coding in walk through to use different color for links (blue?).  Currently, RED is used for both key file names and links, which makes links hard to see.
    - _opted to change code block color and leave links as red_

tlake:
Update documentation with information that we've missed or that has caused confusion for users.

No real testing - just read over the documentation changes and see if everything makes sense and is clear. The issue links below should be beneficial in guiding a review.

**Checklist**
- ~Unit tests~
- ~Smoke tests~
- ~System tests~
- [x] Documentation

closes #254
closes #251

### DIFF
--- a/docs-src/docs/guides/walkthrough/deployment-1.md
+++ b/docs-src/docs/guides/walkthrough/deployment-1.md
@@ -113,6 +113,7 @@ Try running the following commands to get an idea of the information available t
 ### Part 3: Deploy the ECS Task Definition
 
 The `deploy` command is used to specify the ECS task definition that outlines a web application.
+A deploy, once created, can be applied to multiple services - even across different environments!
 
 Here, we'll create a new deploy called **guestbook-dpl** that refers to the **Guestbook.Dockerrun.aws.json** file found in the guides reposiory.
 At the command prompt, execute the following:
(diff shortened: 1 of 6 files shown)


tlake:
The only unaddressed item remaining in #254 concerns a link to Terraform's documentation about providers. The item suggests that the link is unhelpful, yet in my attempt to add some clarifying text, I found that I was essentially paraphrasing the Terraform docs, which leads me to believe either that the link _is_ helpful, or that I'm not seeing the problem.

Is there anything specific that I should write up that neither we nor Terraform cover regarding providers? Otherwise, I'm tempted to close that item for the time being.

Issue #246, opened 2017-05-31
Allow logs to be retrieved using a date range


sesh-kebab:
`Task Logs` or `Service Logs` currently only allows `--tail` parameter to be specified which can be used to limit the results returned. For long running services, it would be useful to also allow the user to specify a date range when requesting logs `--startTimstamp` & `--endTimestamp`

Related to #140

PR #268, opened 2017-06-15
Removed storedBytes check

### LINKED ISSUE #259: service/task logs are returning empty responses
Version: `v0.10.0`

Running `l0 service logs api` returns an empty response from the API when there should be logs present. 

Debug output:
```
$ l0 -d service logs api
DEBU[0000] Sent: GET https://l0-test-api-xxx.us-west-2.elb.amazonaws.com/tag?fuzz=api&type=service

DEBU[0000] Received: 200 OK [
  {
   "entity_id": "api",
   "entity_type": "service",
   "tags": [
    {
     "entity_id": "api",
     "entity_type": "service",
     "key": "environment_id",
     "value": "api"
    },
    {
     "entity_id": "api",
     "entity_type": "service",
     "key": "load_balancer_id",
     "value": "api"
    },
    {
     "entity_id": "api",
     "entity_type": "service",
     "key": "name",
     "value": "api"
    }
   ]
  }
 ]

DEBU[0000] Sent: GET  https://l0-test-api-xxx.us-west-2.elb.amazonaws.com/service/api/logs

DEBU[0000] Received: 200 OK []

```

sesh-kebab:
Fixes a regression which caused `l0 task/service logs` command to return empty results. This regression was introduced by #241.

Create a l0 task or service that generates logs. For example, you can follow the [one-off-task guide](http://layer0.ims.io/guides/one_off_task/). Once the task is created, use `l0 task logs <taskid>` to confirm that the logs written are correctly being output as they are being written to cloudwatch logs.

The logs visible from the Cloudwatch logs sections in AWS console, should match the output generated by `l0 task logs <taskid>`.

**Checklist**
- [x] Unit tests

closes #259

### DIFF
--- a/api/backend/ecs/utilities.go
+++ b/api/backend/ecs/utilities.go
@@ -177,14 +177,11 @@ var getTaskARNs = func(ecs ecs.Provider, ecsEnvironmentID id.ECSEnvironmentID, s
 var GetLogs = func(cloudWatchLogs cloudwatchlogs.Provider, taskARNs []*string, tail int) ([]*models.LogFile, error) {
 	taskIDCatalog := generateTaskIDCatalog(taskARNs)
 
-	fmt.Println("Getting streams")
-
 	orderBy := "LogStreamName"
 	logStreams, err := cloudWatchLogs.DescribeLogStreams(config.AWSLogGroupID(), orderBy)
 	if err != nil {
 		return nil, err
 	}
-	fmt.Println("done")
 
 	logFiles := []*models.LogFile{}
 	for _, logStream := range logStreams {
@@ -204,7 +201,6 @@ var GetLogs = func(cloudWatchLogs cloudwatchlogs.Provider, taskARNs []*string, t
 			Lines: []string{},
 		}
 
-		fmt.Println("Getting events")
 		// since the time range is exclusive, expand the range to get first/last events
 		logEvents, err := cloudWatchLogs.GetLogEvents(
 			config.AWSLogGroupID(),
@@ -219,7 +215,6 @@ var GetLogs = func(cloudWatchLogs cloudwatchlogs.Provider, taskARNs []*string, t
 			logFile.Lines = append(logFile.Lines, *logEvent.Message)
 		}
 
-		fmt.Println("done")
 		logFiles = append(logFiles, logFile)
 	}
 
--- a/api/backend/ecs/utilities.go
+++ b/api/backend/ecs/utilities.go
@@ -2,6 +2,8 @@ package ecsbackend
 
 import (
 	"fmt"
+	"strings"
+
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	awsecs "github.com/aws/aws-sdk-go/service/ecs"
 	"github.com/quintilesims/layer0/api/backend/ecs/id"
@@ -9,7 +11,6 @@ import (
 	"github.com/quintilesims/layer0/common/aws/ecs"
 	"github.com/quintilesims/layer0/common/config"
 	"github.com/quintilesims/layer0/common/models"
-	"strings"
 )
 
 const MAX_TASK_IDS = 100
@@ -187,10 +188,6 @@ var GetLogs = func(cloudWatchLogs cloudwatchlogs.Provider, taskARNs []*string, t
 
 	logFiles := []*models.LogFile{}
 	for _, logStream := range logStreams {
-		if *logStream.StoredBytes == 0 {
-			continue
-		}
-
 		// filter by streams that have <prefix>/<container name>/<stream task id>
 		streamNameSplit := strings.Split(*logStream.LogStreamName, "/")
 		if len(streamNameSplit) != 3 {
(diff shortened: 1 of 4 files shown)


api/backend/ecs/utilities.go
```diff
@@ -187,10 +188,6 @@ var GetLogs = func(cloudWatchLogs cloudwatchlogs.Provider, taskARNs []*string, t
 
 	logFiles := []*models.LogFile{}
 	for _, logStream := range logStreams {
-		if *logStream.StoredBytes == 0 {
```
zpatrick: how are we sure this wont re-introduce the json marshal error when the stream is empty?

sesh-kebab: That's the part i'm working on now. The puzzling thing is I'm unable to reproduce the json marshal error for empty streams using the previous steps. I definitely intend to address it before removing the wip status.

api/backend/ecs/utilities.go
```diff
@@ -187,10 +188,6 @@ var GetLogs = func(cloudWatchLogs cloudwatchlogs.Provider, taskARNs []*string, t
```
zpatrick: also, you have debug statements left over in the code you forgot to remove last time

sesh-kebab: 😱

api/backend/ecs/utilities.go
sesh-kebab: I confirmed the issue with v0.9.0. The error stems from FirstEventTimestamp property being accessed without checking for nil. 

```
// since the time range is exclusive, expand the range to get first/last events
logEvents, err := cloudWatchLogs.GetLogEvents(
	config.AWSLogGroupID(),
	*logStream.LogStreamName,
	*logStream.FirstEventTimestamp-1,
	*logStream.LastEventTimestamp+1,
	int64(tail))
```

As this code has been removed in v0.10.0 empty logs will no longer generate the json marshal error in theory (and observed in practice) . Based on this, we can be sure that it won't re-introduce the original error.

api/backend/ecs/utilities.go
sesh-kebab: not sure which lines you are referring to?

zpatrick: * https://github.com/quintilesims/layer0/blob/develop/api/backend/ecs/utilities.go#L179
* https://github.com/quintilesims/layer0/blob/develop/api/backend/ecs/utilities.go#L186
* https://github.com/quintilesims/layer0/blob/develop/api/backend/ecs/utilities.go#L210
* https://github.com/quintilesims/layer0/blob/develop/api/backend/ecs/utilities.go#L225

zpatrick:
I'm trying to run this, and the `log` command returns `EOF` after about 45 seconds or so
```
[ec2-user@ip-10-100-10-227 layer0]$ l0 environment delete dev
This operation is running as a job. Run `l0 job get 3fbeb4bfb96c` to see progress
[ec2-user@ip-10-100-10-227 layer0]$ l0 job logs 3fbeb4bfb96c
EOF
```

sesh-kebab:
[ec2-user@ip-10-100-10-227 layer0]$ l0 job logs 3fbeb4bfb96c
EOF

Is expected give the previous code path skipped creating an new instance of logFile
```
logFile := &models.LogFile{
	Name:  streamNameSplit[1],
	Lines: []string{},
}
```

This lead to just an empty array being returned.

zpatrick:
An `EOF` response is an error, not the response of an empty array. This can happen when a request takes too long and the loadbalancer terminates the request. It can also happen when the API crashes.

zpatrick:
I think this was the result of a timeout from ` cloudWatchLogs.DescribeLogStreams`. Timeout occurred because there were so many streams in my log group. We should investigate if we can optimize that api call better, or we should set log streams to expire after time.

sesh-kebab:
> An EOF response is an error, not the response of an empty array. 

Ok, my mistake here. Although I did see this even with new environments where I only had a few streams. So i'm assuming the api was crashing and we could be surfacing those errors via the CLI a bit better?

> I think this was the result of a timeout from cloudWatchLogs.DescribeLogStreams. 

nice catch! would you please be able to provide your l0 environment details so I can repro this?

@zpatrick ^

zpatrick:
> Ok, my mistake here. Although I did see this even with new environments where I only had a few streams. So i'm assuming the api was crashing and we could be surfacing those errors via the CLI a bit better?

Yes, but preventing whatever is causing the API to crash is a much higher priority.

zpatrick:
> nice catch! would you please be able to provide your l0 environment details so I can repro this?

Actually no, I unfortunately cleared out my log streams.  My log group had 60k streams when I was testing it.

sesh-kebab:
@zpatrick, following on from our meeting - i've made a couple of key changes for the issue of throttling limits when calling `describeLogStreams` aws api call.

1. Limit call to 5 pages
2. Instead of ordering by `LogStreamName`, i'm now ordering by `LastEventTime`

The intention with these changes is to firstly avoid running into the limit but also cater for the 90% case where the user is likely to want to retrieve **recent** logs. Admittedly the evidence for this is somewhat anecdotal. This would search the 250 streams of the streams with the latest data (page size = 50 and page limit = 5).

So in effect, even if a user has thousands of streams, we are only ever looking at the 250 of the latest stream avoiding the throttling issue and providing consistently quick results regardless of the number of streams in a given log group. As an extension to this, for the 10% case where the user is trying to narrow down on a old time range, #246 should have them covered.

PR #270, opened 2017-06-20
limit streams to 1000/call

zpatrick:
* Removes debug statements
* Limits number of streams to 1000. This takes between 2-4 seconds for a `l0 service logs` call. Seems a reasonable stream:time ratio to me - open to suggestions if it should be changed. 
* Fixed broken exit login on pagination function

### DIFF
--- a/common/aws/cloudwatchlogs/cloudwatchlogs.go
+++ b/common/aws/cloudwatchlogs/cloudwatchlogs.go
@@ -6,6 +6,12 @@ import (
 	"github.com/quintilesims/layer0/common/aws/provider"
 )
 
+const (
+	// DescribeLogStreams is throttled after five transactions per second.
+	// With 50 streams/transaction, 1000 gives a reasonable streams:time ratio
+	MAX_DESCRIBE_STREAMS_COUNT = 1000
+)
+
 type Provider interface {
 	CreateLogGroup(logGroupName string) error
 	DeleteLogGroup(logGroupName string) error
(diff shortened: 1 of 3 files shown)


common/aws/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -166,26 +172,29 @@ func (this *CloudWatchLogs) DescribeLogStreams(logGroupName, orderBy string) ([]
...
-	pageNum := 0
+	streams := []*LogStream{}
 	pagef := func(output *cloudwatchlogs.DescribeLogStreamsOutput, lastPage bool) bool {
-		pageNum++
 		for _, stream := range output.LogStreams {
-			result = append(result, &LogStream{stream})
+			streams = append(streams, &LogStream{stream})
 		}
 
-		return !lastPage || pageNum <= 5
```
sesh-kebab: Yikes - fixed this! 😞

common/aws/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -166,26 +172,29 @@ func (this *CloudWatchLogs) DescribeLogStreams(logGroupName, orderBy string) ([]
 	}
 
 	input := &cloudwatchlogs.DescribeLogStreamsInput{
-		LogGroupName: aws.String(logGroupName),
-		OrderBy:      aws.String(orderBy),
+		LogGroupName:        aws.String(logGroupName),
+		OrderBy:             aws.String(orderBy),
+		Descending:          aws.Bool(true),
```
zpatrick: I may need to test again, but w/o this I'm pretty sure you get oldest logs first. For example, I wasn't getting logs for the `api` service because there are like 6k older streams in front of it.

PR #272, opened 2017-06-29
Added data sources: environment, loadbalancer, service & deploy

### LINKED ISSUE #263: Terraform: Create data sources for entities
All entities should have corresponding data sources:
https://www.terraform.io/docs/configuration/data-sources.html

example:
```
data "layer0_environment" "foo" {
    name = "foo"
}

data "layer0_service" "bar" {
    environment_id = "${data.layer0_environment.foo.id}"
    name           = "bar"
}
```

sesh-kebab:
Modifies the Terraform Layer0 provider to add Data Source support for Layer0 created resources. Data Sources added:

- environment
- load balancer
- service
- deploy

Create a terraform file that references existing l0 environment, load balancer, service & deploy. In the output variables, output all the supported variables supported for a given resource. Execute `terraform apply` and check the output.

For example:
```
provider "layer0" {
  token             = "${var.token}"
  skip_ssl_verify = true
}

data "layer0_deploy" "service-a-dpl" {
  name    = "some-cool-service-dpl"
  version = "1"
}

output "deploy" {
  value = "${data.layer0_deploy.service-a-dpl.id}"
}
```

**Checklist**
- [x] Unit tests

closes #263

### DIFF
--- a/plugins/terraform/data_source_layer0_deploy.go
+++ b/plugins/terraform/data_source_layer0_deploy.go
@@ -31,11 +31,11 @@ func datasourceLayer0DeployRead(d *schema.ResourceData, meta interface{}) error 
 
 	deployName := d.Get("name").(string)
 	version := d.Get("version").(string)
-
-	deployID, err := resolveTags(client, deployName, map[string]string{
-		"type":    "deploy",
+	params := map[string]string{
 		"version": version,
-	})
+	}
+
+	deployID, err := resolveTags(client, deployName, "deploy", params)
 	if err != nil {
 		return err
 	}
@@ -48,7 +48,6 @@ func datasourceLayer0DeployRead(d *schema.ResourceData, meta interface{}) error 
 	d.SetId(deploy.DeployID)
 
 	return setResourceData(d.Set, map[string]interface{}{
-		"id":      deploy.DeployID,
 		"name":    deploy.DeployName,
 		"version": deploy.Version,
 	})
--- a/plugins/terraform/data_source_layer0_deploy.go
+++ b/plugins/terraform/data_source_layer0_deploy.go
@@ -1,0 +1,55 @@
+package main
+
+import (
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/quintilesims/layer0/cli/client"
+)
+
+func dataSourceLayer0Deploy() *schema.Resource {
+	return &schema.Resource{
+		Read: datasourceLayer0DeployRead,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"version": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func datasourceLayer0DeployRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(client.Client)
+
+	deployName := d.Get("name").(string)
+	version := d.Get("version").(string)
+
+	deployID, err := resolveTags(client, deployName, map[string]string{
+		"type":    "deploy",
+		"version": version,
+	})
+	if err != nil {
+		return err
+	}
+
+	deploy, err := client.GetDeploy(deployID)
+	if err != nil {
+		return err
+	}
+
+	d.SetId(deploy.DeployID)
+
+	return setResourceData(d.Set, map[string]interface{}{
+		"id":      deploy.DeployID,
+		"name":    deploy.DeployName,
+		"version": deploy.Version,
+	})
+}
--- a/plugins/terraform/data_source_layer0_deploy_test.go
+++ b/plugins/terraform/data_source_layer0_deploy_test.go
@@ -1,0 +1,34 @@
+package main
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func TestDeployDataResourceSelectByQueryParams(t *testing.T) {
+	ctrl, mockClient, provider := setupUnitTest(t)
+	defer ctrl.Finish()
+
+	deployName := "deploy-name"
+	version := "1"
+
+	params := map[string]string{
+		"type":    "deploy",
+		"version": version,
+		"fuzz":    deployName,
+	}
+
+	mockClient.EXPECT().
+		SelectByQuery(params)
+
+	deployResource := provider.DataSourcesMap["layer0_deploy"]
+	d := schema.TestResourceDataRaw(t, deployResource.Schema, map[string]interface{}{})
+	d.Set("name", deployName)
+	d.Set("version", version)
+
+	if err := deployResource.Read(d, mockClient); err.Error() != fmt.Errorf("No entities found").Error() {
+		t.Fatal(err)
+	}
+}
--- a/plugins/terraform/data_source_layer0_deploy_test.go
+++ b/plugins/terraform/data_source_layer0_deploy_test.go
@@ -1,16 +1,17 @@
 package main
 
 import (
-	"fmt"
 	"testing"
 
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/quintilesims/layer0/common/models"
 )
 
 func TestDeployDataResourceSelectByQueryParams(t *testing.T) {
 	ctrl, mockClient, provider := setupUnitTest(t)
 	defer ctrl.Finish()
 
+	deployId := "deploy-id"
 	deployName := "deploy-name"
 	version := "1"
 
@@ -21,14 +22,25 @@ func TestDeployDataResourceSelectByQueryParams(t *testing.T) {
 	}
 
 	mockClient.EXPECT().
-		SelectByQuery(params)
+		SelectByQuery(params).
+		Return([]*models.EntityWithTags{
+			&models.EntityWithTags{
+				EntityID:   deployId,
+				EntityType: "deploy",
+			},
+		}, nil)
+
+	mockClient.EXPECT().
+		GetDeploy(deployId).
+		Return(&models.Deploy{}, nil)
 
 	deployResource := provider.DataSourcesMap["layer0_deploy"]
-	d := schema.TestResourceDataRaw(t, deployResource.Schema, map[string]interface{}{})
-	d.Set("name", deployName)
-	d.Set("version", version)
+	d := schema.TestResourceDataRaw(t, deployResource.Schema, map[string]interface{}{
+		"name":    deployName,
+		"version": version,
+	})
 
-	if err := deployResource.Read(d, mockClient); err.Error() != fmt.Errorf("No entities found").Error() {
+	if err := deployResource.Read(d, mockClient); err != nil {
 		t.Fatal(err)
 	}
 }
--- a/plugins/terraform/data_source_layer0_environment.go
+++ b/plugins/terraform/data_source_layer0_environment.go
@@ -43,9 +43,7 @@ func datasourceLayer0EnvironmentRead(d *schema.ResourceData, meta interface{}) e
 
 	environmentName := d.Get("name").(string)
 
-	environmentID, err := resolveTags(client, environmentName, map[string]string{
-		"type": "environment",
-	})
+	environmentID, err := resolveTags(client, environmentName, "environment", map[string]string{})
 	if err != nil {
 		return err
 	}
@@ -58,7 +56,6 @@ func datasourceLayer0EnvironmentRead(d *schema.ResourceData, meta interface{}) e
 	d.SetId(environment.EnvironmentID)
 
 	return setResourceData(d.Set, map[string]interface{}{
-		"id":        environment.EnvironmentID,
 		"size":      environment.InstanceSize,
 		"min_count": environment.ClusterCount,
 		"os":        environment.OperatingSystem,
--- a/plugins/terraform/data_source_layer0_environment.go
+++ b/plugins/terraform/data_source_layer0_environment.go
@@ -1,0 +1,67 @@
+package main
+
+import (
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/quintilesims/layer0/cli/client"
+)
+
+func dataSourceLayer0Environment() *schema.Resource {
+	return &schema.Resource{
+		Read: datasourceLayer0EnvironmentRead,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"size": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"min_count": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+			"os": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"ami": {
+				Type:     schema.TypeString,
+				Computed: true,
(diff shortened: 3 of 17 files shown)


zpatrick:
You need to add documentation!

plugins/terraform/data_source_layer0_deploy.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+}
+
+func datasourceLayer0DeployRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(client.Client)
+
+	deployName := d.Get("name").(string)
+	version := d.Get("version").(string)
+
+	deployID, err := resolveTags(client, deployName, map[string]string{
```
zpatrick: [nit] this code block does't read well - maybe do something like:
```
params := map[string]string{
    "type": "deploy",
    "version": version,
}

deployID, err := resolveTags(client, deployName, params)
if err != nil {
    return err
}
...
```

(this applies to all entities)

plugins/terraform/data_source_layer0_deploy_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+	mockClient.EXPECT().
+		SelectByQuery(params)
+
+	deployResource := provider.DataSourcesMap["layer0_deploy"]
+	d := schema.TestResourceDataRaw(t, deployResource.Schema, map[string]interface{}{})
+	d.Set("name", deployName)
+	d.Set("version", version)
+
+	if err := deployResource.Read(d, mockClient); err.Error() != fmt.Errorf("No entities found").Error() {
```
zpatrick: This error checking is awkward. 
You should setup the mock to handle the `GET /deploy/:deployID` call, see: https://github.com/quintilesims/layer0/blob/develop/plugins/terraform/resource_deploy_test.go#L32

(this applies to all entities)

plugins/terraform/data_source_layer0_environment.go
```diff
@@ -0,0 +1,67 @@
...
+
+	environment, err := client.GetEnvironment(environmentID)
+	if err != nil {
+		return err
+	}
+
+	d.SetId(environment.EnvironmentID)
+
+	return setResourceData(d.Set, map[string]interface{}{
+		"id":        environment.EnvironmentID,
```
zpatrick: don't set `id` here

plugins/terraform/data_source_layer0_deploy_test.go
```diff
@@ -0,0 +1,34 @@
...
+		"type":    "deploy",
+		"version": version,
+		"fuzz":    deployName,
+	}
+
+	mockClient.EXPECT().
+		SelectByQuery(params)
+
+	deployResource := provider.DataSourcesMap["layer0_deploy"]
+	d := schema.TestResourceDataRaw(t, deployResource.Schema, map[string]interface{}{})
```
zpatrick: Please follow existing pattern for setting raw data: https://github.com/quintilesims/layer0/blob/develop/plugins/terraform/resource_load_balancer_test.go#L31

(this applies to all entities)

plugins/terraform/data_source_layer0_loadbalancer.go
```diff
@@ -0,0 +1,104 @@
...
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"environment_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"environment_name": {
```
zpatrick: `environment_name` should be computed, not optional

plugins/terraform/data_source_layer0_loadbalancer.go
```diff
@@ -0,0 +1,104 @@
...
+			},
+			"service_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"service_name": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"healthcheck_healthy_threshold": {
```
zpatrick: Having healthcheck data seems overkill in my opinion - I'd vote for removing it. 
If we keep it, then we should follow the existing pattern: https://github.com/quintilesims/layer0/blob/develop/plugins/terraform/resource_load_balancer.go#L113

plugins/terraform/data_source_layer0_loadbalancer.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	loadbalancer, err := client.GetLoadBalancer(loadbalancerID)
+	if err != nil {
+		return err
+	}
+
+	d.SetId(loadbalancer.LoadBalancerID)
+	return setResourceData(d.Set, map[string]interface{}{
+		"id":                              loadbalancer.LoadBalancerID,
```
zpatrick: don't set id

plugins/terraform/data_source_layer0_service.go
```diff
@@ -0,0 +1,74 @@
...
+	service, err := client.GetService(serviceID)
+	if err != nil {
+		return err
+	}
+
+	d.SetId(service.ServiceID)
+
+	return setResourceData(d.Set, map[string]interface{}{
+		"name":               service.ServiceName,
+		"id":                 service.ServiceID,
```
zpatrick: id

plugins/terraform/data_source_tag_resolver.go
```diff
@@ -0,0 +1,53 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/quintilesims/layer0/cli/client"
+)
+
+func resolveTags(client client.Client, target string, params map[string]string) (string, error) {
```
zpatrick: Since `type` is a required param for the API, it should probably be a param here instead of relying on the caller to know to use it, e.g.:
```
 resolveTags(client client.Client, target, entityType string, params map[string]string) (string, error) { ... }
```

plugins/terraform/data_source_tag_resolver.go
```diff
@@ -0,0 +1,53 @@
...
+	params["fuzz"] = target
+
+	taggedEntities, err := client.SelectByQuery(params)
+	if err != nil {
+		return "", err
+	}
+
+	if len(taggedEntities) == 0 {
+		return "", fmt.Errorf("No entities found")
```
zpatrick: This should be improved:
```
return "", fmt.Errorf("No entities of type %s found matching %s", entityType, target)
```

plugins/terraform/data_source_tag_resolver.go
```diff
@@ -0,0 +1,53 @@
...
+		entityIDs = append(entityIDs, taggedEntity.EntityID)
+
+		for _, tag := range taggedEntity.Tags {
+			if tag.Key == "name" && tag.Value == target {
+				return taggedEntity.EntityID, nil
+			}
+		}
+	}
+
+	text := fmt.Sprintf("Multiple entities found: \n")
```
zpatrick: `fmt.Errorf("Multiple entities of type %s found matching %s: ", entityType, target)`

plugins/terraform/data_source_tag_resolver.go
```diff
@@ -0,0 +1,53 @@
...
+	for _, id := range entityIDs {
+		text += fmt.Sprintf("%s \n", id)
+	}
+
+	return "", fmt.Errorf(text)
+}
+
+func setResourceData(setter func(string, interface{}) error, values map[string]interface{}) error {
+	for key, value := range values {
+		err := setter(key, value)
```
zpatrick: go-ism: always keep variables in the smallest scope possible:
```
if err := setter(key, val); err != nil {
    return err
}
```

zpatrick:
I think we should add a system test for these changes. It can be a separate work item, but would definitely give me/us much more confidence in this code than the unit tests can.

plugins/terraform/data_source_layer0_deploy.go
```diff
@@ -0,0 +1,55 @@
...
+
+	deploy, err := client.GetDeploy(deployID)
+	if err != nil {
+		return err
+	}
+
+	d.SetId(deploy.DeployID)
+
+	return setResourceData(d.Set, map[string]interface{}{
+		"id":      deploy.DeployID,
```
sesh-kebab: I was following the example set in data_source_api. Assuming that should be removed also? Reading the docs I wasn't sure how it applies to data sources.

https://godoc.org/github.com/hashicorp/terraform/helper/schema#ResourceData.SetId

> SetId sets the ID of the resource. If the value is blank, then the resource is destroyed.

plugins/terraform/data_source_layer0_deploy.go
sesh-kebab: Not sure I agree:
 - I feel it is more concise
 - with reference to the go-ism referenced in a later comment, params doesn't need to exist in the outside scope
 - but if we make type a fourth argument (as mentioned in a later comment), calling statement would look slightly better I suppose.

plugins/terraform/data_source_layer0_deploy_test.go
sesh-kebab: 👍

plugins/terraform/data_source_layer0_deploy_test.go
sesh-kebab: Yea definitely agree on the awkwardness. So what I was trying to do here was test `SelectByQuery` will be called with the right arguments. The mock I would need to setup would need to return `TaggedEntities` which I thought might be overkill. But will look at how to make this better. 👍

plugins/terraform/data_source_layer0_environment.go
sesh-kebab: ✅

plugins/terraform/data_source_layer0_loadbalancer.go
sesh-kebab: ✅

plugins/terraform/data_source_layer0_loadbalancer.go
sesh-kebab: 🤔

plugins/terraform/data_source_layer0_loadbalancer.go
sesh-kebab: ✅

plugins/terraform/data_source_layer0_service.go
sesh-kebab: ✅

plugins/terraform/data_source_layer0_environment.go
sesh-kebab: ✅

plugins/terraform/data_source_layer0_loadbalancer.go
sesh-kebab: ✅

plugins/terraform/data_source_layer0_service.go
sesh-kebab: ✅

plugins/terraform/data_source_tag_resolver.go
sesh-kebab: ✅

plugins/terraform/data_source_tag_resolver.go
sesh-kebab: ✅

plugins/terraform/data_source_tag_resolver.go
sesh-kebab: ✅

plugins/terraform/data_source_tag_resolver.go
sesh-kebab: 👍

plugins/terraform/data_source_layer0_deploy.go
zpatrick: all good points - but if you recall,  we both forgot to handle the error returned by this function. I can't speak for you, but I know I missed it because I'm so used to seeing error handling on the next line:
```
blah, err := ...
if err != nil { ... }
```

and having written in go for a while, I can say almost every time an error checking is missing,  it's because of a code block that doesn't read well.

zpatrick: All this being said, I admit it's still very much a nitpick!

plugins/terraform/data_source_layer0_deploy.go
zpatrick: Yeah, the blank value thing doesn't make sense. I think you got it correct on line 48, and should remove line 51. At the very least line 51 is redundant, but it could potentially cause issues treating it as a non-special field.

plugins/terraform/data_source_layer0_deploy_test.go
zpatrick: The resource tests are a good point of reference. I think testing a service could look something like:
```
	ewts := []models.EntityWithTags{
		{EntityID: "id", Tags: []models.Tag{Key: "name", "value": "some_service"}},
	}

	mockClient.EXPECT().
		SelectByQuery(params).
		Return(ewts, nil)

       service := models.Service{ServiceID: "some_id", ServiceName: "some_service", ...}

	mockClient.EXPECT().
		GetService("id").
		Return(service, nil)

        // set d.RawData...

	if err := deployResource.Read(d, mockClient); err != nil {
		t.Fatal(err)
	}

      // maybe do some assertions that fields were set correctly
    assert.Equal(t, d.["service_name"], "some_service")
```

plugins/terraform/data_source_layer0_loadbalancer.go
zpatrick: If you keep it, I _think_ you could just do:

```
return setResourceData(d.Set, map[string]interface{}{
    ...
   "health_check": flattenHealthCheck(loadBalancer.HealthCheck),
```

sesh-kebab:
Added a new issue for the system tests task: https://github.com/quintilesims/layer0/issues/280

zpatrick:
@sesh-kebab : this needs documentation

plugins/terraform/data_source_layer0_environment.go
```diff
@@ -0,0 +1,64 @@
...
+		},
+	}
+}
+
+func datasourceLayer0EnvironmentRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(client.Client)
+
+	environmentName := d.Get("name").(string)
+
+	environmentID, err := resolveTags(client, environmentName, "environment", map[string]string{})
```
zpatrick: [nit] zero value for maps is `nil`

plugins/terraform/data_source_layer0_environment_test.go
```diff
@@ -0,0 +1,43 @@
...
+	params := map[string]string{
+		"type": "environment",
+		"fuzz": environmentName,
+	}
+
+	mockClient.EXPECT().
+		SelectByQuery(params).
+		Return([]*models.EntityWithTags{
+			&models.EntityWithTags{
+				EntityID:   environmentId,
```
zpatrick: `environmentID`

plugins/terraform/data_source_layer0_environment.go
sesh-kebab: This is intentional. `resolveTags` expects a map; with `fuzz` and `type` keys being added. Are you saying I should check for a nil map in resolveTags and create one if one is not passed in?

plugins/terraform/data_source_layer0_environment_test.go
sesh-kebab: ✅

sesh-kebab:
@zpatrick regarding the documentation, I would like to submit a separate PR for the documentation  as with the system tests rather than iterating on this PR.

plugins/terraform/data_source_layer0_environment.go
zpatrick: ah I see, I didn't catch that earlier

PR #283, opened 2017-07-12
Issue 280 datasource system tests

### LINKED ISSUE #280: Layer0 Terraform Provider - System Tests for Datasources
Create System Tests to verify Data sources supported in Layer0 tf provider.

Data Sources: 
- Load Balancer
- Environment
- Service
- Deploy

sesh-kebab:
Adds systems tests to verify the output of Layer0 provider's Datasource entities added in #263 . Also fixes functionality that exposes `layer0_deploy.version` & `enviornment.cluster_count` resources.

From the `./tests/system` directory run `go test -v -debug -run TestDataSources`.

**Checklist**
- [ ] Unit tests

closes #280

### DIFF
--- a/tests/system/cases/datasources/main.tf
+++ b/tests/system/cases/datasources/main.tf
@@ -5,33 +5,57 @@ provider "layer0" {
 }
 
 resource "layer0_environment" "datasources" {
-  name = "datasources"
+  name = "dsrctest"
 }
 
-module "sts" {
-  source         = "../modules/sts"
-  environment_id = "${layer0_environment.datasources.id}"
+resource "layer0_load_balancer" "datasources" {
+  name        = "dsrctest_lb"
+  environment = "${layer0_environment.datasources.id}"
+
+  port {
+    host_port      = 80
+    container_port = 80
+    protocol       = "http"
+  }
+}
+
+resource "layer0_service" "datasources" {
+  name          = "dsrctest_svc"
+  environment   = "${layer0_environment.datasources.id}"
+  deploy        = "${layer0_deploy.datasources.id}"
+  load_balancer = "${layer0_load_balancer.datasources.id}"
+  scale         = "1"
+  wait          = true
+}
+
+resource "layer0_deploy" "datasources" {
+  name    = "dsrctest_dpl"
+  content = "${data.template_file.datasources.rendered}"
+}
+
+data "template_file" "datasources" {
+  template = "${file("${path.module}/Dockerrun.aws.json")}"
 }
 
 data "layer0_environment" "datasources" {
   depends_on = ["layer0_environment.datasources"]
-  name       = "datasources"
+  name       = "${layer0_environment.datasources.name}"
 }
 
 data "layer0_deploy" "datasources" {
-  depends_on = ["module.sts"]
-  name       = "${module.sts.deploy_name}"
-  version    = "1"
+  depends_on = ["layer0_deploy.datasources"]
+  name       = "${layer0_deploy.datasources.name}"
+  version    = "${layer0_deploy.datasources.version}"
 }
 
 data "layer0_service" "datasources" {
-  depends_on     = ["module.sts"]
-  name           = "${module.sts.service_id}"
+  depends_on     = ["layer0_service.datasources"]
+  name           = "${layer0_service.datasources.name}"
   environment_id = "${layer0_environment.datasources.id}"
 }
 
 data "layer0_load_balancer" "datasources" {
-  depends_on     = ["module.sts"]
-  name           = "${module.sts.load_balancer_id}"
+  depends_on     = ["layer0_load_balancer.datasources", "layer0_service.datasources"]
+  name           = "${layer0_load_balancer.datasources.id}"
   environment_id = "${layer0_environment.datasources.id}"
 }
--- a/tests/system/cases/datasources/main.tf
+++ b/tests/system/cases/datasources/main.tf
@@ -9,7 +9,7 @@ resource "layer0_environment" "datasources" {
 }
 
 resource "layer0_load_balancer" "datasources" {
-  name        = "dsrctest_lb"
+  name        = "dsrctest"
   environment = "${layer0_environment.datasources.id}"
 
   port {
@@ -20,21 +20,17 @@ resource "layer0_load_balancer" "datasources" {
 }
 
 resource "layer0_service" "datasources" {
-  name          = "dsrctest_svc"
+  name          = "dsrctest"
   environment   = "${layer0_environment.datasources.id}"
   deploy        = "${layer0_deploy.datasources.id}"
   load_balancer = "${layer0_load_balancer.datasources.id}"
   scale         = "1"
-  wait          = true
+  wait          = false
 }
 
 resource "layer0_deploy" "datasources" {
-  name    = "dsrctest_dpl"
-  content = "${data.template_file.datasources.rendered}"
-}
-
-data "template_file" "datasources" {
-  template = "${file("${path.module}/Dockerrun.aws.json")}"
+  name    = "dsrctest"
+  content = "${file("${path.module}/Dockerrun.aws.json")}"
 }
 
 data "layer0_environment" "datasources" {
--- a/tests/system/cases/datasources/main.tf
+++ b/tests/system/cases/datasources/main.tf
@@ -1,0 +1,37 @@
+provider "layer0" {
+  endpoint        = "${var.endpoint}"
+  token           = "${var.token}"
+  skip_ssl_verify = true
+}
+
+resource "layer0_environment" "datasources" {
+  name = "datasources"
+}
+
+module "sts" {
+  source         = "../modules/sts"
+  environment_id = "${layer0_environment.datasources.id}"
+}
+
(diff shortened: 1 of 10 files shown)


tests/system/cases/datasources/main.tf
```diff
@@ -0,0 +1,61 @@
...
+  }
+}
+
+resource "layer0_service" "datasources" {
+  name          = "dsrctest_svc"
+  environment   = "${layer0_environment.datasources.id}"
+  deploy        = "${layer0_deploy.datasources.id}"
+  load_balancer = "${layer0_load_balancer.datasources.id}"
+  scale         = "1"
+  wait          = true
```
zpatrick: don't actually need to wait on this test

tests/system/cases/datasources/main.tf
```diff
@@ -0,0 +1,61 @@
...
+}
+
+resource "layer0_deploy" "datasources" {
+  name    = "dsrctest_dpl"
+  content = "${data.template_file.datasources.rendered}"
```
zpatrick: [nit] we don't really need a `template_file` in this case - we could just have:
```
content = "${file("${path.module}/Dockerrun.aws.json")}"
```

tests/system/cases/datasources/main.tf
```diff
@@ -0,0 +1,61 @@
...
+}
+
+data "layer0_service" "datasources" {
+  depends_on     = ["layer0_service.datasources"]
+  name           = "${layer0_service.datasources.name}"
+  environment_id = "${layer0_environment.datasources.id}"
+}
+
+data "layer0_load_balancer" "datasources" {
+  depends_on     = ["layer0_load_balancer.datasources", "layer0_service.datasources"]
```
zpatrick: why does this depend on the service?

tests/system/datasources_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+	// Compare outputs of data and resource values (resource
+	// values have the '_expected' suffix)
+	checkOutput := func(key string) {
+		log.Debugf("Checking data source vs resource output for key: %s", key)
+		datasourceOutput := s.Terraform.Output(key)
+		resourceOutput := s.Terraform.Output(key + "_expected")
+
+		if datasourceOutput != resourceOutput {
+			log.Fatalf(
```
zpatrick: Don't use `log` during tests to handle errors - it's just used for giving output while the test is running.

Use `t` to handle errors: `t.Error()`, `t.Fatal()`, etc (see: https://golang.org/pkg/testing/#T).

tests/system/datasources_test.go
```diff
@@ -0,0 +1,57 @@
...
+	defer s.Terraform.Destroy()
+
+	// Compare outputs of data and resource values (resource
+	// values have the '_expected' suffix)
+	checkOutput := func(key string) {
+		log.Debugf("Checking data source vs resource output for key: %s", key)
+		datasourceOutput := s.Terraform.Output(key)
+		resourceOutput := s.Terraform.Output(key + "_expected")
+
+		if datasourceOutput != resourceOutput {
```
zpatrick: We tend to follow  a common pattern during tests when we want to test if two things are equal:

```
if v, want := 1, 2; v != want { 
    t.Errorf("thing was %v, expected %v", v, want) 
}
```

so most of `checkOutput` could be condensed to:
```
if v, want := s.Terraform.Output(key), s.Terraform.Output(key+"_expected"); v != want {
    t.Errorf("%s: value was '%s', expected '%s'", key, v, want)
}
```

tests/system/datasources_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+	//check environment outputs
+	checkOutput("environment_id")
+	checkOutput("environment_size")
+	checkOutput("environment_min_count")
+	checkOutput("environment_os")
+	checkOutput("environment_ami")
+
+	//check deploy output
+	checkOutput("deploy_id")
```
zpatrick: shouldn't `version` and `name` be checked?

zpatrick: also, I'd recommend making sure using `version=latest` works for the deploy data source .

tests/system/datasources_test.go
```diff
@@ -0,0 +1,57 @@
...
+	checkOutput("load_balancer_service_name")
+
+	//check service outputs
+	checkOutput("service_id")
+	checkOutput("service_environment_name")
+	checkOutput("service_lb_name")
+	checkOutput("service_lb_id")
+	checkOutput("service_scale")
+
+	log.Debugf("L0 Terraform Provider Data sources Tests completed.")
```
zpatrick: don't need this line, the go testing package does this for us

tests/system/datasources_test.go
```diff
@@ -0,0 +1,57 @@
+package system
+
+import (
+	"testing"
+)
+
```
zpatrick: System tests are supposed to have a small comment that says which resources it creates, see: https://github.com/quintilesims/layer0/blob/develop/tests/system/serviceScale_test.go#L9

tests/system/datasources_test.go
```diff
@@ -0,0 +1,57 @@
...
+	checkOutput("environment_os")
+	checkOutput("environment_ami")
+
+	//check deploy output
+	checkOutput("deploy_id")
+
+	//check load balancer outputs
+	checkOutput("load_balancer_id")
+	checkOutput("load_balancer_name")
+	checkOutput("load_balancer_environment_name")
```
zpatrick: [go-ism?] I like the `checkOutput` pattern - not advocating that it needs to change or anything - but I did want to show a pattern that lot of go devs use for cases like this: using a slice of anonymous structs to create a table test:

````
outputPairs := []struct{
    DataOutputKey     string
    ResourceOutputKey string
 }{
    {
        DataOutputKey:     "environment_id",
        ResourceOutputKey: "environment_id_expected",
    },
    {
        DataOutputKey:     "service_id",
        ResourceOutputKey: "service_id_expected",
    },
...
}

for _, o := range outputPairs {
    if v, want := s.Terraform.Output(o.DataOutputKey), s.Terraform.Output(o.ResourceOutputKey); v != want {
        t.Errorf("%s: value was '%s', expected '%s'", key, v, want)
    }
}
````

Since we only need `string:string` pairs here, we could just use a `map[string]string` instead of anonymous structs, but that's obviously a special case.

File: tests/system/cases/datasources/main.tf [hunk 10]
```diff
@@ -0,0 +1,61 @@
...
+  token           = "${var.token}"
+  skip_ssl_verify = true
+}
+
+resource "layer0_environment" "datasources" {
+  name = "dsrctest"
+}
+
+resource "layer0_load_balancer" "datasources" {
+  name        = "dsrctest_lb"
```
zpatrick: [nit] personal preference, but I don't like putting entity-type suffix in names since it's redundant. It makes since for introductory walkthroughs, but not a pattern I'd expect/want seasoned layer0 users to do.

tests/system/cases/datasources/main.tf
sesh-kebab: As in wait is the default? I explicitly set wait to false which doesn't work as I needed data to be read after the resource creation was complete, especially when accessing service.load_balancer or load_balancer.service. So I explicitly set it to true to (despite it being the default).

But as I've removed load balancer <-> service links, I agree we no longer need the wait.
✅

tests/system/cases/datasources/main.tf
sesh-kebab: ✅

tests/system/cases/datasources/main.tf
sesh-kebab: Same as the load balancer <-> service reason stated above. But will remove as it is no longer needed.

tests/system/datasources_test.go
sesh-kebab: ✅

tests/system/datasources_test.go
sesh-kebab: ✅ noice!

tests/system/datasources_test.go
sesh-kebab: ✅

tests/system/datasources_test.go
sesh-kebab: Yes they should (didn't include all the arguments for some reason). ✅

tests/system/cases/datasources/main.tf
zpatrick: oh I see. Side note - a while ago you mentioned that wait probably shouldn't even be an option. I agree - should probably make that an issue at some point.

tests/system/datasources_test.go
sesh-kebab: 👍 good to know. add one more go-ism to my tool belt. actually in this particular case, I was thinking of about doing something like this, https://github.com/quintilesims/layer0/blob/develop/plugins/terraform/data_source_layer0_api.go#L45-L50 but figured, it would end up just being an array really as map would have redundant information.

tests/system/cases/datasources/main.tf
sesh-kebab: ✅

zpatrick:
:+1: LGTM

PR #289, opened 2017-07-14
Upload tfstate to s3 by default when 'l0-setup apply' is executed

### LINKED ISSUE #271: L0-Setup no longer automatically uploads tfstate file to s3
### Expected behavior
L0-setup transparently makes a copy of the state files in s3 when creating a new Layer0 instance.

### Actual behavior
As of v0.10.0 it no longer does this and requires explicitly running a `push` command.

Create a new layer0 environment using l0-setup. Confirm the s3 bucket corresponding to the new environment doesn't contain tfstate information.

Although the behavior has changed, the issue isn't stating that it is incorrect. Rather the issue is whether we should re-visit the decision to not automatically upload the tfstate files for a newly created l0 environment or not.

sesh-kebab:
v0.10.0 no longer uploaded the tfstate generated locally to s3. This PR changes that behavior to push the tfstate to s3 by default. Additionally an optional flag `--push=false` has been enabled which can be used to maintain existing behavior when executing `l0-setup apply <name>`.

Create a new Layer0 instance by executing the following commands:
```
l0-setup init test
l0-setup apply --push=false test
```

Confirm the s3 bucket created for the new Layer0 instance doesn't contain tfstate files.

Execute the apply command again, without the `--push` flag this time.
```
l0-setup apply test
```

Confirm the s3 bucket created for the previously created Layer0 instance does contain tfstate files.

**Checklist**
- [ ] Unit tests

closes #271

### DIFF
--- a/docs-src/docs/reference/setup-cli.md
+++ b/docs-src/docs/reference/setup-cli.md
@@ -59,7 +59,7 @@ There are no options for this command
 
 ---
 ## Apply
-The **apply** command is used to create and update Layer0 instances. Note that the default behavior of apply is to push the layer0 configuration to an S3 bucket unless the `--push=false` flag is set to false. Pushing the configuration to an S3 bucket requires aws crendentials which if not set via the optional `--aws-*` flags, are read from the environment variables or a credentials file. 
+The **apply** command is used to create and update Layer0 instances. Note that the default behavior of apply is to push the layer0 configuration to an S3 bucket unless the `--push=false` flag is set to false. Pushing the configuration to an S3 bucket requires aws credentials which if not set via the optional `--aws-*` flags, are read from the environment variables or a credentials file. 
 
 ### Usage
 ```
--- a/docs-src/docs/reference/setup-cli.md
+++ b/docs-src/docs/reference/setup-cli.md
@@ -68,7 +68,7 @@ $ l0-setup apply [options] <instance_name>
 
 ### Options
 * `--quick` - Skips verification checks that normally run after `terraform apply` has completed
-* `--push` - Skips uploading local tfstate file to s3. (default: true)
+* `--push` - Skips uploading local Layer0 configuration files to an S3 bucket
 
 ---
 ## List
--- a/docs-src/docs/reference/setup-cli.md
+++ b/docs-src/docs/reference/setup-cli.md
@@ -59,7 +59,7 @@ There are no options for this command
 
 ---
 ## Apply
-The **apply** command is used to create and update Layer0 instances.
+The **apply** command is used to create and update Layer0 instances. Note that the default behavior of apply is to push the layer0 configuration to an S3 bucket unless the `--push=false` flag is set to false. Pushing the configuration to an S3 bucket requires aws crendentials (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`), either set as environment variables or a credentials file. 
 
 ### Usage
 ```
--- a/docs-src/docs/reference/setup-cli.md
+++ b/docs-src/docs/reference/setup-cli.md
@@ -68,6 +68,7 @@ $ l0-setup apply [options] <instance_name>
 
 ### Options
 * `--quick` - Skips verification checks that normally run after `terraform apply` has completed
+* `--push` - Skips uploading local tfstate file to s3. (default: true)
 
 ---
 ## List
--- a/docs-src/docs/reference/setup-cli.md
+++ b/docs-src/docs/reference/setup-cli.md
@@ -59,7 +59,7 @@ There are no options for this command
 
 ---
 ## Apply
-The **apply** command is used to create and update Layer0 instances. Note that the default behavior of apply is to push the layer0 configuration to an S3 bucket unless the `--push=false` flag is set to false. Pushing the configuration to an S3 bucket requires aws crendentials (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`), either set as environment variables or a credentials file. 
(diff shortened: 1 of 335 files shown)


setup/command/apply.go
```diff
@@ -16,15 +16,24 @@ func (f *CommandFactory) Apply() cli.Command {
...
+				Usage: "setting it to false skips pushing local tfstate to s3 (default: true)",
+			},
 		},
 		Action: func(c *cli.Context) error {
 			args, err := extractArgs(c.Args(), "NAME")
 			if err != nil {
 				return err
 			}
 
+			provider, err := f.newAWSProviderHelper(c)
```
zpatrick: Ok, so in order for `push` or `pull`  commands to work, the user needs to give credentials for AWS. This can be done through cli flags (the `awsFlags` mentioned earlier), or environment variables.  What we used to do in the old l0-setup was grab the aws keys from whatever they entered during [the equivalent of] `l0-setup init`. 

This is nice for the user, but introduced a ton of complexities and bad coupling patterns in the codebase.
I intentionally took out that behavior so we don't have to deal with it anymore. 

But now we have a UX issue - if we `push` after every `apply`, then user needs to make sure they input aws credentials every time they run `apply`. This seems acceptable for `push`, but I could see customers complaining about `apply`.

docs-src/docs/reference/setup-cli.md
```diff
@@ -68,6 +68,7 @@ $ l0-setup apply [options] <instance_name>
 
 ### Options
 * `--quick` - Skips verification checks that normally run after `terraform apply` has completed
+* `--push` - Skips uploading local tfstate file to s3. (default: true)
```
zpatrick: This applies here, but also is a general statement:

L0-setup `push` and `pull` is not just for `terraform.tfstate`! 
These commands sync **everything** in the `~/.layer0/<instance>` directory to/from S3. 
This is the primary way we should recommend users add custom configurations to their Layer0 instances.

sesh-kebab: I incorrectly assumed we weren't exposing that detail to the user but checking the documentation [push](http://layer0.ims.io/reference/setup-cli/#push), it looks like we do. Will update to match description already in documentation.

> The push command is used to back up your Layer0 configuration files to an S3 bucket.

setup/command/apply.go
```diff
@@ -16,15 +16,24 @@ func (f *CommandFactory) Apply() cli.Command {
...
+			if err != nil {
+				return err
+			}
+
 			instance := f.NewInstance(args["NAME"])
-			if err := instance.Apply(!c.Bool("quick")); err != nil {
+			if err := instance.Apply(!c.Bool("quick"), provider.S3, c.Bool("push")); err != nil {
```
sesh-kebab: ✅

setup/command/apply.go
sesh-kebab: So as I understand it, the current code path is, if the awsFlags aren't explicitly passed in, we would read fall back on `aws.Config.Credentials`which searches for credentials in env variables, credential files etc.

- Is this the desired behavior?
- If so, can we consider UX acceptable?

zpatrick: > Is this the desired behavior?

Yes

> Can we consider UX acceptable?

I'm not sure - perhaps we should ask the chunnels team

setup/command/apply.go
```diff
@@ -16,18 +16,33 @@ func (f *CommandFactory) Apply() cli.Command {
 				Name:  "quick",
```
zpatrick: You still needs to add in the `awsFlags` here

setup/command/apply.go
```diff
@@ -16,18 +16,33 @@ func (f *CommandFactory) Apply() cli.Command {
...
+				return err
+			}
+
 			instance := f.NewInstance(args["NAME"])
 			if err := instance.Apply(!c.Bool("quick")); err != nil {
 				return err
 			}
 
+			if c.Bool("push") {
+				if err := instance.Push(provider.S3); err != nil {
```
zpatrick: Let's move the block:
```
provider, err := f.newAWSProviderHelper(c)
if err != nil {
    return err
}
```

into the `if c.Bool("push") {}` block. Reason being: we don't want the error from ` f.newAWSProviderHelper(c)` to occur if the user specifies `--push=false`

setup/command/apply.go
sesh-kebab: I enquired with the chunnel team and can confirm that they ensure that aws env vars are set. Based on that information I believe it should not break their automation.

zpatrick: Ok cool - can I trouble ya to add that line somewhere in the docs for `l0-setup apply` :smile: ?

sesh-kebab: sure - done.

setup/command/apply.go
sesh-kebab: ✅

setup/command/apply.go
sesh-kebab: ✅

docs-src/docs/reference/setup-cli.md
```diff
@@ -59,7 +59,7 @@ There are no options for this command
 
 ---
 ## Apply
-The **apply** command is used to create and update Layer0 instances.
+The **apply** command is used to create and update Layer0 instances. Note that the default behavior of apply is to push the layer0 configuration to an S3 bucket unless the `--push=false` flag is set to false. Pushing the configuration to an S3 bucket requires aws crendentials which if not set via the optional `--aws-*` flags, are read from the environment variables or a credentials file.
```
zpatrick: typo "crendentials"

PR #292, opened 2017-07-18
WIP: Issue 290 credential helper

### LINKED ISSUE #290: Add support for docker credential helper
There are docker repositories that require custom docker credential helpers to handle authentication to push/pull images or to store credentials more securely than in plain text. As a Layer0 user, it would be useful if I could I specify the path to a credential helper, similar to how the user currently specifies the path to docker config file, during `l0-setup init`.

I would anticipate this working exactly the same as the docker config file:

1. During `l0-setup` the user specifies the path to a docker credential helper
`l0-setup init --docker-cred-helper ~/my-cred-helper abc`
2. Executing `l0-setup apply abc` would copy the helper to the s3 bucket (as the docker config file would)
3. Userdata.sh (or Userdata.ps1 as in the case of windows containers) would be updated to pull the helper to the machine and copy it to an executable path such as `/usr/local/bin`.

sesh-kebab:
This pull request enables support for docker credential helper that can be used to authenticate against a private docker repository.

**Overview**
1. It does so by allowing the user to specify the path to a docker credential helper, much like how the user currently specifies the path to a docker config file, when executing `l0-setup init`.
2. The docker helper is then copied to the s3 bucket/terraform
3. Userdata.sh pulls down all files from the s3 bucket that have a prefix of docker-credential- and copy it to /usr/local/bin
4. This is assuming that the user would use a docker config file with the following config
```
{
    "credsStore": "ecr-login"
}
```

**WIP:**

- Credential support has only been added for linux layer0 environments. I would like to get feedback on the general direction of the change and iterate as need be before making relevant changes to windows layer0 environments also.

- The docker helper probably needs to be copied to the bootstrap folder rather than the terraform folder as copying to the terraform folder is optional?

closes #290

### DIFF
--- a/api/backend/ecs/environment_manager.go
+++ b/api/backend/ecs/environment_manager.go
@@ -566,12 +566,13 @@ Add-JobTrigger -Name $jobname -Trigger (New-JobTrigger -AtStartup -RandomDelay 0
 
 var defaultLinuxUserDataTemplate = []byte(
 	`#!/bin/bash
-    echo ECS_CLUSTER={{ .ECSEnvironmentID }} >> /etc/ecs/ecs.config
-    echo ECS_ENGINE_AUTH_TYPE=dockercfg >> /etc/ecs/ecs.config
-    yum install -y aws-cli awslogs jq
-    aws s3 cp s3://{{ .S3Bucket }}/bootstrap/dockercfg dockercfg
-    cfg=$(cat dockercfg)
-    echo ECS_ENGINE_AUTH_DATA=$cfg >> /etc/ecs/ecs.config
-    docker pull amazon/amazon-ecs-agent:latest
-    start ecs
+echo ECS_CLUSTER={{ .ECSEnvironmentID }} >> /etc/ecs/ecs.config
+echo ECS_ENGINE_AUTH_TYPE=dockercfg >> /etc/ecs/ecs.config
+yum install -y aws-cli awslogs jq
+aws s3 cp s3://{{ .S3Bucket }}/bootstrap/dockercfg dockercfg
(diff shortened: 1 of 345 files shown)


zpatrick:
I don't believe this is supported by ECS: http://docs.aws.amazon.com/AmazonECS/latest/developerguide/private-auth.html

sesh-kebab:
The documentation certainly does indicate that it isn't supported but what I've observed during testing indicate otherwise. I will investigate further to figure out how the ecs agent is parsing the docker config files to pull images from private repositories.

diemonster:
@zpatrick there's a few blog posts from AWS that suggest `credsStore` is a thing:
(for example) https://aws.amazon.com/blogs/compute/authenticating-amazon-ecr-repositories-for-docker-cli-with-credential-helper/

sesh-kebab:
This PR fortunately isn't needed. The ecs-agent has built-in support to pull ECR images based on IAM credentials. Furthermore, the ecs-agent has built in support for caching tokens and automatically re-authenticating to pull images as needed.

Initializing ECR auth provider
https://github.com/aws/amazon-ecs-agent/blob/v1.14.3/agent/engine/docker_container_engine.go#L332-L342

ECR which calls AWS CLI GetAuthorizationToken to return the docker credentials needed to login
https://github.com/aws/amazon-ecs-agent/blob/v1.14.3/agent/engine/dockerauth/ecr.go#L50

I made some incorrect assumptions here though:
- we can use Route53 to create a custom domain for an ECR(epository)

Without the above, we would need a reverse proxy service like we have with d.ims.io if we wanted to have images with custom domain names. But it does pose the question, are we ok to put up infrastructure in front of a managed service (ECR) if we don't strictly need to? Although the developer experience is vastly better with d.ims.io over aws cli for development; in production, do we really want to introduce a single point of failure?

PR #300, opened 2017-07-24
Allow importing of layer0 terraform resources

### LINKED ISSUE #297: Terraform module: allow importing of resources
This would allow us to import existing layer0 resources into a terraform configuration, e.g.:
```
terraform import layer0_environment.foo foo123
```

I believe we only need to add a couple lines to each resource's schema:
```
return &schema.Resource{
		...
		Importer: &schema.ResourceImporter{
			State: schema.ImportStatePassthrough,
		},
                ...
```

Sources:
* https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_elb.go#L27
* https://www.terraform.io/docs/plugins/provider.html#importer

zpatrick:
* Allows existing layer0 resources to be imported to terraform configurations using `terraform import`
* Removes `wait` field from `layer0_service` resource, as it doesn't mesh with other terraform resources / is a field that cannot be 'read' from the api. 
* Populate the `content` field of `layer0_deploy` during read
* Includes DiffSuppression of the `deploy.content` field - this helps to more accurately calculate the diff between two deploys. 

System test

**Checklist**
- [ ] ~Unit tests~

Closes #297


zpatrick:
@seshi: updated docs and fixed broken tests

PR #303, opened 2017-07-27
Consistently catch <Entity> does not exist errors

### LINKED ISSUE #113: API: consistently catch "EntityDoesNotExist" errors
Our API isn't consistent on catching if an entity does not exist. 
This it makes it difficult for the terraform plugin to be idempotent when creating/deleting entities.

For example, if a service and environment are deleted, calling GetService will return "cannot find environment id for service", instead of the expected API error with a "ServiceDoesNotExist" error code. 
We should check that entities existing before operating on them in the logic layer.

zpatrick:
API returns `<Entity>DoesNotExist` error code when GET requests are made on an entity that does not exist. This is captured more explicitly in the Terraform plugin, allowing Terraform to gracefully handle when entities don't exist. 

1. Create entities using Terraform
2. Delete the entities outside of Terraform (e.g. the CLI)
3. Run `terraform plan`; it should show that the deleted entities are marked for re-creation.   

closes #113

### DIFF
--- a/cli/client/errors.go
+++ b/cli/client/errors.go
@@ -18,10 +18,10 @@ func sslError(err error) error {
 
 type ServerError models.ServerError
 
-func (this *ServerError) Error() string {
-	return this.Message
+func (s *ServerError) Error() string {
+	return s.Message
 }
 
-func (this *ServerError) ToCommonError() *errors.ServerError {
-	return errors.Newf(errors.ErrorCode(this.ErrorCode), this.Message)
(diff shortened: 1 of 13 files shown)


cli/client/errors.go
```diff
@@ -20,3 +21,7 @@ type ServerError models.ServerError
 func (this *ServerError) Error() string {
```
sesh-kebab: `this` doesn't seem to be consistent with style guidelines.
https://github.com/golang/go/wiki/CodeReviewComments#receiver-names

PR #305, opened 2017-08-01
Add leftover tags janitor and handle resolving leftover tags in CLI.

### LINKED ISSUE #140: Tasks get removed over time by ECS, leaves leftover tags in Layer0
By default, tasks get cleaned up in ECS 3 hours after they have completed (see: http://docs.aws.amazon.com/AmazonECS/latest/developerguide/automated_image_cleanup.html). However, the task's tags in Layer0 aren't cleaned up unless a user runs `l0 task delete`. 

This leads to issues when resolving tasks in the cli. For example, if you have any leftover tags you will get the following behavior:
```
$ l0 task get \*
Task 123abc456 does not exist
```

One should also note that the "problem" tag won't show up if a user runs `$ l0 task list`, since that command doesn't use tags to resolve anything.

For a simple solution, we should have a job in the layer0 api that deletes these leftover tags. 
For a more robust solution, we should check for and handle this error in the cli.

sesh-kebab:
Addresses the situation where the dynamodb task ends up with leftover tags which don't correspond to a ecs task. This causes the l0 cli to return errors when when executing commands like `l0 task get \*`. This PR addresses the CLI and the leftover Tags in 2 parts:

**CLI:**
Return a debug message when iterating over tasks and a task is not found.

**API:**
Add a new Tags Janitor that searches for and deletes leftover tags as result of ECS Tasks being removed, every 10 minutes.

---

Unit Test: `go test`

You can also simulate the left over task tag with the following steps:

1. Create a one off task in layer0
2. Ensure a corresponding tag exists in the dynamo db table l0_<l0_prefix>_tags table
3. Duplicate the corresponding tag item and update the `EntityID` with random suffix
4. Run the api locally `go run api/main.go`
5. Execute `l0 -d task get \*` to confirm:
  a. cli doesn't crash
  b. debug flag correctly mentions tag without a corresponding task
5. Check that the duplicated tag has been deleted from the random suffix in DynamoDB

---

**Checklist**
- [x] Unit tests

closes #140

### DIFF
--- a/api/logic/tag_janitor.go
+++ b/api/logic/tag_janitor.go
@@ -1,7 +1,6 @@
 package logic
 
 import (
-	"fmt"
 	"time"
 
 	"github.com/quintilesims/layer0/common/db/tag_store"
@@ -65,14 +64,13 @@ func (t *TagJanitor) pulse() error {
 
 	errs := []error{}
 	for _, tag := range tags {
-		fmt.Printf("tag: %s, %s, %s, %s \n", tag.EntityID, tag.EntityType, tag.Key, tag.Value)
 		if !taskExists(tag.EntityID) {
 			if err := t.TagStore.Delete(tag.EntityType, tag.EntityID, tag.Key); err != nil {
-				tagLogger.Errorf("Could not delete tag (id: %s, type: %s, key: %s) -  %s\n",
-					tag.EntityID, tag.EntityType, tag.Key, err.Error())
-			} else {
-				tagLogger.Infof("Tag for task (%s) has been deleted\n", tag.EntityID)
+				tagLogger.Errorf("Could not delete tag (%#v) -  %s\n", tag, err.Error())
+				continue
 			}
+
+			tagLogger.Infof("Tag for task (%s) has been deleted\n", tag.EntityID)
 		}
 	}
 
--- a/api/logic/tag_janitor.go
+++ b/api/logic/tag_janitor.go
@@ -1,0 +1,80 @@
+package logic
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/quintilesims/layer0/common/db/tag_store"
+	"github.com/quintilesims/layer0/common/errors"
+	"github.com/quintilesims/layer0/common/logutils"
+	"github.com/quintilesims/layer0/common/waitutils"
+)
+
+const (
+	taskJanitorSleepDuration = time.Minute * 10
+)
+
+var tagLogger = logutils.NewStackTraceLogger("Tags Janitor")
+
+type TagJanitor struct {
+	TaskLogic TaskLogic
+	TagStore  tag_store.TagStore
+	Clock     waitutils.Clock
+}
+
+func NewTagJanitor(taskLogic TaskLogic, tagStore tag_store.TagStore) *TagJanitor {
+	return &TagJanitor{
+		TaskLogic: taskLogic,
+		TagStore:  tagStore,
+		Clock:     waitutils.RealClock{},
+	}
+}
+
+func (t *TagJanitor) Run() {
+	go func() {
+		for {
+			tagLogger.Info("Starting cleanup")
+			t.pulse()
+			tagLogger.Infof("Finished cleanup")
+			t.Clock.Sleep(taskJanitorSleepDuration)
+		}
+	}()
+}
+
+func (t *TagJanitor) pulse() error {
+	tasks, err := t.TaskLogic.ListTasks()
+	if err != nil {
+		tagLogger.Errorf("Failed to list tasks: %v", err)
+		return err
+	}
+
+	taskExists := func(id string) bool {
+		for _, t := range tasks {
+			if t.TaskID == id {
+				return true
+			}
+		}
+
+		return false
+	}
+
+	tags, err := t.TagStore.SelectByType("task")
+	if err != nil {
+		tagLogger.Errorln("Could not query for tag store for task entity type - ", err.Error())
+	}
+
+	errs := []error{}
+	for _, tag := range tags {
+		fmt.Printf("tag: %s, %s, %s, %s \n", tag.EntityID, tag.EntityType, tag.Key, tag.Value)
+		if !taskExists(tag.EntityID) {
+			if err := t.TagStore.Delete(tag.EntityType, tag.EntityID, tag.Key); err != nil {
+				tagLogger.Errorf("Could not delete tag (id: %s, type: %s, key: %s) -  %s\n",
+					tag.EntityID, tag.EntityType, tag.Key, err.Error())
+			} else {
+				tagLogger.Infof("Tag for task (%s) has been deleted\n", tag.EntityID)
+			}
+		}
+	}
+
+	return errors.MultiError(errs)
+}
--- a/api/logic/tag_janitor_test.go
+++ b/api/logic/tag_janitor_test.go
@@ -1,0 +1,71 @@
+package logic
+
+import (
+	"testing"
+
+	"github.com/golang/mock/gomock"
+	"github.com/quintilesims/layer0/api/logic/mock_logic"
+	"github.com/quintilesims/layer0/common/db/tag_store"
+	"github.com/quintilesims/layer0/common/models"
+	"github.com/quintilesims/layer0/common/testutils"
+)
+
+func TestTagJanitorPulse(t *testing.T) {
+	ctrl := gomock.NewController(t)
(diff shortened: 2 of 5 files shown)


api/logic/tag_janitor.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	tags, err := t.TagStore.SelectByType("task")
+	if err != nil {
+		tagLogger.Errorln("Could not query for tag store for task entity type - ", err.Error())
+	}
+
+	errs := []error{}
+	for _, tag := range tags {
+		fmt.Printf("tag: %s, %s, %s, %s \n", tag.EntityID, tag.EntityType, tag.Key, tag.Value)
```
zpatrick: remove debug code

api/logic/tag_janitor.go
```diff
@@ -0,0 +1,80 @@
...
+		if !taskExists(tag.EntityID) {
+			if err := t.TagStore.Delete(tag.EntityType, tag.EntityID, tag.Key); err != nil {
+				tagLogger.Errorf("Could not delete tag (id: %s, type: %s, key: %s) -  %s\n",
```
zpatrick: [nit] you can shorthand printing all fields and values of a struct with `%#v`, e.g. `"Could not delete tag %#v - ..."`

api/logic/tag_janitor.go
```diff
@@ -0,0 +1,80 @@
...
+					tag.EntityID, tag.EntityType, tag.Key, err.Error())
+			} else {
```
zpatrick: [super nit] we tend to avoid creating nested blocks by using `continue` or `return` - e.g.
```
for _, tag := range tags {
    if err := thing(); err != nil {
        log.Errorf("Failure")
        continue
    }

    log.Info("Success")
}
```

api/logic/tag_janitor_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/quintilesims/layer0/common/models"
+	"github.com/quintilesims/layer0/common/testutils"
+)
+
+func TestTagJanitorPulse(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	taskLogicMock := mock_logic.NewMockTaskLogic(ctrl)
+	tagStore, tagsAdded := getTagStore()
```
zpatrick: I think this helper function could be improved by specifying which tags to insert instead of the other way around. That would make it easier to read and understand the test, e.g.:

(taken from https://github.com/quintilesims/layer0/blob/develop/common/db/tag_store/dynamo_test.go#L14)
```
taskTags := []models.Tag{
        {EntityID: "t1", EntityType: "task", Key: "name", Value: "tsk1"},
	{EntityID: "t1", EntityType: "task", Key: "environment_id", Value: "e1"},
	{EntityID: "t2", EntityType: "task", Key: "name", Value: "tsk2"},
	{EntityID: "t2", EntityType: "task", Key: "environment_id", Value: "e2"},
}

tagStore := getTagStore(taskTags)
...

testutils.AssertEqual(t, 2, len(tags))
testutils.AssertEqual(t, "t2", tags[0].EntityID)
testutils.AssertEqual(t, "t2", tags[1].EntityID)
```

cli/command/task_command.go
```diff
@@ -160,8 +161,28 @@ func (t *TaskCommand) Delete(c *cli.Context) error {
...
+			}
+		}
+
+		return false
+	}
+
 	tasks := []*models.Task{}
 	getTaskf := func(id string) error {
+		if !taskExists(id) {
+			log.Debugln("No corresponding Task found for TaskID:", id)
```
zpatrick: [super nit] we tend to use `f` instead of `ln` for printing for more explicitness

cli/command/task_command_test.go
```diff
@@ -190,6 +190,12 @@ func TestGetTask(t *testing.T) {
 		Return([]string{"id"}, nil)
 
 	tc.Client.EXPECT().
+		ListTasks().
+		Return([]*models.TaskSummary{
```
zpatrick: this looks gross and makes this test harder to read. Can you please split out the return value into its own variable?

cli/command/task_command_test.go
```diff
@@ -208,6 +214,10 @@ func TestGetTask_userInputErrors(t *testing.T) {
 		"Missing NAME arg": testutils.GetCLIContext(t, nil, nil),
 	}
 
+	tc.Client.EXPECT().
+		ListTasks().
+		Return(nil, nil)
+
 	for name, c := range contexts {
 		if err := command.Get(c); err == nil {
 			t.Fatalf("%s: error was nil!", name)
```
zpatrick: We should have a new test that verifies this new ignoring behavior

api/logic/tag_janitor.go
sesh-kebab: ✅

api/logic/tag_janitor.go
sesh-kebab: 👍

api/logic/tag_janitor.go
sesh-kebab: ✅

cli/command/task_command.go
sesh-kebab: Not sure what you mean by more "explicitness"?

Given `log.Debugln("something", id)` vs `log.Debugf("something %s\n", id)`; are you suggesting the latter?

cli/command/task_command_test.go
sesh-kebab: ✅

cli/command/task_command_test.go
sesh-kebab: ✅

cli/command/task_command.go
zpatrick: yes, explicitness in terms of which print 'verb' (their term) we use: `%s` vs `%v` vs `%#v` etc. etc.

sesh-kebab: ✅

api/logic/tag_janitor_test.go
```diff
@@ -39,33 +48,17 @@ func TestTagJanitorPulse(t *testing.T) {
 
 	//expecting 3 tags for task1 to be deleted from the tagstore
 	testutils.AssertEqual(t, tagsAdded-3, len(tags))
-	testutils.AssertEqual(t, "task2", tags[0].EntityID)
+	testutils.AssertEqual(t, "t2", tags[0].EntityID)
 }
 
-func getTagStore() (tag_store.TagStore, int) {
+func getTagStore(tags []models.Tag) (tag_store.TagStore, int) {
```
zpatrick: not sure why you need to return the `tasksAdded int` since the user knows this from the input param?

cli/command/task_command_test.go
```diff
@@ -184,16 +184,17 @@ func TestGetTask(t *testing.T) {
...
 	tc.Resolver.EXPECT().
 		Resolve("task", "name").
 		Return([]string{"id"}, nil)
 
 	tc.Client.EXPECT().
 		ListTasks().
-		Return([]*models.TaskSummary{
-			{TaskID: "id"},
-		}, nil)
+		Return(result, nil)
```
zpatrick: :sweat_smile: much better

cli/command/task_command_test.go
```diff
@@ -225,6 +226,33 @@ func TestGetTask_userInputErrors(t *testing.T) {
...
+		Return([]string{"id3", "id4", "id5"}, nil)
+
+	tc.Client.EXPECT().
+		ListTasks().
+		Return(result, nil)
+
+	tc.Client.EXPECT().
+		GetTask(gomock.Any()).
+		Return(&models.Task{}, nil).
+		Times(1) //only task 'id3' should result in a gettask call
```
zpatrick: I like the comment, but the `Times(1)` is actually redundant with the `gomock` framework (you need to explicitly put `AnyTimes()` to allow the behavior you're looking to avoid)

api/logic/tag_janitor_test.go
sesh-kebab: yea I agree - it is redundant. i thought having an explicit variable named tagsAdded in makes the intent clearer. I don't feel strongly about it though. remove?
```
testutils.AssertEqual(t, tagsAdded-3, len(tags))
vs
testutils.AssertEqual(t, len(taskTags)-3, len(tags))
```

zpatrick: the former reads better i think

PR #306, opened 2017-08-01
Return '404 Not Found' errors when resource not found

### LINKED ISSUE #304: Returning HTTP 400 errors rather than the more appropriate 404
### Expected behavior
API endpoints for entities return unexpected codes like `200 Http OK`, `400 Bad Request` or other codes like `500 Internal Server Error` errors when entities can't be found.

### Actual behavior
The API should be returning `404 Not Found` error instead. The API was able to understand the request but just couldn't find the requested entity so the request should be considered well formed and valid.

l0 task get <task-name>

sesh-kebab:
Return 404 Http response when a resource is not found:

- Deploy
- Service
- Environment
- Job
- Load Balancer
- Task

Run API locally. Test the below endpoints with a random id value that doesn't exist. Confirm you receive a 404 rather than a 500.

1. GET, DELETE entity/{id} for deploy, service, environment, job, load balancer & task
2. PUT service/{id}/scale & service/{id}/deploy
3. GET service/{id}/logs & task/{id}/logs 

**Checklist**
- [ ] ~Unit tests~

closes #304

### DIFF
--- a/api/backend/ecs/task_manager.go
+++ b/api/backend/ecs/task_manager.go
@@ -179,7 +179,7 @@ func (this *ECSTaskManager) GetTaskLogs(environmentID, taskID, start, end string
 // Assumes the tasks are all of the same type
 func modelFromTasks(tasks []*ecs.Task) (*models.Task, error) {
 	if len(tasks) == 0 {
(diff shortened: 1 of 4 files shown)


zpatrick:
:+1: nice!

PR #312, opened 2017-08-08
Skeleton for Environment

zpatrick:
This includes a bunch of misc changes for how dealing with environment entities will work. 

I've cleared out a bunch of scheduler code so we can re-visit that once the other entities are done.


api/controllers/controller.go
```diff
@@ -0,0 +1,15 @@
+package controllers
```
diemonster: just noting that the relevant change here is renaming the previous `handlers` package to `controllers`

api/controllers/controller_test.go
```diff
@@ -0,0 +1,36 @@
+package controllers
```
diemonster: Far more simple than the previous handler tests, I definitely prefer this.

api/controllers/environment.go
```diff
@@ -28,7 +32,7 @@ func (e *EnvironmentController) CreateEnvironment(request *restful.Request, resp
 		// todo: handle error
 	}
 
-	environmentModel, err := environment.Read()
+	environmentModel, err := environment.Model()
```
diemonster: [nit] maybe `GetModel` versus `Model`?

diemonster: ahhh I see why you did `Model` when put in context with the other `Get` methods....

api/controllers/environment.go
```diff
@@ -37,10 +41,55 @@ func (e *EnvironmentController) CreateEnvironment(request *restful.Request, resp
...
+	WriteJobResponse(response, jobID)
 }
 
 func (e *EnvironmentController) GetEnvironment(request *restful.Request, response *restful.Response) {
+	id := request.PathParameter("id")
+	if id == "" {
+		// todo: handle error
+	}
+
+	environment := e.Provider.GetEnvironment(id)
```
diemonster: Discussions about the ambiguity of `Provider` aside, this is pretty clean from the calling code's point of view. slightly confusing to me to see an abstract object returned from `GetEnvironment` and then a model returned from just `Model`...

common/models/job.go
```diff
@@ -2,13 +2,15 @@ package models
...
+
+	"github.com/quintilesims/layer0/common/job"
 )
 
 type Job struct {
 	JobID       string            `json:"job_id"`
 	TaskID      string            `json:"task_id"`
-	JobStatus   int64             `json:"job_status"`
-	JobType     int64             `json:"job_type"`
+	JobStatus   job.JobStatus     `json:"job_status"`
```
diemonster: would this be a breaking change given that this is now a string?

api/controllers/environment.go
zpatrick: Open to using different verb if we can find one. Don't think this should be a blocker for this pr though

common/models/job.go
zpatrick: It's still an int64

diemonster: opps, disregard!

api/controllers/environment.go
diemonster: Best I can think of for now is `InitModel` or `NewModel`, I agree it's not a blocker for the PR

PR #314, opened 2017-08-09
replace restful w fireball

zpatrick:
Backend skeleton for environment coming after this


diemonster:
@sesh-kebab @zpatrick 

- Change is not required, but (in my mind) the gains are: a decrease in complexity (w our dependency in question ) and easier/less tedious interaction with our restful endpoints. 

The loses are: less provided features that we'd have to internally implement if we ended up needing them, decreased automatic integration with Swagger. 

But I think at this point, we know what Layer0's API does and what it's going to do --- much of what `go-restful` provides is simply not needed.

- There are myriad Go web frameworks, I don't believe there was an exhaustive investigation to determine a new one. That said, I personally like https://github.com/zpatrick/fireball because there's not much to it --- it wraps Go's stdlib in a reasonable way and it's easy to consume. I've also seen it work in production and can attest that it's performant. 

- There are definitely potential risks. We haven't yet benchmarked the existing API to see what sorts of performance loses we might encounter making this switch. Most of the risks stem from this, which suggests we should have some baseline numbers to make sure we're not shooting ourselves in the foot.

sesh-kebab:
Yea, just looking at the diff to see the equivalent code, agreed; fireball syntax is pretty sweet.

PR #318, opened 2017-08-10
Increase parametrization; b.Run() for multiple benchmarks

tlake:


tlake:
Given that we're iterating on work outside of this PR, maybe we should just close this PR since any changes from it are already being manually incorporated elsewhere?

PR #324, opened 2017-08-16
Create/Delete/Update for Environment Provider

zpatrick:
This continues work on the `EnvironmentProvider`, adds some tests and finishes the first (untested) implementation/patterns to use on the provider layer.


api/main.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/quintilesims/layer0/api/provider/aws"
 	awsclient "github.com/quintilesims/layer0/common/aws"
 	"github.com/quintilesims/layer0/common/config"
+	"github.com/quintilesims/layer0/common/db/tag_store"
```
sesh-kebab: Would it be possible to change the package name to something more concise without the underscore? Even `tags` would be sufficient as it is prefixed by `/common/db/` I would think?

https://golang.org/doc/effective_go.html#package-names

api/provider/aws/environment_create.go
```diff
@@ -83,10 +95,10 @@ func (e *EnvironmentProvider) Create(req models.CreateEnvironmentRequest) (*mode
 	return e.Read(environmentID)
 }
 
-func (e *EnvironmentProvider) createSG(environmentID, vpcID string) (*ec2.SecurityGroup, error) {
+func (e *EnvironmentProvider) createSG(groupName, description, vpcID string) (*ec2.SecurityGroup, error) {
 	input := &ec2.CreateSecurityGroupInput{}
-	input.SetGroupName(environmentID)
-	input.SetDescription(fmt.Sprintf("SG for Layer0 environment %s", environmentID))
+	input.SetGroupName(groupName)
```
sesh-kebab: Question: Why do we not initialize the struct inline rather than calling setters? Is it just a styling convention?

api/provider/aws/environment_create.go
```diff
@@ -19,60 +19,72 @@ func (e *EnvironmentProvider) Create(req models.CreateEnvironmentRequest) (*mode
...
-	// todo: use config to get instance name
 	environmentID := generateEntityID(req.EnvironmentName)
-	fqEnvironmentID := addLayer0Prefix("INSTANCE", environmentID)
-
-	// todo: use config to get vpc id
-	securityGroup, err := e.createSG(fqEnvironmentID, "VPC")
-	if err != nil {
-		return nil, err
-	}
+	fqEnvironmentID := addLayer0Prefix(e.Config.Instance(), environmentID)
```
sesh-kebab: Question: what does the `fq` prefix stand for?

api/provider/aws/environment_delete.go
```diff
@@ -11,14 +11,15 @@ import (
 )
 
 func (e *EnvironmentProvider) Delete(environmentID string) error {
-	//todo: use conifig.Instance()
-	fqEnvironmentID := addLayer0Prefix("INSTANCE", environmentID)
+	fqEnvironmentID := addLayer0Prefix(e.Config.Instance(), environmentID)
 
-	if err := e.deleteASG(fqEnvironmentID); err != nil {
+	autoScalingGroupName := fqEnvironmentID
```
sesh-kebab: This is a good way of indicating that fqEnvironmentID maps to asg name, launch config. 👍

api/provider/aws/environment_list_test.go
```diff
@@ -0,0 +1,73 @@
...
+	results := []models.EnvironmentSummary{
+		{EnvironmentID: "eid1"},
+		{EnvironmentID: "eid2"},
+	}
+
+	if err := environment.listTags(results); err != nil {
+		t.Fatal(err)
+	}
+
+	assert.Equal(t, "ename1", results[0].EnvironmentName)
```
sesh-kebab: Would it be a good test to also check for the number of items you are expecting to return based on the items you've added to tagstore?

api/provider/aws/environment_create.go
zpatrick: fully-qualified

api/main.go
zpatrick: yep, but let's do it in a separate issue

api/provider/aws/environment_list_test.go
zpatrick: added

api/provider/aws/environment_create.go
```diff
@@ -83,10 +95,10 @@ func (e *EnvironmentProvider) Create(req models.CreateEnvironmentRequest) (*mode
 	return e.Read(environmentID)
 }
 
-func (e *EnvironmentProvider) createSG(environmentID, vpcID string) (*ec2.SecurityGroup, error) {
+func (e *EnvironmentProvider) createSG(groupName, description, vpcID string) (*ec2.SecurityGroup, error) {
```
zpatrick: This function is doing to much - we should try to make helper functions wrap a single aws api call. This should be split into `createSG` and `allowSGIngress`

PR #326, opened 2017-08-16
Service and Task controllers

diemonster:
Adds implementation at the controller layer for Services and Task entities. 

Adds initial interfaces for Service and Task providers

Navigate to ./api and execute go test ./...

**Checklist**
- [x] Unit tests


api/controllers/service.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return fireball.NewJSONResponse(202, model)
+}
+
+func (s *ServiceController) DeleteService(c *fireball.Context) (fireball.Response, error) {
+	id := c.PathVariables["id"]
+	if err := s.ServiceProvider.Delete(id); err != nil {
```
sesh-kebab: Service deletion should be via a job right?

```
type JobType int64

const (
	DeleteEnvironmentJob JobType = 1 + iota
	DeleteServiceJob
	DeleteLoadBalancerJob
	DeleteTaskJob
	CreateTaskJob
)
```

api/controllers/service_test.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+func TestDeleteService(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockService := mock_provider.NewMockServiceProvider(ctrl)
+	controller := NewServiceController(mockService)
+
+	mockService.EXPECT().
```
sesh-kebab: Same as above - should be done via a job request I would've thought.

api/controllers/task.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return fireball.NewJSONResponse(202, model)
+}
+
+func (t *TaskController) DeleteTask(c *fireball.Context) (fireball.Response, error) {
+	id := c.PathVariables["id"]
+	if err := t.TaskProvider.Delete(id); err != nil {
```
sesh-kebab: Deletion of a Task should also happen via a job request?

zpatrick: I think we're still a unclear on how tasks will look. This may look radically different later on

zpatrick: But yeah - more likely than not it'll be in a job

diemonster: Lets revisit this when we figure out how `task` will work in this refactor?

PR #328, opened 2017-08-17
LB Provider

zpatrick:


api/provider/aws/load_balancer_create.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+
+	if _, err := l.AWS.IAM.PutRolePolicy(input); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (l *LoadBalancerProvider) authorizeSGIngressFromPort(groupID string, port models.Port) error {
```
zpatrick: this only uses the `port.HostPort`, so this should just take an `int` param

api/provider/aws/load_balancer_create.go
```diff
@@ -0,0 +1,269 @@
...
+) error {
+	listeners := make([]*elb.Listener, len(ports))
+	for i, port := range ports {
+		listener := &elb.Listener{}
+		listener.SetProtocol(port.Protocol)
+		listener.SetLoadBalancerPort(port.HostPort)
+		listener.SetInstancePort(port.ContainerPort)
+
+		if port.CertificateName != "" {
+			certificateARN, err := l.lookupCertificateARN(port.CertificateName)
```
zpatrick: Should I convert `models.Port` to `elb.Listener` in a separate function?

api/provider/aws/load_balancer_delete.go
```diff
@@ -0,0 +1,102 @@
...
+		return err
+	}
+
+	for _, tag := range tags {
+		if err := l.TagStore.Delete(tag.EntityType, tag.EntityID, tag.Key); err != nil {
+			return err
+		}
+	}
+
+	serviceTags, err := l.TagStore.SelectByType("service")
```
zpatrick: does deleting a service's tag make sense here? Or should that only be done on delete service?

api/provider/aws/load_balancer_read.go
```diff
@@ -0,0 +1,115 @@
...
+
+	allServiceTags, err := l.TagStore.SelectByType("service")
+	if err != nil {
+		return err
+	}
+
+	if tag, ok := allServiceTags.WithKey("load_balancer_id").WithValue(loadBalancerID).First(); ok {
+		model.ServiceID = tag.EntityID
+
+		serviceTags, err := l.TagStore.SelectByTypeAndID("service", tag.EntityID)
```
zpatrick: I don't need to do another `Select` here since I already have all of the service tags.
This should be 
```
serviceTags := allServiceTags.WithID(tag.EntityID)
...
`

File: api/provider/aws/load_balancer_delete.go (hunk 3 above)
Comment (zpatrick) [inline api/provider/aws/load_balancer_delete.go]: Since this tag "belongs" to the service, I think it makes more sense to delete it with the service.

File: api/provider/aws/load_balancer_delete.go [hunk 5]
```diff
@@ -0,0 +1,91 @@
...
+)
+
+func (l *LoadBalancerProvider) Delete(loadBalancerID string) error {
+	fqLoadBalancerID := addLayer0Prefix(l.Config.Instance(), loadBalancerID)
+
+	if err := l.deleteLoadBalancer(fqLoadBalancerID); err != nil {
+		return err
+	}
+
+	roleName := fmt.Sprintf("%s-lb", fqLoadBalancerID)
```
sesh-kebab: nit: I think when constructing a role (or sg), even though it is simple, would be good to separate into a method.

api/provider/aws/load_balancer_delete.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/elb"
+	"github.com/aws/aws-sdk-go/service/iam"
+)
+
+func (l *LoadBalancerProvider) Delete(loadBalancerID string) error {
+	fqLoadBalancerID := addLayer0Prefix(l.Config.Instance(), loadBalancerID)
+
+	if err := l.deleteLoadBalancer(fqLoadBalancerID); err != nil {
```
sesh-kebab: question: when deleting resources like r1, r2 & r3, we only check if an r actually exists or not for r3 (for example). Is there an implicit logic being applied here?

api/provider/aws/load_balancer_read.go
```diff
@@ -0,0 +1,110 @@
...
+	input := &elb.DescribeLoadBalancersInput{}
+	input.SetLoadBalancerNames([]*string{aws.String(loadBalancerName)})
+	input.SetPageSize(1)
+
+	output, err := l.AWS.ELB.DescribeLoadBalancers(input)
+	if err != nil {
+		return nil, err
+	}
+
+	return output.LoadBalancerDescriptions[0], nil
```
sesh-kebab: Usually this would seem really odd - accessing an array without checking the array's length but i'm guessing if the call to `DescribeLoadBalancers` doesn't return an error, the output will *definitely* contain a single loadbalancer element?

api/provider/aws/load_balancer_read.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func (l *LoadBalancerProvider) readTags(loadBalancerID string, model *models.LoadBalancer) error {
```
sesh-kebab: nit: maybe rename to `populateWithTagInformation` or something that is more representative of the what the method is doing? It wasn't immediately clear that we're modifying the object being passed in as a parameter.

api/provider/aws/load_balancer_create.go
```diff
@@ -0,0 +1,278 @@
...
+	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/quintilesims/layer0/common/models"
+)
+
+func (l *LoadBalancerProvider) Create(req models.CreateLoadBalancerRequest) (*models.LoadBalancer, error) {
+	loadBalancerID := generateEntityID(req.LoadBalancerName)
+	fqLoadBalancerID := addLayer0Prefix(l.Config.Instance(), loadBalancerID)
+	fqEnvironmentID := addLayer0Prefix(l.Config.Instance(), req.EnvironmentID)
+
+	environmentSGName := fmt.Sprintf("%s-env", fqEnvironmentID)
```
zpatrick: Yeah I'm a little torn on this. Reason being it takes a step closer to our [current ID implementation](https://github.com/quintilesims/layer0/blob/develop/api/backend/ecs/id/id.go#L107). I don't really feel too strongly either way - only that if we use a common helper function we are very vigilant about keeping it simple/sane.

api/provider/aws/load_balancer_delete.go
zpatrick: It depends on the AWS resource. For example, security groups don't have a logical name-id mapping, so we have to do a lookup to find the security group's ID before we can delete it. If it doesn't need that step, then we just go straight to deletion.

api/provider/aws/load_balancer_read.go
zpatrick: That's the promise we've been banking on so far - but I can't think of a good reason for us not to add the extra couple lines of code to verify the length.

api/provider/aws/load_balancer_list.go
```diff
@@ -0,0 +1,76 @@
...
+
+	if err := e.populateSummariesTags(summaries); err != nil {
+		return nil, err
+	}
+
+	return summaries, nil
+}
+
+func (e *LoadBalancerProvider) listLoadBalancerNames() ([]string, error) {
+	output, err := e.AWS.ELB.DescribeLoadBalancers(&elb.DescribeLoadBalancersInput{})
```
sesh-kebab: Load balancers can return multiple pages of results.
http://docs.aws.amazon.com/sdk-for-go/api/service/elb/#DescribeLoadBalancersInput

Even though the default is 400 for the page size, I think either a comment here stating we aren't checking for multiple pages because the default is sufficiently high enough (if that is the case) or ensure we iterate through all the pages (possibly better option).

PR #329, opened 2017-08-18
232: LoadBalancers as new benchmark parameter, more baseline benchmark tests

jparsons04:
Merging in work @tlake and I have combined to do into our shared feature branch.
Implemented Load Balancers as a parameterized value that can be provided for a benchmark test.
Added several baseline benchmark tests

The benchmarks should run by running `make test` from the tests/stress directory. The README in the tests/ directory provide some notes for consideration when preparing to run the stress tests.

**Checklist**
~- [ ] Unit tests~


tests/stress/cases/modules/loadbalancer/main.tf
```diff
@@ -0,0 +1,12 @@
+resource "layer0_load_balancer" "base" {
```
zpatrick: I think with some fancy interpolation, we can reduce everything into `cases/modules` into a single, flexible terraform configuration. 


A simple approach is to change the input variables for environment-specific entities from flat numbers to per-environment numbers:
```
variable "num_environments" {
    type = "int"
    default = 1
}

variable "services_per_environment" {
  type = "int"
  default = 1
}

resource "layer0_environment" "base" {
    ...
    count = "${var.num_environments}"
}

resource "layer0_service" "base" {
    ...
    environment_id = "${layer0_environment.base[count.index % var.num_environments].id}"
    count          = "${var.num_environments * var.services_per_environment}"
}
```

The name of the tests could potentially add some more detail:
```
func Benchmark_25ServicesAcross5Environments(b *testing.B) {
	benchmarkStress(b, 5, 5, 0, 0)
}
```

_edit - I think a struct could make this even more explicit_
```
func Benchmark_25ServicesAcross5Environments(b *testing.B) {
    benchmarkStress(b, BenchmarkTestCase{
        Environments:           5,
        ServicesPerEnvironment: 5,
    })
}
```

zpatrick: Also, a more complex approach would be to accept maps as input parameters. That way, we could test something like `there are 3 environments: environment 1 has 5 services, environment 2 has 1 service, and environment 3 has 2 services`.  

There's probably a more sophisticated way to do this, but if you had:
```
variable "num_environments" {
    type = "int"
    default = 1
}

variable "serviceEnvironmentMap" {
  type = "map"
  default = {}
}

resource "layer0_environment" "base" {
    ...
    count = "${var.num_environments}"
}

resource "layer0_service" "base" {
    ...
    environment_id = "${layer0_environment.base[lookup(var.serviceEnvironmentMap, count.index)].id}`
    count          = "${length(serviceEnvironmentMap)}"
}
```

Then you could accomplish `there are 3 environments: environment 1 has 5 services, environment 2 has 1 service, and environment 3 has 2 services` with: 
```
// assuming we would translate the map[int]int to a string later
"serviceEnvironmentMap" = map[int]int{
    0: 0,
    1: 0,
    2: 0,
    3: 0,
    4: 0,
    5: 1,
    6: 2,
    7: 2,
}
```

---
**Note**
my original approach was to have a map of environment indices to an integer representing the number of services for that environment, e.g.:
```
variable "environment_services_counts" {
  default = {
    0: 5,
    1: 1,
    2: 2,
}
``` 

I found 2 problems with this: 

* To calculate the `count` for the `layer0_service` resource, you need to sum up all of the values in the map. There is a `values` helper function, but no `sum` function for some reason. But, as a workaround, you could just pass in a `num_services` variable. 
* How do you calculate the index of the environment the service belongs to given `count.index`? In the best case it would require some pretty complicated interpolation (or at least I think it would - maybe it's easier than I'm making it?). In the worst case, it's not even possible given the current available [functions](https://www.terraform.io/docs/configuration/interpolation.html).

jparsons04:
Duly noted on the commentary from @zpatrick. I agree that they do deserve further contemplation.

PR #332, opened 2017-08-24
Update to align with changes in `tftest` package

tlake:
The [tftest](https://github.com/quintilesims/tftest) package was updated so that various `testing` objects could be used (for example, `testing.B` in addition to `testing.T`).

- Stripped out an unnecessary declaration of the `Tester` interface in favor of interfaces defined in the `tftest` package.
- Used `testing.B` as both `Tester` and `Logger`

`make test`

**Checklist**
-  ~Unit tests~
-  ~Smoke tests (if applicable)~
-  ~System tests (if applicable)~
-  ~Documentation (if applicable)~

closes #


tests/stress/stress_test.go
```diff
@@ -30,7 +23,7 @@ func NewStressTest(t Tester, dir string, vars map[string]string) *StressTest {
 		tftest.Dir(dir),
 		tftest.Vars(vars),
 		tftest.DryRun(*dry),
-		tftest.Log(log))
+		tftest.Log(t.(tftest.Logger)))
```
zpatrick: this will cause a panic

tests/stress/stress_test.go
```diff
@@ -6,19 +6,12 @@ import (
...
-	Fatalf(format string, args ...interface{})
-}
-
 type StressTest struct {
 	Terraform *tftest.TestContext
 	Layer0    *clients.Layer0TestClient
 }
 
-func NewStressTest(t Tester, dir string, vars map[string]string) *StressTest {
+func NewStressTest(t tftest.Tester, dir string, vars map[string]string) *StressTest {
```
zpatrick: This should probably be a `*testing.B` object.

tests/stress/stress_test.go
tlake: Will it? It seemed to be happy when I ran the stress tests. Why would this cause a panic?

tlake: Only if `t` can't implement both `Tester` and `Logger` though, right? I didn't get any complaints when I ran a test locally, but that's probably because `testing.B` can be both a `Tester` and a `Logger`.

I understand, though, that the func sig doesn't actually enforce that behavior, so in the event that a user passes in a `Tester` that doesn't also implement `Logger`, Go would fall over. 👍 ✅

PR #333, opened 2017-08-28
232 stress tests

### LINKED ISSUE #232: beef up task stress test
* We should increase the total number of tasks created from 100 to 500. Possibly 1000. 
* Tasks should task some random amount of time to complete. Probably between 30 seconds to 5 minutes.
* We should make the create task requests async, to simulate multiple sources hitting the api.
* Instead of creating all the tasks right at the start, we should create tasks in waves. This would simulate a more realistic scenario, and would ensure the environment scalar runs properly over a long period of stress.

tlake:
Introduces an update to the stress tests which features multi-dimensional deployment tests and  benchmarking of CLI commands under these various deployment configurations.

(This work also spawned some changes in the [tftest](https://github.com/quintilesims/tftest) package.)

Run `make test` in the `layer0/tests/stress/` directory (check out `layer0/tests/README.md` beforehand to make sure that you won't break things by spinning up too many resources).

**Checklist**
- ~Unit tests~

closes  _(eventually - **TODO: that LIMITS file**)_ #232


tests/stress/Makefile
```diff
@@ -1,4 +1,4 @@
 test: 
-	go test -v -parallel 10 -timeout 1h
+	go test -v -run NOTHING -bench . -timeout 2h
```
zpatrick: does this need `-run NOTHING` ?

tests/stress/benchmarkstress_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"strconv"
+	"testing"
+)
+
+const (
+	deployCommand = "while true ; do echo LONG RUNNING SERVICE ; sleep 5 ; done"
+)
+
+func BenchmarkStress25Environments0Deploys0Services0LoadBalancers(b *testing.B) {
+	benchmarkStress(25, 0, 0, 0, deployCommand, b)
```
zpatrick: Since `deployCommand` doesn't change, I think it makes sense to just hardcode it into the terraform configuration.

tests/stress/benchmarkstress_test.go
```diff
@@ -0,0 +1,90 @@
...
+}
+func BenchmarkStress10Environments10Deploys0Services0LoadBalancers(b *testing.B) {
+	benchmarkStress(10, 10, 0, 0, deployCommand, b)
```
zpatrick: [nit] helper functions for tests that take a `*testing.T` or `*testing.B` tend to be the first parameter.

tests/stress/benchmarkstress_test.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"strconv"
+	"testing"
+)
+
+const (
+	deployCommand = "while true ; do echo LONG RUNNING SERVICE ; sleep 5 ; done"
+)
+
+func BenchmarkStress25Environments0Deploys0Services0LoadBalancers(b *testing.B) {
```
zpatrick: Adding the zero-value entities to the test name seems redundant (also the word "Stress"). 

The following names seems to read better IMO:
```
Benchmark25Environments(b *testing.B) { ... }
Benchmark20Environments10Services(b *testing.B) { ... }
```

tests/stress/benchmarkstress_test.go
```diff
@@ -0,0 +1,90 @@
...
+func BenchmarkStress10Environments10Deploys0Services0LoadBalancers(b *testing.B) {
+	benchmarkStress(10, 10, 0, 0, deployCommand, b)
+}
+func BenchmarkStress20Environments20Deploys0Services0LoadBalancers(b *testing.B) {
+	benchmarkStress(20, 20, 0, 0, deployCommand, b)
+}
+func BenchmarkStress5Environments50Deploys0Services0LoadBalancers(b *testing.B) {
+	benchmarkStress(5, 50, 0, 0, deployCommand, b)
+}
+func BenchmarkStress5Environments100Deploys0Services0LoadBalancers(b *testing.B) {
```
zpatrick: Deploys don't provide very meaningful stats IMO. I think we'd like to know some sort of max number we could support (e.g. 10k), but experience with Layer0 shows that this is rarely limiting factor compared to services, tasks, and environments.

tests/stress/benchmarkstress_test.go
```diff
@@ -0,0 +1,90 @@
...
+func benchmarkStress(envs, deps, servs, lbs int, deploycomm string, b *testing.B) {
+	tfvars := map[string]string{
+		"num_environments":  strconv.Itoa(envs),
+		"num_deploys":       strconv.Itoa(deps),
+		"num_services":      strconv.Itoa(servs),
+		"num_loadbalancers": strconv.Itoa(lbs),
+		"deploy_command":    deploycomm,
+	}
+
+	log.Debugf("Testing with Environments: %v, Deploys: %v, Services: %v, Load Balancers: %v, Command: %v", envs, deps, servs, lbs, deploycomm)
```
zpatrick: [nit] lines a little long, could probably get away with `log.Debugf("Testing with params: %v", tfvars)`

tests/stress/benchmarkstress_test.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+	})
+
+	b.Run("ListTasks", func(b *testing.B) {
+		for n := 0; n < b.N; n++ {
+			s.Layer0.ListTasks()
+		}
+	})
+
+	b.Run("ListJobs", func(b *testing.B) {
```
zpatrick: Given that we know `ListJobs` is just doing a "SelectAll" to the jobs database under the covers, and the number of jobs don't change for any of these tests, I don't think this gives us very meaningful stats.

tests/stress/cases/modules/deploy/main.tf
```diff
@@ -0,0 +1,13 @@
+resource "layer0_deploy" "base" {
```
zpatrick: As stated in the previous pull request, I think we should attempt to use only a single Terraform configuration for all stress tests, with "entity-per-environment" input parameters.

tests/stress/benchmarkstress_test.go
diemonster: Seems like there's a few different approaches you could take here that would make these function names not have to be oddly named. However, I'm guessing you did it this way because then each benchmark gets it's own context/duration in the resulting output? 

Open question then becomes: what's more important? Readable output of our benchmarks or readable code? I think I personally prefer the latter.

tests/stress/cases/modules/main.tf
```diff
@@ -0,0 +1,29 @@
+provider "layer0" {
```
zpatrick: This Terraform configuration is more-or-less doing the single configuration approach I had mentioned earlier. The difference being the use of sub-modules here. With some interpolation, we can get rid of the sub-modules and combine everything we need into a single configuration.

tests/stress/stress_test.go
```diff
@@ -13,19 +13,21 @@ type StressTest struct {
 	Layer0    *clients.Layer0TestClient
 }
 
-func NewStressTest(t *testing.T, dir string, vars map[string]string) *StressTest {
+func NewStressTest(t *testing.B, dir string, vars map[string]string) *StressTest {
```
zpatrick: This param should be `b`, not `t`

tests/stress/benchmarkstress_test.go
```diff
@@ -0,0 +1,90 @@
+package system
+
+import (
+	"strconv"
+	"testing"
+)
+
+const (
+	deployCommand = "while true ; do echo LONG RUNNING SERVICE ; sleep 5 ; done"
```
diemonster: Can you explain how this works and why it's needed?

tests/stress/stress_test.go
```diff
@@ -13,19 +13,21 @@ type StressTest struct {
...
 	vars["token"] = config.AuthToken()
 
-	tfContext := tftest.NewTestContext(t,
+	tfContext := tftest.NewTestContext(
+		t,
 		tftest.Dir(dir),
 		tftest.Vars(vars),
 		tftest.DryRun(*dry),
-		tftest.Log(log))
+		tftest.Log(t),
```
zpatrick: This shouldn't change - should still be `tftest.Log(log)` since that's how the `-debug` flag gets used

tests/stress/stress_test.go
```diff
@@ -13,19 +13,21 @@ type StressTest struct {
 	Layer0    *clients.Layer0TestClient
```
zpatrick: Now that I look at it, the `type StressTest struct` seems heavy-handed to me (same problem exists in system tests). 
I think it would be easier to just have `func NewStressTest` return the `APIClient` and `Terraform` objects.

```
terraform, apiClient := NewStressTest(b, "cases/modules", tfvars)
terraform.Apply()
defer terraform.Destroy()

b.Run("ListEnvironments", func(b *testing.B) {
    for n := 0; n < b.N; n++ {
        apiClient.ListEnvironments()
    }
})
```

tests/stress/benchmarkstress_test.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+	})
+
+	b.Run("ListServices", func(b *testing.B) {
+		for n := 0; n < b.N; n++ {
+			s.Layer0.ListServices()
+		}
+	})
+
+	b.Run("ListTasks", func(b *testing.B) {
```
zpatrick: We need to capture `GetEntity` functions since these use different aws api calls than `List`, and are expected to be used more frequently than `ListEntity` functions.

tests/stress/taskStress_test.go
```diff
@@ -1,73 +0,0 @@
...
-	"testing"
-	"time"
-
-	"github.com/quintilesims/layer0/common/testutils"
-)
-
-// Test Resources:
-// This test creates an environment named 'tp'
-// and a deploy named 'alpine'
-func TestTaskStress(t *testing.T) {
```
zpatrick: I think this test should be cleaned up a bit, but definitely not deleted, as it's the only test that creates tasks.

diemonster:
Another thing I'd like to see, and I think we could spin off a new issue for this if need be, is a new page on our official docs (layer0.ims.io) that talks about our service limits. We should communicate our findings to our users in some fashion, and that seems like the best place to do it.

tests/stress/benchmarkstress_test.go
diemonster: ^^^ disregard my question as long as you remove `deployCommand` as a param for `benchmarkStress`, in which case this makes a lot more sense

tests/stress/benchmarkstress_test.go
zpatrick: The 1-function-per-case for benchmark testing is a standard Go thing (see: https://dave.cheney.net/2013/06/30/how-to-write-benchmarks-in-go). So I think the pattern is correct, just the names of function themselves could be improved.

tests/stress/taskStress_test.go
tlake: Cool - I had a comment on the `232-tlake` -> `232-stress-tests` PR earlier about this; in `stress_test.go:NewStressTest()`, you had advocated for replacing the `t Tester` param with a straight-up `b *testing.B` param, which broke compatibility with these non-benchmarking tasks.

Offhand, this test doesn't make use of benchmarking, and so probably shouldn't be using `*testing.B`. Should `NewStressTest()` be modified to accept either `*testing.T` or `*testing.B`, or should `TestTaskStress()` call a new `NewXTest()` function?

tests/stress/benchmarkstress_test.go
tlake: ✅

tests/stress/benchmarkstress_test.go
tlake: Will adjust names as suggested below.

✅

tests/stress/Makefile
tlake: ✅

tests/stress/benchmarkstress_test.go
tlake: ✅

tests/stress/benchmarkstress_test.go
tlake: ✅

tests/stress/cases/modules/main.tf
zpatrick: Interesting idea may be to randomly shuffle entities across environments using the [random shuffle](https://www.terraform.io/docs/providers/random/r/shuffle.html) provider

tests/stress/benchmarkstress_test.go
tlake: ✅

tests/stress/benchmarkstress_test.go
tlake: ✅

tests/stress/taskStress_test.go
zpatrick: I think there's a case to be made for making tasks a param for `benchmarkStress`

tests/stress/stress_test.go
tlake: ✅

tests/stress/stress_test.go
tlake: ✅

tests/stress/stress_test.go
tlake: ✅

tests/stress/benchmarkstress_test.go
tlake: Will remove.

✅

tests/stress/cases/modules/main.tf
diemonster: `random_shuffle` would be cool if these tests were part of a CI pipeline that was executed on a frequent basis. Not sure if that was our plan overall for these tests, though.

tlake: We shouldn't be automating these benchmarking/stress tests in a pipeline, I don't think, since they're expensive and lengthy, and don't really add testing coverage that other unit/smoke/system tests do. I think `random_shuffle` is just a way to distribute entities across other entities.

diemonster: 💯 🍪

tests/stress/benchmarkstress_test.go
tlake: Update: Since we should recreate zpatrick's old Task stress test with this newer sort of pattern, I'm going to declare two `const`s for deploy commands: a while loop to simulate a long-running service, and a simple sleep to simulate a one-off task. There will be a `DeployCommand` field in the `StressTestCase` struct which will reference one of these two `const`s.

tests/stress/benchmarkstress_test.go
tlake: ✅

tests/stress/cases/modules/deploy/main.tf
tlake: Use of `random_pet` to generate names and proper leveraging of `element()` has resulted in a single-file terraform configuration! (Excluding `variables.tf` and `outputs.tf`, that is.)

✅

tests/stress/cases/modules/main.tf
tlake: Update: A single-file Terraform configuration has been achieved; `random_shuffle`, while cool and fun and interesting, doesn't provide any added value here. I've refactored to use `element()` again, which has modulus built-in.

✅

tests/stress/stress_test.go
```diff
@@ -15,9 +16,9 @@ const (
 )
 
 type StressTestCase struct {
```
zpatrick: Overall, these changes are looking really good :+1: 

I think we're overthinking the Deploy for Task vs. Service. 
There's really two parts I think we need to tackle: the dependency and the command. 


Since we already have an environment dependency for all of our entities (e.g. you could never run with `NumServices > 0 && NumEnvironments == 0`), I think it's reasonable and consistent to require that `NumDeploys > 0` for any tests with Tasks and/or Services.

As for the command in the Deploy, I don't think it actually matters. We don't need tasks to exit, I think it's not unreasonable that we just hardcode the command into `Dockerrun.aws.json`. This will simplify the test code and the terraform config a bit (since we won't need to template the deploy).

tlake:
@diemonster Created issue #337 to address presentation of our findings to users.

tests/stress/taskStress_test.go
tlake: Refactored task stress test into benchmark pattern.

tests/stress/benchmarkstress_test.go
tlake: Removed lesser Deploy test.

✅

tests/stress/stress_test.go
tlake: Added a couple of top-of-test checks for Environments and Deploys given Services and/or Tasks. Also reverted hardcoded deploy command.

✅

tests/stress/stress_test.go
```diff
@@ -42,32 +46,65 @@ func runTest(b *testing.B, c StressTestCase) {
...
+	}
+
+	if c.NumTasks > 0 {
+		methodsToBenchmark["ListTasks"] = func() { layer0.ListTasks() }
+
+		deployIDs := strings.Split(terraform.Output("deploy_ids"), ",\n")
+		environmentIDs := strings.Split(terraform.Output("environment_ids"), ",\n")
+
+		tasksCreated := 0
+		for copies := c.NumTasks / 2; tasksCreated < c.NumTasks; copies = copies / 2 {
```
zpatrick: This `copies` logic is a little odd to me, especially the `copies = copies / 2` . 
I think there's a more simple/clear way to do what you want here. 

However, I don't think we need to mess with copies at all in this test. 
Reason being, unlike `Deploys` where the distribution among the family makes a difference in underlying architecture/api calls, 
the number of copies for a task is more-or-less an alias for running `TaskCreate()` N times.

tests/stress/stress_test.go
```diff
@@ -42,32 +46,65 @@ func runTest(b *testing.B, c StressTestCase) {
...
+			taskName := fmt.Sprintf("Task%v", copies)
+			go func(taskName string, copies int) {
```
zpatrick: The args into the `go` function aren't needed here. 
As far as I've seen, the only scenario where we need them is when we run `go` functions inside of a `range` loop that iterates over pointers (since the iterator re-uses the same pointer each time).

tests/stress/stress_test.go
```diff
@@ -42,32 +46,65 @@ func runTest(b *testing.B, c StressTestCase) {
...
+				layer0.CreateTask(taskName, environmentIDs[0], deployIDs[0], copies, nil)
+			}(taskName, copies)
+
+			tasksCreated += copies
+			if copies <= 1 {
+				copies++
+			}
+		}
+
+		testutils.WaitFor(b, time.Second*30, time.Minute*10, func() bool {
```
zpatrick: I don't think we need to wait for the tasks to finish in this test. To me, that is behavior that should be captured in a system test, not a benchmark test.

tests/stress/stress_test.go
```diff
@@ -111,13 +148,6 @@ func Benchmark20Environments20Deploys(b *testing.B) {
...
 }
 
-func Benchmark5Environments50Deploys(b *testing.B) {
-	runTest(b, StressTestCase{
-		NumDeploys:      50,
-		NumEnvironments: 5,
-	})
-}
-
 func Benchmark5Environments100Deploys(b *testing.B) {
```
zpatrick: `Benchmark5Environments100Deploys` is an odd metric to test. Deploys are totally independent of environments, why wouldn't just test deploys by themselves? 

I think the **only** deploy-related benchmark tests should be:
`BenchmarkNDeploysAcrossXFamilies`

tests/stress/stress_test.go
```diff
@@ -158,3 +188,11 @@ func Benchmark25Environments25Deploys25Services25LoadBalancers(b *testing.B) {
 		NumServices:      25,
```
zpatrick: By themselves, these metrics don't tell me much. If/when the API starts crapping out, its hard to tell what the cause may be. I think we need to have more logical groupings of tests for each entity, probably a small/med/large size for that entity type with all other variables as consistent as possible. 

```
Benchmark1Family10Deploys(b *testing.B)
Benchmark1Family50Deploys(b *testing.B)
Benchmark1Family100Deploys(b *testing.B)

Benchmark10Families10Deploys(b *testing.B)
Benchmark50Families50Deploys(b *testing.B)
Benchmark100Families100Deploys(b *testing.B)

Benchmark5Environment(b *testing.B)
Benchmark10Environments(b *testing.B)
Benchmark25Environments(b *testing.B)

Benchmark1Environment5Services(b *testing.B)
Benchmark1Environment25Services(b *testing.B)
Benchmark1Environment50Services(b *testing.B)

Benchmark10Environments25Services(b *testing.B)
Benchmark25Environments50Services(b *testing.B)

Benchmark1Environment5LoadBalancers(b *testing.B)
Benchmark1Environment25LoadBalancers(b *testing.B)
Benchmark1Environment50LoadBalancers(b *testing.B)

Benchmark10Environments25LoadBalancers(b *testing.B)
Benchmark25Environments50LoadBalancers(b *testing.B)

Benchmark1Environment10Tasks(b *testing.B)
Benchmark1Environment50Tasks(b *testing.B)
Benchmark1Environment100Tasks(b *testing.B)

Benchmark10Environments50Tasks(b *testing.B)
Benchmark25Environments100Tasks(b *testing.B)

// with each individual entity tested, now we can test the aggregate scenarios
Benchmark1Environment5Services5LoadBalancers10Tasks(b *testing.B)
Benchmark10Environments25Services25LoadBalancers50Tasks(b *testing.B)
Benchmark25Environments50Services50LoadBalancers100Tasks(b *testing.B)
```

zpatrick: With this in mind, I think it could be a decent idea to keep `const` values for the `min`, `max`, and `average` scale for each entity.

```
const (
    ENVIRONMENT_SCALE_MIN = 1
    ENVIRONMENT_SCALE_AVG = 10
    ENVIRONMENT_SCALE_MAX = 25
    SERVICE_SCALE_MIN     = 5
    SERVICE_SCALE_AVG     = 25
    SERVICE_SCALE_MAX     = 50
)

func BenchmarkMinEnvironmentsMinServices(b *testing.B)
func BenchmarkMinEnvironmentsAvgServices(b *testing.B)
func BenchmarkMinEnvironmentsMaxServices(b *testing.B)

func BenchmarkAvgEnvironmentsAvgServices(b *testing.B)
func BenchmarkMaxEnvironmentsMaxServices(b *testing.B)
```

zpatrick:
By themselves, these metrics don't tell me much. If/when the API starts crapping out, its hard to tell what the cause may be. I think we need to have more logical groupings of tests for each entity, probably a small/med/large size for that entity type with all other variables as consistent as possible. 

```
Benchmark1Family10Deploys(b *testing.B)
Benchmark1Family50Deploys(b *testing.B)
Benchmark1Family100Deploys(b *testing.B)

Benchmark10Families10Deploys(b *testing.B)
Benchmark50Families50Deploys(b *testing.B)
Benchmark100Families100Deploys(b *testing.B)

Benchmark5Environment(b *testing.B)
Benchmark10Environments(b *testing.B)
Benchmark25Environments(b *testing.B)

Benchmark1Environment5Services(b *testing.B)
Benchmark1Environment25Services(b *testing.B)
Benchmark1Environment50Services(b *testing.B)

Benchmark10Environments25Services(b *testing.B)
Benchmark25Environments50Services(b *testing.B)

Benchmark1Environment5LoadBalancers(b *testing.B)
Benchmark1Environment25LoadBalancers(b *testing.B)
Benchmark1Environment50LoadBalancers(b *testing.B)

Benchmark10Environments25LoadBalancers(b *testing.B)
Benchmark25Environments50LoadBalancers(b *testing.B)

Benchmark1Environment10Tasks(b *testing.B)
Benchmark1Environment50Tasks(b *testing.B)
Benchmark1Environment100Tasks(b *testing.B)

Benchmark10Environments50Tasks(b *testing.B)
Benchmark25Environments100Tasks(b *testing.B)

// with each individual entity tested, now we can test the aggregate scenarios
Benchmark1Environment5Services5LoadBalancers10Tasks(b *testing.B)
Benchmark10Environments25Services25LoadBalancers50Tasks(b *testing.B)
Benchmark25Environments50Services50LoadBalancers100Tasks(b *testing.B)
```

With this in mind, I think it could be a decent idea to keep `const` values for the `min`, `max`, and `average` scale for each entity.

```
const (
    ENVIRONMENT_SCALE_MIN = 1
    ENVIRONMENT_SCALE_AVG = 10
    ENVIRONMENT_SCALE_MAX = 25
    SERVICE_SCALE_MIN     = 5
    SERVICE_SCALE_AVG     = 25
    SERVICE_SCALE_MAX     = 50
)

func BenchmarkMinEnvironmentsMinServices(b *testing.B)
func BenchmarkMinEnvironmentsAvgServices(b *testing.B)
func BenchmarkMinEnvironmentsMaxServices(b *testing.B)

func BenchmarkAvgEnvironmentsAvgServices(b *testing.B)
func BenchmarkMaxEnvironmentsMaxServices(b *testing.B)
```

tests/stress/stress_test.go
tlake: The `copies` logic is a way to sort-of recreate the different batches of Tasks that you had in that map in the old test (i.e. simulating multiple users submitting Tasks), but without hardcoding batch sizes or knowing how many total Tasks are to be created:
```
taskNameCopies := map[string]int{
		"TaskA": 50,
		"TaskB": 25,
		"TaskC": 10,
		"TaskD": 5,
		"TaskE": 5,
		"TaskF": 1,
		"TaskG": 1,
		"TaskH": 1,
		"TaskI": 1,
		"TaskJ": 1,
}
```

For example, given 100 tests, this loop would make batch submissions of 50, 25, 12, 6, 3, 1, 1, 1, 1.

tests/stress/stress_test.go
tlake: ✅

tests/stress/stress_test.go
tlake: Ah, so the `go func` will reach out into the state of the `for` loop.

✅

tests/stress/stress_test.go
zpatrick: I think it's quite a bit of complexity for little-to-no benefit, since, under the covers, it's essentially doing `RunTask()` N times. 

Also fwiw,  I'm going to push for removing `copies` in the api refactor

tlake: Cool, I'll gut it here.

✅

tests/stress/stress_test.go
```diff
@@ -85,20 +84,7 @@ func runTest(b *testing.B, c StressTestCase) {
...
-				log.Debugf("Creating task %v", taskName)
-				layer0.CreateTask(taskName, environmentIDs[0], deployIDs[0], copies, nil)
-			}()
-
-			tasksCreated += copies
-			if copies <= 1 {
-				copies++
-			}
-		}
+		layer0.CreateTask("tt", environmentIDs[0], deployIDs[0], c.NumTasks, nil)
```
zpatrick: I think we would want to copy the even spread across environments, e.g. 
```
for i := range c.NumTasks {
    environmentID := environmentIDs[len(environmentIDs) % i]
    layer0.CreateTask(...)
}
```

tlake: Oh, good catch - I meant for that to be in there.

✅

tests/stress/stress_test.go
```diff
@@ -1,39 +1,298 @@
...
+	})
+}
+
+func BenchmarkMaxEnvironments(b *testing.B) {
+	runTest(b, StressTestCase{
+		NumEnvironments: ENVIRONMENT_SCALE_MAX,
+	})
+}
+
+func BenchmarkMinServices(b *testing.B) {
```
zpatrick: don't these tests need 1 environment and 1 deploy?

tests/stress/stress_test.go
```diff
@@ -1,39 +1,298 @@
...
+	for name, fn := range methods {
+		b.Run(name, func(b *testing.B) {
+			for n := 0; n < b.N; n++ {
+				fn()
+			}
+		})
+	}
+}
+
+func BenchmarkMinFamiliesMinDeploys(b *testing.B) {
```
zpatrick: This may just be personal preference, but I don't think deploys are "interesting" enough to warrant this amount of test cases. Maybe we just have 3 tests?
```
MinFamiliesMinDeploys
MedFamiliesMedDeploys
MaxFamiliesMaxDeploys
```

tests/stress/stress_test.go
tlake: They do, but I've built fixes for missing dependencies into the start of `runTest()`. I don't know if we like the pattern, but it trades explicitness in the Benchmark test construction for a tighter focus on what the test is intended to be testing.

tests/stress/stress_test.go
tlake: Sounds fine to me. 

✅

tests/stress/stress_test.go
zpatrick: :+1:

PR #336, opened 2017-08-31
Scheduler skeleton

zpatrick:
**TODO**
- [x] Update Controller Unit tests
- [x] Update import paths for `job_store` 
- [x] Remove remaining `runner` package references  (e.g. in `l0-setup`)
- [x] Skeleton runner in `provider/aws` package


diemonster:
@zpatrick As I review this, let's test if my assumptions are correct on the behavior:

- A given entity's method may schedule a `job` via `scheduleJob`
- `scheduleJob` is a utility function that calls `scheduler.Schedule()`, which passes in ` job` object. 
- Once a job is scheduled, it can be `Unscheduled` and/or obtained via the Scheduler interfaces' `List()` and `Read()`

From here, things get a little more confusing for me. I don't see any corresponding `Schedule` method that implements the `Scheduler` interface other than in tests. I do, however, see the the `Dispatcher` class, which appears to a means of execution as called by the Scheduler.

diemonster:
A few more comments: 

- [nit] I wasn't a fan of `scheduleJob` at first, but as I sit with the code, I like it more and more. I wish there was a simpler pattern that didn't require a helper function like this, but I can't see how to do it without leaking implementation details or creating more annoying issues down the line.

- [not entirely related] Could `MemoryJobStore` read from JSON potentially or is that not worth mucking around with?

- What's the difference between `pending` and `InProgress` states when a job is running via the API? I'm assuming `pending` is for when a job is sitting in the queue waiting for a lock to be released?

PR #341, opened 2017-09-11
Add worker queue and dispatcher to job package

zpatrick:
Adds a worker queue to the job package. This allows only N number of jobs to be ran at a given time. 

`go test`

**Checklist**
- [x] Unit tests


api/job/mock_job/mock_scheduler.go
```diff
@@ -1,73 +0,0 @@
-// Automatically generated by MockGen. DO NOT EDIT!
```
jparsons04: It looks like this file needs to be removed rather than cleared out. The current code state indicates that this file still exists, which causes an issue if i run `go test ./...`

```
$ go test ./...
can't load package: package github.com/quintilesims/layer0/api/job/mock_job: 
mock_job/mock_scheduler.go:1:1: expected 'package', found 'EOF'
```

jparsons04:
Absent mock_job/mock_scheduler.go needing to be removed, running `go test` passed without issue for me.

api/job/mock_job/mock_scheduler.go
zpatrick: This is removed on the branch, not sure why it's not reflecting that for you locally

zpatrick: fwiw I believe a cleared out file would have 1 line of green, this is a removed file

diemonster:
Had some more time to sit with this today, I think it's fine to merge as is. 👍

api/job/dispatcher.go
```diff
@@ -0,0 +1,62 @@
...
+	queue := make(chan models.Job)
+	for i := 0; i < numWorkers; i++ {
+		worker := NewWorker(i+1, store, queue, runner)
+		worker.Start()
+	}
+
+	dispatcher := NewDispatcher(store, queue)
+	ticker := time.NewTicker(DISPATCHER_PERIOD)
+	go func() {
+		for range ticker.C {
```
tlake: Oh woah, you can `range` a channel? That's super cool. So with something like a Ticker channel, you're just fetching the next item in a (controllably infinite) list, where items come out at specific intervals?

Sort of similar to something like
```
while True:
    sleep 5
    do_thing()
```
only you can control it with `Ticker.Stop()`, and it's smarter about keeping accurate time than a sleep would be.

api/job/worker.go
```diff
@@ -0,0 +1,60 @@
...
+				log.Printf("[INFO] [JobWorker %d]: Finished job %s", w.ID, job.JobID)
+                                w.Store.SetJobStatus(job.JobID, Completed)
+			case <-quit:
+				log.Printf("[DEBUG] [JobWorker %d]: Quit signalled", w.ID)
+				return
+			}
+		}
+	}()
+
+	return func() { quit <- true }
```
tlake: So each `Worker.Start()` precommits to a termination state; it spins off all its work and quit-checking into a goroutine and then sends a quit status into the channel. Once there's nothing more in `w.Queue`, it falls through to checking the `quit` channel, which will have been preloaded with `true`.

api/job/dispatcher.go
zpatrick: Correct, it isn't greedy since the underlying mechanism in a channel

PR #343, opened 2017-09-12
274: Add Update for Load Balancer Controller

jparsons04:
New model for UpdateLoadBalancerRequest, combines HealthCheck and *Ports[] into one struct
UpdateLoadBalancer is now wired up to the API and properly sends it through job scheduler

**Checklist**
- [x] Unit tests


api/controllers/load_balancer.go
```diff
@@ -78,3 +79,19 @@ func (l *LoadBalancerController) ListLoadBalancers(c *fireball.Context) (firebal
...
+
+func (l *LoadBalancerController) UpdateLoadBalancer(c *fireball.Context) (fireball.Response, error) {
+	id := c.PathVariables["id"]
+
+	var req models.UpdateLoadBalancerRequest
+	if err := json.NewDecoder(c.Request.Body).Decode(&req); err != nil {
+		return nil, errors.New(errors.InvalidRequest, err)
+	}
+
+	req.LoadBalancerID = id
```
zpatrick: This behavior is a little odd. Since we need the `LoadBalancerID` field in the `UpdateLoadBalancerRequest` model for it to be properly handled as a job, let's move  this request to `PUT /loadbalancer`

api/provider/load_balancer.go
```diff
@@ -7,4 +7,5 @@ type LoadBalancerProvider interface {
 	Read(loadBalancerID string) (*models.LoadBalancer, error)
 	List() ([]models.LoadBalancerSummary, error)
 	Delete(loadBalancerID string) error
+	Update(req models.UpdateLoadBalancerRequest) (*models.LoadBalancer, error)
```
zpatrick: Let's have updates return only an `error`. Given the job pattern, client's are going to call a `Read` after the job is completed anyways.

common/models/update_load_balancer_request.go
```diff
@@ -0,0 +1,38 @@
+package models
+
+import (
+	"fmt"
+
+	swagger "github.com/zpatrick/go-plugin-swagger"
+)
+
+type UpdateLoadBalancerRequest struct {
+	LoadBalancerID string
```
zpatrick: This needs the `json:""` tag

common/models/update_load_balancer_request.go
```diff
@@ -0,0 +1,38 @@
...
+	Ports          *[]Port     `json:"ports"`
+	HealthCheck    HealthCheck `json:"health_check"`
```
zpatrick: This should be a `*HealthCheck`

common/models/update_load_balancer_request.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+func (u UpdateLoadBalancerRequest) Validate() error {
+	if u.LoadBalancerID == "" {
+		return fmt.Errorf("LoadBalancerID is required")
+	}
+
```
zpatrick:
```
if u.Ports != nil {
    for _, port := range u.Ports {
        if err := port.Validate(); err != nil {
            return err
        }
    }
}

if u.HealthCheck != nil {
    if err := u.HealthCheck.Validate(); err != nil {
        return err
    }
}
```

(if those sub-`Validate()` functions don't exist yet, we should add them)

common/models/update_load_balancer_request.go
```diff
@@ -0,0 +1,38 @@
...
+	return nil
+}
+
+func (u UpdateLoadBalancerRequest) Definition() swagger.Definition {
+	return swagger.Definition{
+		Type: "object",
+		Properties: map[string]swagger.Property{
```
zpatrick: The swagger definition needs to match the fields in the `UpdateLoadBalancer` struct. To do nested objects, I believe you use `NewObjectProperty("name")`, e.g.:
```
"load_balancer_id":  swagger.NewStringProperty(),
"ports":             swagger.NewObjectSliceProperty("Port"),
"health_check":      swagger.NewObjectProperty("HealthCheck"),
```

The object names are declared and need to match in `/api/controllers/swagger.go`

api/provider/aws/load_balancer_update.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/quintilesims/layer0/common/models"
+)
+
+func (l *LoadBalancerProvider) Update(req models.UpdateLoadBalancerRequest) (*models.LoadBalancer, error) {
+	ports := req.Ports
```
zpatrick: Looks like you forgot to delete the body after copy+paste here?

api/controllers/load_balancer_test.go
```diff
@@ -162,3 +162,47 @@ func TestListLoadBalancers(t *testing.T) {
...
+		Ports:          &[]models.Port{},
+		HealthCheck: models.HealthCheck{
+			Target:             "80",
+			Interval:           60,
+			Timeout:            60,
+			HealthyThreshold:   3,
+			UnhealthyThreshold: 3,
+		},
+	}
+
```
zpatrick: Please target the `job_runner` branch, as it is the most up-to-date

api/provider/aws/load_balancer_update.go
jparsons04: No, this was an attempt to actually implement the function. I think I used [Read()](https://github.com/quintilesims/layer0/blob/274-scheduler-skeleton-jparsons/api/provider/aws/load_balancer_read.go#L13) as the basic structure for attempting to implement it

common/models/update_load_balancer_request.go
jparsons04: These sub-`Validate()` functions do exist, I checked the respective models.

api/controllers/load_balancer.go
jparsons04: :+1: Handler is now PUT against /loadbalancer

common/models/update_load_balancer_request.go
jparsons04: :+1:

common/models/update_load_balancer_request.go
jparsons04: :+1:

common/models/update_load_balancer_request.go
jparsons04: :+1:

api/controllers/load_balancer_test.go
jparsons04: :+1:

common/models/update_load_balancer_request.go
jparsons04: Just a quick note, I had to dereference u.Ports, but otherwise, I used this block verbatim.

api/controllers/controller.go
```diff
@@ -1,23 +1,33 @@
...
 	"github.com/quintilesims/layer0/api/job"
 	"github.com/quintilesims/layer0/common/models"
 	"github.com/zpatrick/fireball"
 )
 
-func scheduleJob(scheduler job.Scheduler, jobType job.JobType, req interface{}) (fireball.Response, error) {
-	job := models.ScheduleJobRequest{
-		JobType: jobType.String(),
-		Request: req,
+func createJob(store job.Store, jobType job.JobType, req interface{}) (fireball.Response, error) {
```
zpatrick: looks like you forgot to merge `job-runner` into your branch

api/provider/aws/load_balancer_update.go
zpatrick: Implementing the `provider` layer is kind of a beast on its own. I think bivers is trying to keep the implementation of these layers as discrete tasks.

jparsons04:
I had initially run into an issue using `make api` from the scripts directory because it appears that gomock is substituting RecordCall() with RecordCallWithMethodType(). Also had to make additional fixes to the UpdateLoadBalancer controller to address failing unit tests. I will take another look at what I've done tomorrow morning just to make sure your comments have been properly addressed.

jparsons04:
I still need to wire up the swagger declarations in `api/controllers/swagger.go`, so that will be forthcoming in another commit.

common/models/update_load_balancer_request.go
```diff
@@ -1,20 +1,28 @@
...
-	"fmt"
-
 	swagger "github.com/zpatrick/go-plugin-swagger"
 )
 
 type UpdateLoadBalancerRequest struct {
-	LoadBalancerID string
-	Ports          *[]Port     `json:"ports"`
-	HealthCheck    HealthCheck `json:"health_check"`
+	LoadBalancerID string       `json:""`
```
zpatrick: `json:"load_balancer_id"` (didn't mean `json:""` literal :smile: )

common/models/update_load_balancer_request.go
```diff
@@ -1,20 +1,28 @@
...
+	Ports          *[]Port      `json:"ports"`
+	HealthCheck    *HealthCheck `json:"health_check"`
 }
 
 func (u UpdateLoadBalancerRequest) Validate() error {
-	if u.LoadBalancerID == "" {
-		return fmt.Errorf("LoadBalancerID is required")
+	if u.Ports != nil {
+		for _, port := range *(u.Ports) {
```
zpatrick: [nit] no need for parens `range *u.Ports`

common/models/update_load_balancer_request.go
jparsons04: Ah I see, ok, this makes more sense.

zpatrick:
> I had initially run into an issue using make api from the scripts directory because it appears that gomock is substituting RecordCall() with RecordCallWithMethodType(). Also had to make additional fixes to the UpdateLoadBalancer controller to address failing unit tests. I will take another look at what I've done tomorrow morning just to make sure your comments have been properly addressed.

Should we all update `gomock` to the latest version?

jparsons04:
I'm not sure we should because I started to immediately run into issues with failing unittests after I ran `make api` from the scripts directory. I had to effectively checkout the previous version of the mock files with minor differences for the unittests to resolve.

api/provider/aws/load_balancer_update.go
```diff
@@ -0,0 +1,50 @@
+package aws
+
+import (
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/quintilesims/layer0/common/models"
+)
+
+func (l *LoadBalancerProvider) Update(req models.UpdateLoadBalancerRequest) error {
```
zpatrick: will merge for now, this will need to be revisited later

PR #346, opened 2017-09-14
Removed warning message when no auths property exists

### LINKED ISSUE #344: Invalid assumptions about the expected format a docker config file can be in.
### Expected behavior
Should be able to upload docker config file with the format below (without nesting in a parent `auths` property}

```
{
  "https://index.docker.io/v1/": {
    "auth": "zq212MzEXAMPLE7o6T25Dk0i",
    "email": "email@example.com"
  }
}
```
### Actual behavior
Doing so generates a warning during l0-setup init.

```
[WARNING] Even though you have specified a path to a docker config file it does not contain any auth information. If you need to add auth information to the docker config file, you can do so and re-run the l0-setup init command to include private registry authentication.
```

L0-setup init with a docker config file that has the `auths` property:
```
{
  "auths": {
    "https://index.docker.io/v1/": {
      "auth": "zq212MzEXAMPLE7o6T25Dk0i"
    }
  }
}
```

Version: `v0.10.3`

sesh-kebab:
Removes the warning message when an `auths` property doesn't exists in a docker config file being used for private repo authentication.

Run `l0-setup init <name>` with a docker config file like:
```
{
  "https://index.docker.io/v1/": {
    "auth": "zq212MzEXAMPLE7o6T25Dk0i",
    "email": "email@example.com"
  }
}
```

Even though the above format is valid, l0-setup generates a warning.

**Checklist**
- ~[ ] Unit tests~

closes #344

### DIFF
--- a/setup/instance/init.go
+++ b/setup/instance/init.go
@@ -124,17 +124,7 @@ func (l *LocalInstance) createOrWriteDockerCFG(dockerInputPath string) error {
 		return err
 	}
 
-	if len(config.Auths) == 0 {
-		fmt.Println("[WARNING] Even though you have specified a path to a docker config file, " +
-			"it does not contain any auth information. If you need to add auth information " +
-			"to the docker config file, you can do so and re-run the l0-setup init command to " +
-			"include private registry authentication.\n")
-
-		fmt.Println("Press 'enter' to continue without private registry authentication: ")
-
-		var input string
-		fmt.Scanln(&input)
-	} else if config.CredsStore != "" {
+	if config.CredsStore != "" {
 		fmt.Printf("[WARNING] You are using a credential store '%s'. "+
 			"Layer0 does not support credential store authentication.\n\n",
 			config.CredsStore)


zpatrick:
Version `0.10.0` of `l0-setup` chose to no longer support the old `~/.dockercfg` format in favor of the newer `~/.docker/config.json` format. In that regard, this is behaving as expected. 

Also, these changes remove any sort of validation of the auth file, opening the door for false positives. For example, a user could specify 
```
{ 
   "blah blah": "garbage"
}
```

and it would pass through `l0-setup` with out any notification that the auth schema is invalid.

sesh-kebab:
@zpatrick - the newer docker `~/.docker/config.json` format wraps config inside an `auths` object. Although the ecsagent strips this out anyway, the documentation requires ~/.docker/config.json` without the `auths` object. I.e., the below should be a valid but generates a warning during l0-setup init:

```
{
  "my-private-repo.io": {
    "auth": "abc123"
  }
}
```

> Newer versions of Docker create a configuration file as shown above with an outer auths object. The Amazon ECS agent only supports dockercfg authentication data that is in the below format, without the auths object. If you have the jq utility installed, you can extract this data with the following command: cat ~/.docker/config.json | jq .auths
http://docs.aws.amazon.com/AmazonECS/latest/developerguide/private-auth.html

zpatrick:
Yes, what you said about the ecs agent is true. The issue that you need to tackle in this portion of the code is: 

"How do we convert the user-specified docker authentication file user into `.dockercfg` format? How do we detect when the file is in invalid format?" 

Our current answer is:
1. Only allow the user-specified file to be in `config.json` format (since `.dockercfg` is depreciated) 
2. Marshal the user-specified file into an object that has the `auths` field.
3. Unmarshal **only** the `auths` field (a.k.a stripping it out like in the docs) when we write the file. See the [MarshalJSON](https://github.com/quintilesims/layer0/blob/develop/setup/docker/config.go#L23) implementation. 
4. If the user-specified file fails to marshal json, or the `auths` field is empty, then the file is in invalid format. 

This PR attempts to support user-specified files in `.dockercfg` **or** `config.json` format. This is probably a bigger ask than one would initially thought, given you need to adress:
* How do you reliably detect which format the user-specified file is in?
* How do you reliably perform validation on `.dockercfg` format, given the nature of its format is `map[string]interface{}` ?
* Is this feature worth the added complexity?

We tried to support both formats in past versions of `l0-setup`, but chose to drop it in the refactor because it added a lot of complexity and unreliability to support a format that is depreciated.

sesh-kebab:
Marshaling to our own object - I completely forgot about this step. You're absolutely right, this is working as intended.

PR #347, opened 2017-09-15
Job controller, runner, and insert hook

zpatrick:
Adds the job controller, the aws provider job runner, and adds an insert hook in to the job store. 
The job runner should likely be revisited to discover retry scenarios for jobs.


api/provider/aws/job_runner.go
```diff
@@ -1,25 +1,256 @@
...
+	case job.UpdateLoadBalancerJob:
+		return r.updateLoadBalancer(j.JobID, j.Request)
+	case job.UpdateServiceJob:
+		return r.updateService(j.JobID, j.Request)
+	default:
+		return fmt.Errorf("Unrecognized JobType '%s'", j.Type)
+	}
+}
+
+func (r *JobRunner) createDeploy(jobID, request string) error {
```
tlake: I really like how clean these patterns are. 👌

api/controllers/job.go
```diff
@@ -0,0 +1,63 @@
...
+	JobStore job.Store
+}
+
+func NewJobController(j job.Store) *JobController {
+	return &JobController{
+		JobStore: j,
+	}
+}
+
+func (e *JobController) Routes() []*fireball.Route {
```
tlake: [nit] I think `(e *JobController)` should be `(j *JobController)`.

api/provider/aws/job_runner.go
```diff
@@ -1,25 +1,256 @@
...
+
+	return r.jobStore.SetJobResult(jobID, task.TaskID)
+}
+
+func (r *JobRunner) deleteDeploy(jobID, deployID string) error {
+	return r.deployProvider.Delete(deployID)
+}
+
+func (r *JobRunner) deleteEnvironment(jobID, environmentID string) error {
+	loadBalancers, err := r.loadBalancerProvider.List()
```
sesh-kebab: On an initial look this doesn't seem right. The implementation details about how an environment should delete itself should be encapsulated at the environment level. But this isn't strictly an environment delete so I can see the case for why the method contains logic for deleting environment and all related entities.

api/provider/aws/job_runner.go
```diff
@@ -1,25 +1,256 @@
...
+	for _, task := range tasks {
+		if task.EnvironmentID == environmentID {
+			if err := r.deleteTask(jobID, task.TaskID); err != nil {
+				return err
+			}
 		}
+	}
 
-		return nil
+	return catchAndRetry(15, func() (shouldRetry bool, err error) {
```
sesh-kebab: Indicates leaky abstraction. Can this retry logic be encapsulated within the environment provider itself?

api/provider/aws/job_runner.go
```diff
@@ -1,25 +1,256 @@
...
+	task, err := r.taskProvider.Create(req)
+	if err != nil {
+		return err
+	}
+
+	return r.jobStore.SetJobResult(jobID, task.TaskID)
+}
+
+func (r *JobRunner) deleteDeploy(jobID, deployID string) error {
+	return r.deployProvider.Delete(deployID)
```
sesh-kebab: Why don't deletion jobs update the Job Store?

common/errors/code.go
```diff
@@ -1,14 +1,18 @@
...
 const (
-	InvalidRequest ErrorCode = 1 + iota
-	DeployDoesNotExist
-	EnvironmentDoesNotExist
-	JobDoesNotExist
-	LoadBalancerDoesNotExist
-	ServiceDoesNotExist
-	TaskDoesNotExist
-	UnexpectedError
+	InvalidRequest           ErrorCode = "InvalidReqest"
```
sesh-kebab: This is cool.

api/provider/aws/job_runner.go
```diff
@@ -1,25 +1,256 @@
...
+	return r.taskProvider.Delete(taskID)
+}
+
+func (r *JobRunner) updateEnvironment(jobID, request string) error {
+	var req models.UpdateEnvironmentRequest
+	if err := json.Unmarshal([]byte(request), &req); err != nil {
+		return errors.New(errors.InvalidRequest, err)
+	}
+
+	return r.environmentProvider.Update(req)
```
sesh-kebab: Same things as the deletes - why don't we update the JobStore here?

api/job/dispatcher.go
```diff
@@ -6,7 +6,7 @@ import (
 )
 
 const (
-	DISPATCHER_PERIOD = time.Second * 5
+	DISPATCHER_PERIOD = time.Minute * 5
```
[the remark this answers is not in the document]
zpatrick: I should probably add a comment - the 5 minutes relates to the burst read limit in dynamodb. The increase went up so high because this added the insertHook: instead of waiting every 5 seconds for a job to be picked up by the workers, the job get's immediately put into the queue. The 5 minutes is there to pick up any leftovers/jobs stuck way in the back of the queue.

api/job/dynamo_store.go
```diff
@@ -12,14 +12,16 @@ import (
...
+	insertHook func(jobID string)
 }
 
 func NewDynamoStore(session *session.Session, table string) *DynamoStore {
 	db := dynamo.New(session)
 
 	return &DynamoStore{
-		table: db.Table(table),
+		table:      db.Table(table),
+		insertHook: func(string) {},
```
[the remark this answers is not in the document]
zpatrick: No, it's meant to act like a hook, e.g. a webhook: perform X event after Y event happens.

api/job/dynamo_store.go
```diff
@@ -52,9 +54,14 @@ func (d *DynamoStore) Insert(jobType JobType, req string) (string, error) {
 		return "", err
 	}
 
+	d.insertHook(job.JobID)
 	return job.JobID, nil
 }
 
+func (d *DynamoStore) SetInsertHook(hook func(jobID string)) {
```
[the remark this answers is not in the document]
zpatrick: circular dependency on the store and dispatcher

api/provider/aws/job_runner.go
```diff
@@ -1,25 +1,256 @@
...
+
+func (r *JobRunner) createDeploy(jobID, request string) error {
+	var req models.CreateDeployRequest
+	if err := json.Unmarshal([]byte(request), &req); err != nil {
+		return errors.New(errors.InvalidRequest, err)
+	}
+
+	deploy, err := r.deployProvider.Create(req)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
zpatrick: nope, job worker is responsible for that.

api/provider/aws/job_runner.go
zpatrick: workers update the job status. outside of that, there is nothing for delete jobs to update.

api/provider/aws/job_runner.go
zpatrick: answered above

api/provider/aws/job_runner.go
zpatrick: adding changes in a future pr so we don't capture specific errors

api/provider/aws/job_runner.go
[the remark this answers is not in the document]
zpatrick: we could change the job runner signature to return `(string, error)`  instead of just `error`. The worker would know to set the `JobStatus` to whatever string was returned. That could help make it clearer which part is responsible for updating the job store.

PR #349, opened 2017-09-15
Deploy Provider

diemonster:
Adds functionality for Layer0 API's Deploy Provider, includes CR-D operations (no Update needed for Deploy). 
 

Manual testing for now. 

**Checklist**
- [x] Unit tests
- [X] Manual testing
~~- [ ] Smoke tests (if applicable)~~
~~- [ ] System tests (if applicable)~~
~~- [ ] Documentation (if applicable)~~


api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,131 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/quintilesims/layer0/common/models"
+)
+
+func (d *DeployProvider) Create(req models.CreateDeployRequest) (*models.Deploy, error) {
+	if req.DeployName == "" {
```
zpatrick: Validation should be done on the model, see [CreateEnvironmentRequest](https://github.com/quintilesims/layer0/blob/api-refactor/common/models/create_environment_request.go#L18)

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,131 @@
...
+		return nil, fmt.Errorf("DeployName is required")
+	}
+
```
zpatrick: We need to create a unique id for the deploy

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,131 @@
...
+	if err := d.TagStore.Insert(models.Tag{EntityID: deploy.DeployID, EntityType: "deploy", Key: "version", Value: deploy.Version}); err != nil {
+		return deploy, err
+	}
+
+	return deploy, nil
+}
+
+func (d *DeployProvider) createDeploy(deployName string, body []byte) (*models.Deploy, error) {
```
zpatrick: This helper is doing too much. The rule of thumb we've been using so far is:
* If a helper is wrapping an aws call, it should only be a single aws call
* A helper should do one "thing"
* The "Master" function (e.g. `CreateDeploy()`) is responsible for lining up all of these helper functions. 


This helper doing quite a bit: name validation, rendering, an aws call, and populating tags. 
I think [EnvironmentCreate](https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/environment_create.go#L49) is a good example showing how it splits out rendering `userData` and the aws api call into different functions.

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,131 @@
...
+	taskDef.SetContainerDefinitions(dockerrun.ContainerDefinitions)
+	taskDef.SetVolumes(dockerrun.Volumes)
+	// TODO: Understand why our model had to be updated to use ecs.TaskDefinitionPlacementConstraint
+	// rather than ecs.PlacementConstraint
+	taskDef.SetPlacementConstraints(dockerrun.PlacementConstraints)
+
+	return d.populateModel(taskDef)
+}
+
+func (d *DeployProvider) populateModel(taskDef *ecs.RegisterTaskDefinitionInput) (*models.Deploy, error) {
```
zpatrick: This function name and argument type isn't correct. I think it would make more sense to simply marshal the deploy's `*ecs.TaskDefintition` and place that into the body. This could probably be done in the "master" function:

```
taskDefinition, err := d.createTaskDefinition(...)
...

bytes, err := json.Marshal(taskDefinition)
....

model.Dockerrun = bytes
```

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,131 @@
...
+		DeployID:  L0DeployID(d.Config.Instance(), ecsDeployID),
+		Version:   GetRevision(d.Config.Instance(), ecsDeployID),
+		Dockerrun: dockerrun,
+	}
+
+	return deploy, nil
+}
+
+func (d *DeployProvider) createRenderedDockerrun(body []byte) (*models.Dockerrun, error) {
+	var dockerrun *models.Dockerrun
```
zpatrick: No need for this model anymore. We should just marshal straight into a `*ecs.TaskDefinition`.

diemonster: to clarify: you're saying we have no need for `models.Dockerrun`? As in, that model should be removed entirely?

diemonster: ✅

api/provider/aws/deploy_create.go
diemonster: ✅

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/quintilesims/layer0/common/models"
+)
+
+func (d *DeployProvider) Create(req models.CreateDeployRequest) (*models.Deploy, error) {
+	// TODO: see Line 59. Which entity ID generation pattern do we prefer? ARN to ID or the
+	// example below?
+	deployID := generateEntityID(req.DeployName)
+
+	if err := req.Validate(); err != nil {
```
zpatrick: request validation happens in the `controller` layer

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,102 @@
...
+	taskDefinition.SetVolumes(input.Volumes)
+	taskDefinition.SetFamily(aws.StringValue(input.Family))
+	taskDefinition.SetNetworkMode(aws.StringValue(input.NetworkMode))
+	taskDefinition.SetTaskRoleArn(aws.StringValue(input.TaskRoleArn))
+	taskDefinition.SetPlacementConstraints(input.PlacementConstraints)
+
+	return taskDefinition
+}
+
+func (d *DeployProvider) createRenderedDeploy(body []byte) (*ecs.TaskDefinition, error) {
```
zpatrick: [nit] let's be exact on what object types we are generating:  `renderTaskDefinition`

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,102 @@
...
+	var deployFile *ecs.TaskDefinition
```
zpatrick: [nit] `var taskDefinition *ecs.TaskDefinition`

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,102 @@
...
+	if err := req.Validate(); err != nil {
+		return nil, err
+	}
+
+	deployFile, err := d.createRenderedDeploy(req.DeployFile)
+	if err != nil {
+		return nil, err
+	}
+
+	familyName := L0DeployID(d.Config.Instance(), req.DeployName)
```
zpatrick: ahh! don't reintroduce the `L0DeployID` garbage!
The new pattern is something like:
```
deployID := generateEntityID(req.DeployName)
fqDeployID := addLayer0Prefix(deployID)
```

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,102 @@
...
+	if deployFile.Family != nil && deployFile.Family != aws.String(familyName) {
```
zpatrick: this should be handled in `renderTaskDefintion` helper

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,102 @@
...
+		return nil, fmt.Errorf("Custom family names are currently unsupported in Layer0")
+	}
+
+	input := &ecs.RegisterTaskDefinitionInput{}
```
zpatrick: this should be handled in a helper function

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,102 @@
...
+	// ecsDeployID := TaskDefinitionARNToECSDeployID(*input.TaskRoleArn)
+
+	taskDefinition := &ecs.TaskDefinition{}
+	taskDefinition.SetContainerDefinitions(input.ContainerDefinitions)
+	taskDefinition.SetVolumes(input.Volumes)
+	taskDefinition.SetFamily(aws.StringValue(input.Family))
+	taskDefinition.SetNetworkMode(aws.StringValue(input.NetworkMode))
+	taskDefinition.SetTaskRoleArn(aws.StringValue(input.TaskRoleArn))
+	taskDefinition.SetPlacementConstraints(input.PlacementConstraints)
+
```
zpatrick: missing `input.Validate()`  and the actual aws call

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	if err := d.TagStore.Insert(models.Tag{EntityID: deploy.DeployID, EntityType: "deploy", Key: "version", Value: deploy.Version}); err != nil {
+		return deploy, err
+	}
+
+	return deploy, nil
+}
+
+func (d *DeployProvider) createTaskDefinition(input *ecs.RegisterTaskDefinitionInput) *ecs.TaskDefinition {
```
zpatrick: matching the return value from `renderTaskDefinition(body []byte) (*ecs.TaskDefinition, error)`, 
the argument for `createTaskDefinition` should probably be a `*ecs.Task` as well. 
That way, the code is pretty simple in the "Master" function, e.g.:
```
func CreateDeploy(...) {
    fqDeployID := ...
    ...
    taskDefinition, err := renderTaskDefinition(req.DeployFile)
    if err != nil {
        return nil, err
    }

    createdTask, err := d.createTaskDefinition(fqDeployID, taskDefinition)
    if err != nil {
        return nil, err
    }

   ...
}
```

api/provider/aws/id.go
```diff
@@ -64,3 +65,26 @@ func hashNow() string {
 	salt := time.Now().Format(time.StampNano)
 	return fmt.Sprintf("%x", md5.Sum([]byte(salt)))
 }
+
+func TaskDefinitionARNToECSDeployID(arn string) string {
```
zpatrick: the stuff added in this file isn't needed, this is bad code to carryover

common/models/Dockerrun.go
```diff
@@ -5,10 +5,10 @@ import (
 )
 
 type Dockerrun struct {
```
zpatrick: I thought you were deleting this model?

common/models/create_deploy_request.go
```diff
@@ -1,16 +1,27 @@
...
-	Dockerrun  []byte `json:"dockerrun"`
+	DeployFile []byte `json:"deploy_file"`
 }
 
 func (c CreateDeployRequest) Validate() error {
 	if c.DeployName == "" {
 		return fmt.Errorf("Deploy Name is required")
 	}
 
+	if strings.Contains(c.DeployName, ".") {
```
zpatrick: This is ok to have, our new implementation should be able to work around this

api/provider/aws/deploy_create.go
diemonster: ✅

common/models/create_deploy_request.go
diemonster: ✅

common/models/Dockerrun.go
diemonster: ✅

api/provider/aws/id.go
diemonster: ✅

api/provider/aws/deploy_create.go
diemonster: ✅

api/provider/aws/deploy_create.go
diemonster: ✅

api/provider/aws/deploy_create.go
diemonster: ✅ there was some edge case (or what looked like it) handling of quotes and prefix mangling happening here; still trying to keep track of the reasons for the 3-4 extra helper functions in `id.go` before I delete them.

api/provider/aws/deploy_create.go
diemonster: ✅

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,136 @@
...
+	deployID := generateEntityID(req.DeployName)
+	fqDeployID := addLayer0Prefix(d.Config.Instance(), deployID)
+	familyName := fqDeployID
+
+	taskDefinition, err := d.renderTaskDefinition(req.DeployFile, fqDeployID)
+	if err != nil {
+		return nil, err
+	}
+
+	input := &ecs.TaskDefinition{}
```
zpatrick: generating the `input` for the aws call should happen inside the helper function. see [some examples](https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/environment_create.go#L125)

zpatrick: You already have an `*ecs.TaskDefinition` from line 30. This code is redundant.

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	deploy := &models.Deploy{
+		// TODO: we previously removed prefix from deployID ala below. Is this
+		// no longer needed?
+		// func removePrefix(id string) string {
+		// 	return strings.TrimPrefix(id, PREFIX)
+		// }
+		DeployID:   deployID,
+		Version:    d.getDeployRevision(deployID),
```
zpatrick: You get the deploy version from the `taskDefinitionOutput`

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,136 @@
...
+		DeployFile: bytes,
+	}
+
+	if err := d.TagStore.Insert(models.Tag{EntityID: deploy.DeployID, EntityType: "deploy", Key: "name", Value: req.DeployName}); err != nil {
```
zpatrick: Adding tags should be a separate helper function, see [create environment example](https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/environment_create.go#L193)

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,136 @@
...
+	input.SetNetworkMode(aws.StringValue(taskDefinitionRequest.NetworkMode))
+	input.SetContainerDefinitions(taskDefinitionRequest.ContainerDefinitions)
+	input.SetVolumes(taskDefinitionRequest.Volumes)
+	input.SetPlacementConstraints(taskDefinitionRequest.PlacementConstraints)
+
+	if err := input.Validate(); err != nil {
+		return nil, err
+	}
+
+	_, err := d.AWS.ECS.RegisterTaskDefinition(input)
```
zpatrick:
```
output, err := d.AWS.ECS.RegisterTaskDefinition(input)
if err != nil {
    return nil, err
}

return output.TaskDefinition, nil
```

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,136 @@
...
+	if err := json.Unmarshal(body, &taskDefinition); err != nil {
+		return nil, fmt.Errorf("Failed to decode deploy: %s", err.Error())
+	}
+
+	if len(taskDefinition.ContainerDefinitions) == 0 {
+		return nil, fmt.Errorf("Deploy must have at least one container definition")
+	}
+
+	if taskDefinition.Family != nil && taskDefinition.Family != aws.String(familyName) {
+		return nil, fmt.Errorf("Custom family names are currently unsupported in Layer0")
```
zpatrick: Use our own error type: `errors.New(errors.InvalidRequest, "Custom family names are unsupported in Layer0")`

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,136 @@
...
+				},
+			}
+		}
+	}
+
+	return taskDefinition, nil
+}
+
+// TODO: remove this if possible
+func (d *DeployProvider) getDeployRevision(id string) string {
```
zpatrick: remove - I think we should track deploy versions solely through tags

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,136 @@
...
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/quintilesims/layer0/common/models"
+)
+
+// Create takes a request from the Controller layer, formats the request,
```
zpatrick: This should be clear with well-written code

api/provider/aws/deploy_create.go
diemonster: What's the criteria for using our own error type versus `fmt.Errorf`?

api/provider/aws/deploy_create.go
diemonster: I'm fine with doing this, but it isn't quite consistent with what's in `environment`, and I think consistency is something really critical for this package. That said, with your `environment` helper functions, you aren't passing AWS objects around like I am... not sure if passing those objects is the best way forward here or not.

api/provider/aws/deploy_create.go
diemonster: Sure, but we get the value to place in the tag partially via this function. I'm fine with deleting this and putting the logic in Create() if need be, but it's just ugly

api/provider/aws/deploy_create.go
zpatrick: What's not consistent about it? 

We pass in `aws-sdk-go` objects into helpers when it makes the most sense (see [CreateLoadBalancer](https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/load_balancer_create.go#L148)). We also return `aws-sdk-go` objects from helper functions when it makes the most sense (see [ReadEnvironment](https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/environment_read.go#L48)).

api/provider/aws/deploy_create.go
zpatrick: If we can capture an error and assign a meaningful error code to it

api/provider/aws/deploy_create.go
diemonster: OK, but this contradicts the '~two helper function" standard we were going for (as I recall...)

api/provider/aws/deploy_create.go
zpatrick: The `version` should be taken from the `Revision` field of the [ecs.TaskDefinition](https://github.com/aws/aws-sdk-go/blob/master/service/ecs/api.go#L8494)

api/provider/aws/deploy_create_test.go
```diff
@@ -0,0 +1,114 @@
...
+		},
+	}
+
+	for _, tag := range expectedTags {
+		assert.Contains(t, tagStore.Tags(), tag)
+	}
+}
+
+func TestDeploy_renderTaskDefinition_InvalidRequest(t *testing.T) {
+	deploy := NewDeployProvider(nil, nil, nil)
```
zpatrick: Can we switch these to use a table test?
```
func TestDeploy_renderTaskDefinitionErrors(t *testing.T){
    cases := map[string]*ecs.TaskDefinition{
        "Custom Family Name": &ecs.TaskDefinition{...},
        "No Container Definitions": &ecs.TaskDefinition{},
    }
 
   ...
}
```

api/provider/aws/deploy_create_test.go
```diff
@@ -0,0 +1,114 @@
...
+		t.Fatal("Expected error was nil")
+	}
+}
+
+func TestDeploy_renderTaskDefinition(t *testing.T) {
+	deploy := NewDeployProvider(nil, nil, nil)
+
+	container := &ecs.ContainerDefinition{}
+	container.SetName("test_name")
+	logConfig := &ecs.LogConfiguration{
```
zpatrick: This test is a little backwards. Shouldn't we send in a valid `*ecsTaskDefinition` into `renderTaskDefinition` and then check that the log groups were added correctly?

api/provider/aws/deploy_create_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	taskDef := &ecs.TaskDefinition{}
+	taskDef.SetContainerDefinitions(containers)
+
+	bytes, err := json.Marshal(taskDef)
+	if err != nil {
+		t.Fatal("Failed to extract deploy file")
+	}
+
+	if _, err := deploy.renderTaskDefinition(bytes, "familyName"); err != nil {
```
zpatrick: The `renderTaskDefinition` doesn't need to be a function for the `DeployProvider` - it makes this more difficult to test since we need to do `deploy := NewDeployProvider(nil, nil, nil)`.

api/provider/aws/deploy_read.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/quintilesims/layer0/common/models"
+)
+
+func (d *DeployProvider) Read(deployID string) (*models.Deploy, error) {
+	fqDeployID := addLayer0Prefix(d.Config.Instance(), deployID)
```
zpatrick: This implementation will need to change w the new arn stuff - just marking it for now since I know it's WIP

api/provider/aws/deploy_read_test.go
```diff
@@ -0,0 +1,61 @@
...
+			t.Fatal(err)
+		}
+	}
+
+	result := &models.Deploy{}
+	if err := deploy.populateModelTags("deploy_id_1", result); err != nil {
+		t.Fatal(err)
+	}
+
+	assert.Equal(t, "deploy_name_1", result.DeployName)
```
zpatrick: We should probably test the `deploy_2` name/version got set

common/models/create_deploy_request.go
```diff
@@ -1,16 +1,22 @@
...
-	Dockerrun  []byte `json:"dockerrun"`
+	DeployFile []byte `json:"deploy_file"`
 }
 
 func (c CreateDeployRequest) Validate() error {
 	if c.DeployName == "" {
 		return fmt.Errorf("Deploy Name is required")
 	}
 
+	if c.DeployFile == nil {
```
zpatrick: [nit] checking `if len(c.DeployFile) == 0` is safe even if a slice is `nil`, and is a little bit more robust.

api/provider/aws/deploy_create_test.go
diemonster: Ahh good call, will do

api/provider/aws/deploy_create_test.go
diemonster: Having `renderTaskDefinition` as a method off of `DeployProvider` allows us to use `d.Config.Instance()` within the function, which we need when setting the LogConfiguration details. We _could_ pass that in as a arg for `renderTaskDefinition` I guess, but we're trading one slight improvement for another.

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,133 @@
...
+				},
+			}
+			container.SetLogConfiguration(logConfig)
+		}
+	}
+
+	return taskDefinition, nil
+}
+
+func (d *DeployProvider) createTags(model *models.Deploy, arn string) error {
```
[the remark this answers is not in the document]
diemonster: We use all of the fields of `models.Deploy` within `createTags`. Seems like more of a [nit] here.

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,133 @@
...
+
+	if err := json.Unmarshal(body, &taskDefinition); err != nil {
+		return nil, fmt.Errorf("Failed to decode deploy: %s", err.Error())
+	}
+
+	if len(taskDefinition.ContainerDefinitions) == 0 {
+		return nil, fmt.Errorf("Deploy must have at least one container definition")
+	}
+
+	if taskDefinition.Family != nil && taskDefinition.Family != aws.String(familyName) {
```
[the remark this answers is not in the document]
diemonster: ✅

common/models/create_deploy_request.go
diemonster: ✅

api/provider/aws/deploy_create_test.go
zpatrick: ah I didn't realize that. disregard me

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,133 @@
...
+	deployID := generateEntityID(req.DeployName)
+	fqDeployID := addLayer0Prefix(d.Config.Instance(), deployID)
+	familyName := fqDeployID
+
+	taskDefinition, err := d.renderTaskDefinition(req.DeployFile, familyName)
+	if err != nil {
+		return nil, err
+	}
+
+	taskDefinitionOutput, err := d.createTaskDefinition(taskDefinition)
```
zpatrick: [nit] this returns a `taskDefinition`, not a `taskDefinitionOutput` (since that object only exists in the helper)

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	bytes, err := json.Marshal(taskDefinitionOutput)
```
zpatrick: We don't need to do this part since it already exists in `req.DockerFile`

api/provider/aws/deploy_create.go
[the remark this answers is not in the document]
zpatrick: True - however, I think standardizing on a RoT in regards to how we write the `createTags`  helper functions for each entity makes sense. Right now, the `environment`, `task`, and `loadBalancer` implementations use the same pattern.

api/provider/aws/deploy_create.go
diemonster: gonna skip that one as we have a `taskDefinition` from `renderTaskDefinition` already. `Output` maybe isn't the greatest suffix in the world, but it helps delineate the return from `createTaskDefinition`.

api/provider/aws/deploy_create.go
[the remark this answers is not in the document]
diemonster: ✅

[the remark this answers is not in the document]
diemonster: On second thought, I'm seeing inconsistency in our docs / implementations elsewhere. As an example, several of the `populate...` functions from https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/README.md take models as args. And your impl of `populateModelTags` in `loadbalancer` also takes a model as well.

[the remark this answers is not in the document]
diemonster: perhaps a more worthy change would be to conform the function name to `populateModelTags` or something similar. That is more consistent with what we have elsewhere at least.

api/provider/aws/deploy_create.go
diemonster: Hmm, I would still rather marshal the output of `createTaskDefinition` and return that as part of `Create()` rather than passing the `req` back.

api/provider/aws/deploy_create_test.go
diemonster: ✅

api/provider/aws/deploy_create.go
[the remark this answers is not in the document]
diemonster: ✅

api/provider/aws/deploy_delete.go
```diff
@@ -0,0 +1,52 @@
+package aws
+
+import (
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/quintilesims/layer0/common/errors"
+)
+
+func (d *DeployProvider) Delete(deployID string) error {
+	fqDeployID := addLayer0Prefix(d.Config.Instance(), deployID)
```
zpatrick: the deploy id doesn't map directly to a family name. You'll need to do the task definition arn lookup here and pass that along to delete.

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,133 @@
...
+	"strconv"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/quintilesims/layer0/common/errors"
+	"github.com/quintilesims/layer0/common/models"
+)
+
+func (d *DeployProvider) Create(req models.CreateDeployRequest) (*models.Deploy, error) {
+	deployID := generateEntityID(req.DeployName)
```
zpatrick: So I was thinking about this yesterday, and there's a pretty significant flaw with this pattern is it stands: it doesn't leverage aws's versioning system at all. Each deploy will get a random ID, and will always be version `1` for that family. 

Fortunately, I think there is a relatively simple fix: instead of using the `fqDeployID` as the `familyName`, pass along the fully-qualified `req.DeployName`, e.g.

```
deployID := generateEntityID(req.DeployName)
familyName := addLayer0Prefix(d.Config.Instance(), eq.DeployName)

taskDefinition, err := d.renderTaskDefinition(req.DeployFile, familyName)
...
```

This change will allow us to leverage aws's versioning system, while still maintaining our entity ID system

zpatrick: @tlake: if we decide to make this changes, it'll have some effects on how we do deploy read and lists

diemonster: oddly enough, I made essentially these same changes this morning. After manual testing :p

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,141 @@
...
+
+	if len(taskDefinition.ContainerDefinitions) == 0 {
+		return nil, fmt.Errorf("Deploy must have at least one container definition")
+	}
+
+	if taskDefinition.Family != nil && aws.StringValue(taskDefinition.Family) != familyName {
+		return nil, errors.Newf(errors.InvalidRequest, "Custom family names are unsupported in Layer0")
+	}
+
+	// TODO: this is a strange bit of behavior... the network mode is being dropped from Unmarshaling.
```
diemonster: This is some strange behavior I encountered. Doesn't seem like we accounted for this in v0.10.3, so I'm guessing I introduced a bug or behavior has changed with latest aws sdk?

api/provider/aws/deploy_create.go
diemonster: @jparsons04 ^^^

jparsons04: I made this comment in my [documentation PR)(https://github.com/quintilesims/layer0/pull/365#discussion_r141735889), but if we are going to use `req.DeployName`, then I feel that [CreateTaskRequest](https://github.com/quintilesims/layer0/blob/api-refactor/common/models/create_task_request.go) also needs a notion of `DeployName`

diemonster: ✅

jparsons04: Disregard my previous statement. I don't think we're going to add `DeployName` to CreateTaskRequest.

api/controllers/swagger.go
```diff
@@ -296,6 +302,58 @@ func (s *SwaggerController) ServeSwaggerSpec(c *fireball.Context) (fireball.Resp
...
+							Description: "The added task",
+							Schema:      swagger.NewObjectSchema("Task"),
+						},
+					},
+				},
+			},
+			"/deploy/{id}": map[string]swagger.Method{
+				"get": {
+					Summary: "Describe a Deploy",
+					Tags:    []string{"Task"},
```
jparsons04: This needs to be `Tags:    []string{"Deploy"},` otherwise this will appear in the Task grouping of methods in swagger.

api/controllers/deploy_test.go
```diff
@@ -21,7 +21,7 @@ func TestCreateDeploy(t *testing.T) {
 
 	req := models.CreateDeployRequest{
 		DeployName: "deploy1",
-		Dockerrun:  ([]byte("content")),
+		DeployFile: ([]byte("content")),
```
zpatrick: [nit] you don't need those extra parens: `DeployFile: []byte("content"),`

api/controllers/deploy_test.go
```diff
@@ -71,9 +71,9 @@ func TestGetDeploy(t *testing.T) {
 	defer ctrl.Finish()
 
 	DeployModel := models.Deploy{
+		DeployFile: ([]byte("content")),
```
zpatrick: ditto

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	deploy := &models.Deploy{
+		DeployName: req.DeployName,
+		DeployID:   deployID,
+		Version:    strconv.FormatInt(aws.Int64Value(taskDefinitionOutput.Revision), 10),
+		DeployFile: bytes,
+	}
+
+	taskDefinitionArn := aws.StringValue(taskDefinitionOutput.TaskDefinitionArn)
```
zpatrick: [nit] go best-practice is to capitalize abbreviations: `taskDefinitionARN`

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,139 @@
...
+		return nil, fmt.Errorf("Failed to decode deploy: %s", err.Error())
+	}
+
+	taskDefinition.SetFamily(familyName)
+
+	if len(taskDefinition.ContainerDefinitions) == 0 {
+		return nil, fmt.Errorf("Deploy must have at least one container definition")
+	}
+
+	if taskDefinition.Family != nil && aws.StringValue(taskDefinition.Family) != familyName {
```
zpatrick: this line isn't needed since you set it above with: 
```
taskDefinition.SetFamily(familyName)
```

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,139 @@
...
+		return nil, errors.Newf(errors.InvalidRequest, "Custom family names are unsupported in Layer0")
+	}
+
+	// Valid NetworkMode options are [bridge, host, none], "bridge" is most common so defaulting to that.
+	if taskDefinition.NetworkMode == nil {
```
zpatrick: do we need this? Can't we let aws handle the default? Seems to be fine in the past. What if users don't want networking mode options?

api/provider/aws/deploy_create_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+func TestDeploy_renderTaskDefinition_Errors(t *testing.T) {
+	deploy := NewDeployProvider(nil, nil, nil)
+	model := &ecs.TaskDefinition{}
+
+	container := &ecs.ContainerDefinition{}
+	container.SetName("container_name")
+
+	testCases := map[string]*ecs.TaskDefinition{
+		"Custom Family Name": &ecs.TaskDefinition{
```
zpatrick: We can drop this case since the helper function explicitly set the familyName

api/provider/aws/deploy_delete.go
```diff
@@ -0,0 +1,55 @@
+package aws
+
+import (
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/quintilesims/layer0/common/errors"
+)
+
+func (d *DeployProvider) Delete(deployID string) error {
+	taskArn, err := d.lookupTaskDefinitionARN(deployID)
```
zpatrick: `taskARN`

api/provider/aws/deploy_delete.go
```diff
@@ -0,0 +1,55 @@
...
+func (d *DeployProvider) deleteDeploy(taskArn string) error {
+	input := &ecs.DeregisterTaskDefinitionInput{}
+	input.SetTaskDefinition(taskArn)
+
+	if err := input.Validate(); err != nil {
+		return err
+	}
+
+	if _, err := d.AWS.ECS.DeregisterTaskDefinition(input); err != nil {
+		if err, ok := err.(awserr.Error); ok && err.Code() == "does not exist" {
```
zpatrick: has this path been tested? I've never seen an aws api code that has spaces in it.

api/provider/aws/deploy_delete_test.go
```diff
@@ -0,0 +1,49 @@
...
+		},
+	}
+
+	for _, tag := range tags {
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	for _, tag := range tags {
```
zpatrick: this doesn't test how the function is/should be called. This part should be:
```
for _, tag := range tags {
    if err := tagStore.Insert(tag); err != nil {
        t.Fatal(err)
    }
}

if err := deploy.deleteDeployTags("deploy_id"); err != nil {
    t.Fatal(err)
}

assert.Len(t, tagStore.Tags(), 0)
```

api/provider/aws/deploy_list.go
```diff
@@ -0,0 +1,72 @@
...
+	taskDefinitionARNs := []string{}
+	fn := func(output *ecs.ListTaskDefinitionsOutput, lastPage bool) bool {
+		for _, taskDefinitionARN := range output.TaskDefinitionArns {
+			taskDefinitionARNs = append(taskDefinitionARNs, aws.StringValue(taskDefinitionARN))
+		}
+		return !lastPage
+	}
+
+	input := &ecs.ListTaskDefinitionsInput{}
+	if err := d.AWS.ECS.ListTaskDefinitionsPages(input, fn); err != nil {
```
zpatrick: This implementation is not very efficient. Listing taskDefinitionARNs for a layer0 instance should be done like this for optimum performance: 

* Use [ListTaskDefinitionFamiliesPages](https://github.com/aws/aws-sdk-go/blob/master/service/ecs/api.go#L1944) to get all of the families for a layer0 instance. The `input` should be constructed like so:
```
input := &ecs.ListTaskDefinitionFamiliesInput{}
// familyPrefix is the fully-qualified prefix for your layer0 instance, e.g. `l0-bivers-`
input.SetFamilyPrefix(familyPrefix)
// you probably should double-check this, but I believe we only want to 
// show 'active' task definitions. I think 'inactive' ones have been 'deleted'. 
input.SetStatus(ecs.TaskDefinitionFamilyStatusActive)
```

* Once you have a list of the task definition families, use [ListTaskDefinitionsPages](https://github.com/aws/aws-sdk-go/blob/master/service/ecs/api.go#L6492) for each family returned.  The `input` should be constructed like:
```
input := &ecs.ListTaskDefinitionsInput{}
// note that the 'FamilyPrefix' param is poorly named. This actually needs to be an exact match
// of the family name
input.SetFamilyPrefix(familyName)
// again, you probably should double-check this, but I believe we only want to 
// show 'active' task definitions. I think 'inactive' ones have been 'deleted'. 
input.SetStatus(ecs.TaskDefinitionFamilyStatusActive)
```

api/provider/aws/deploy_read.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	if tag, ok := tags.WithKey("arn").First(); ok {
+		return tag.Value, nil
+	}
+
+	return "", fmt.Errorf("Failed to find ARN for deploy '%s'", deployID)
+}
+
+func (d *DeployProvider) newDeployModel(deployID string) (*models.Deploy, error) {
```
zpatrick: [nit] `new`->`make`

api/provider/aws/deploy_read.go
```diff
@@ -0,0 +1,86 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	taskDefinitionARN, err := d.lookupTaskDefinitionARN(deployID)
+	if err != nil {
+		return nil, err
+	}
+
+	taskDefinitionOutput, err := d.describeTaskDefinition(taskDefinitionARN)
```
zpatrick: [nit] `taskDefinition, err := d.describe...`

api/provider/aws/deploy_read.go
```diff
@@ -0,0 +1,86 @@
...
+	"encoding/json"
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/quintilesims/layer0/common/errors"
+	"github.com/quintilesims/layer0/common/models"
+)
+
+func (d *DeployProvider) Read(deployID string) (*models.Deploy, error) {
+	deployModel, err := d.newDeployModel(deployID)
```
zpatrick: This seems a little "cart before the horse". The other read implementations follow the pattern:

* (optional) lookup the entities arn
* Describe the aws entity by it's arn or id
* Using the description, pass arguments along to `makeEntityModel`

So in this case, it would be something like:
```
deployFile, err := json.Marshal(...)
...

return makeDeployModel(deployID, deployFile)
```

api/provider/aws/deploy_read.go
```diff
@@ -0,0 +1,86 @@
...
+	deployModel.DeployFile = deployFile
+
+	return deployModel, nil
+}
+
+func (d *DeployProvider) describeTaskDefinition(taskDefinitionARN string) (*ecs.TaskDefinition, error) {
+	input := &ecs.DescribeTaskDefinitionInput{}
+	input.SetTaskDefinition(taskDefinitionARN)
+
+	output, err := d.AWS.ECS.DescribeTaskDefinition(input)
```
zpatrick: this should catch when the task does not exist

api/provider/aws/deploy_create.go
diemonster: In testing, AWS was requiring that `NetworkMode` was in the task definition, otherwise the error "NetworkMode must be specified" was returned. If a user wants no networking mode, they can select `none`.

api/provider/aws/deploy_create.go
diemonster: I know we've been over this before, but it ends up being a weird translation to/from AWS, who use `Arn`

api/provider/aws/deploy_delete.go
diemonster: The original code is here, it may be cruft, I couldn't repro the behavior:
https://github.com/quintilesims/layer0/blob/develop/api/backend/ecs/deploy_manager.go#L61

api/provider/aws/deploy_list.go
jparsons04: It's not clear how I would derive the `familyPrefix` if neither the public function List() nor the helper function listTaskDefinitionARNs are passed anything that could help derive what the `familyPrefix` should be.

api/provider/aws/deploy_read.go
jparsons04: :white_check_mark: 

I went with

```     output, err := d.AWS.ECS.DescribeTaskDefinition(input)
     if err != nil {
         if err, ok := err.(awserr.Error); ok && strings.Contains(err.Message(), "Unable to describe task definition") {
             return nil, errors.Newf(errors.DeployDoesNotExist, "Deploy '%s' does not exist", taskDefinitionARN)
         }
 
         return nil, err
     }```

File: api/provider/aws/deploy_delete.go (hunk 45 above)
Comment (jparsons04) [reply api/provider/aws/deploy_delete.go]: The ClientException that is thrown when `DeregisterTaskDefinition` can't find the task definition is "The specified task definition does not exist."

Comment (diemonster) [reply api/provider/aws/deploy_delete.go]: 👍

File: api/provider/aws/deploy_create.go [hunk 52]
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	taskDefinition.SetFamily(familyName)
+
+	if len(taskDefinition.ContainerDefinitions) == 0 {
+		return nil, fmt.Errorf("Deploy must have at least one container definition")
+	}
+
+	// Valid NetworkMode options are [bridge, host, none], "bridge" is most common so defaulting to that.
+	if taskDefinition.NetworkMode == nil {
```
zpatrick: are we sure we want this? Shouldn't users be able to specify no network mode?

api/provider/aws/deploy_create.go
```diff
@@ -0,0 +1,134 @@
...
+		taskDefinition.SetNetworkMode("bridge")
+	}
+
+	for _, container := range taskDefinition.ContainerDefinitions {
+		if container.LogConfiguration == nil {
+			logConfig := &ecs.LogConfiguration{
+				LogDriver: aws.String("awslogs"),
+				Options: map[string]*string{
+					"awslogs-group":         aws.String(fmt.Sprintf("l0-%s", d.Config.Instance())),
```
zpatrick: this should be `d.Config.LogGroupName()`, although that doesn't exist in this branch yet. See [here](https://github.com/quintilesims/layer0/pull/370/files#diff-c0668ce31281b12d2f7e67824be3a0baR15)

api/provider/aws/deploy_delete.go
zpatrick: so this will need to be updated to use the `strings.Contains` pattern then

api/provider/aws/deploy_create.go
diemonster: I had some bizarre behavior during testing that I can no longer repro. As such, this was added to be explicit about setting a default network mode, and I'm fine with removing it. That said, a user can specify no network mode by specifying `none` as a param to `NetworkMode` in their task definition.

api/provider/aws/deploy_create.go
diemonster: I can add that after the logs branches are merged --- I don't want the log implementation(s) to be a dependency on getting this PR closed.

api/provider/aws/deploy_create.go
zpatrick: perhaps we should set this block to `none` if it is nil and throw a comment like "hey, this needs to be explicitly set"

api/provider/aws/deploy_create.go
zpatrick: ok cool

zpatrick:
This is looking pretty good to me. There's still some cleanup to be done, mainly the unit test I put a comment on, and putting the `all bool` param to `deploy.List()`. 

But - given the size and how long this PR has been open, I'm ok with merging it now as long as we add those cleanup items immediately after.

Issue #274, opened 2017-07-07
Revisit aws api layer


zpatrick:
* [x] `controller` Package  -- For CRUD, filled letter means done, * means not yet, - means not needed.

**Note: Only Environment, LoadBalancer, and Service entities have the `Update` function!**

| Entity | Developer | Status | CRUD |
| - | - | - | - |
| Admin | nick | ✔️ | |
| Environment | tlake | ✔️ | CRUD |
| Service | tlake | ✔️ | CRUD |
| Task | Josh | ✔️  | CR-D |
| Deploy | Josh | ✔️  | CR-D |
| LB | Josh | ✔️  | CRUD |
| Job | - | - | - |
| Tag | Seshi | ✔️ |CR-D |

* [x] `provider/aws` Package

| Entity | Developer | Status | CRUD |
| - | - | - | - |
| Environment | - | ✔️ | CRUD |
| Service | tlake | ✔️ | CRUD |
| Task | Josh | ✔️ | CR-D |
| Deploy | Brandon/Josh | ✔️ | CR-D |
| LB | Zack | ✔️ | CRUD  |

* [x] `provider/aws` Unit Tests

| Entity | Developer | Status |
| - | - | - |
| Environment | - | - | 
| Service | tlake | ✔️ | 
| Task | Nick | ✔️ | 
| Deploy | Brandon | ✔️ | 
| LB | Josh | ✔️ | 

* [x] `client` Package

| Entity | Developer | Status |
| - | - | - |
| Environment | zack | ✔️ |
| Service | zack | ✔️ |
| Task | zack | ✔️ |
| Deploy | zack | ✔️ |
| LB | zack | ✔️ |

* [x] `command` Package

| Entity | Developer | Status |
| - | - | - |
| Environment | seshi | ✔️ |
| Service | tlake | ✔️ |
| Task | nick | ✔️ |
| Deploy | brandon | ✔️ |
| LB | josh | ✔️ |
| Resolver | zack | ✔️ |

* [x] `plugin/terraform` Package

| Entity | Developer | Status |
| - | - | - |
| Environment | zpatrick | qa |
| Service  | zpatrick | qa |
| Task   | zpatrick | qa |
| Deploy  | zpatrick | qa |
| LB    | zpatrick | qa  |

* [x] `job` Package
* [ ] `scaler` Package
   * [ ] #421 needs to be addressed
   * [ ] Does not scale horizontally
   * [ ] Should we be keeping ec2 instances around for at least an hour?
[...]

PR #350, opened 2017-09-18
274: Task List provider/aws implementation

jparsons04:
List() returns a slice of models.TaskSummary objects which describe the Task and the Environment it is contained in
Task entity CRUD methods will be split into separate files rather than one big file
listClusterNames() now in common.go since it is also used by Environment List
List Provider now includes the config.APIConfig object in the constructor
Unit test for populating Task Summaries

**Checklist**
- [x] Unit tests

Work is going toward #274

### DIFF
--- a/api/provider/aws/common.go
+++ b/api/provider/aws/common.go
@@ -2,10 +2,13 @@ package aws
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/aws/aws-sdk-go/service/ecs/ecsiface"
 	"github.com/quintilesims/layer0/api/tag"
 )
 
@@ -85,4 +88,26 @@ func deleteEntityTags(tagStore tag.Store, entityType, entityID string) error {
 	}
 
 	return nil
+}
+
+func listClusterNames(ecsapi ecsiface.ECSAPI, instance string) ([]string, error) {
+	clusterNames := []string{}
+	fn := func(output *ecs.ListClustersOutput, lastPage bool) bool {
+		for _, arn := range output.ClusterArns {
+			// cluster arn format: arn:aws:ecs:region:012345678910:cluster/name
+			clusterName := strings.Split(aws.StringValue(arn), "/")[1]
+
+			if hasLayer0Prefix(instance, clusterName) {
+				clusterNames = append(clusterNames, clusterName)
+			}
+		}
+
+		return !lastPage
+	}
+
+	if err := ecsapi.ListClustersPages(&ecs.ListClustersInput{}, fn); err != nil {
+		return nil, err
+	}
+
+	return clusterNames, nil
 }
--- a/api/provider/aws/task_list.go
+++ b/api/provider/aws/task_list.go
@@ -1,0 +1,93 @@
+package aws
+
+import (
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/quintilesims/layer0/common/models"
+)
+
+func (t *TaskProvider) List() ([]models.TaskSummary, error) {
+	clusterNames, err := listClusterNames(t.AWS.ECS, t.Config.Instance())
+	if err != nil {
+		return nil, err
+	}
+
+	summaries := []models.TaskSummary{}
+	for _, clusterName := range clusterNames {
+		runningTaskIDs, err := t.listTaskIDs(clusterName, ecs.DesiredStatusRunning)
+		if err != nil {
+			return nil, err
+		}
+
+		stoppedTaskIDs, err := t.listTaskIDs(clusterName, ecs.DesiredStatusStopped)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, taskID := range append(runningTaskIDs, stoppedTaskIDs...) {
+			summary := models.TaskSummary{
+				TaskID: taskID,
+			}
+			summaries = append(summaries, summary)
+		}
+
+	}
+
+	if err := t.populateTaskSummaries(summaries); err != nil {
+		return nil, err
+	}
+
+	return summaries, nil
+}
+
+func (t *TaskProvider) listTaskIDs(clusterName, status string) ([]string, error) {
+	taskIDs := []string{}
+	fn := func(output *ecs.ListTasksOutput, lastPage bool) bool {
+		for _, arn := range output.TaskArns {
+			// task arn format: arn:aws:ecs:region:012345678910:task/taskID
+			taskID := strings.Split(aws.StringValue(arn), "/")[1]
+			taskIDs = append(taskIDs, taskID)
+		}
+		return !lastPage
+	}
+
+	input := &ecs.ListTasksInput{}
+	input.SetCluster(clusterName)
+	input.SetDesiredStatus(status)
+
+	if err := t.AWS.ECS.ListTasksPages(input, fn); err != nil {
+		return nil, err
+	}
+
+	return taskIDs, nil
+}
+
+func (t *TaskProvider) populateTaskSummaries(summaries []models.TaskSummary) error {
+	environmentTags, err := t.TagStore.SelectByType("environment")
+	if err != nil {
+		return err
+	}
+
+	taskTags, err := t.TagStore.SelectByType("task")
+	if err != nil {
+		return err
+	}
+
+	for i, summary := range summaries {
+		if tag, ok := taskTags.WithID(summary.TaskID).WithKey("name").First(); ok {
+			summaries[i].TaskName = tag.Value
+		}
+
+		if tag, ok := taskTags.WithID(summary.TaskID).WithKey("environment_id").First(); ok {
+			summaries[i].EnvironmentID = tag.Value
+
+			if t, ok := environmentTags.WithID(tag.Value).WithKey("name").First(); ok {
+				summaries[i].EnvironmentName = t.Value
+			}
+		}
+	}
+
+	return nil
+}
--- a/api/provider/aws/task_read.go
+++ b/api/provider/aws/task_read.go
@@ -1,0 +1,9 @@
+package aws
+
+import (
+	"github.com/quintilesims/layer0/common/models"
+)
+
+func (t *TaskProvider) Read(taskID string) (*models.Task, error) {
+	return nil, nil
+}
--- a/api/main.go
+++ b/api/main.go
@@ -65,7 +65,7 @@ func main() {
 		environmentProvider := aws.NewEnvironmentProvider(client, tagStore, cfg)
 		loadBalancerProvider := aws.NewLoadBalancerProvider(client, tagStore, cfg)
 		serviceProvider := aws.NewServiceProvider(client, tagStore)
-		taskProvider := aws.NewTaskProvider(client, tagStore)
+		taskProvider := aws.NewTaskProvider(client, tagStore, cfg)
 		jobRunner := aws.NewJobRunner(
 			deployProvider,
 			environmentProvider,
--- a/api/provider/aws/environment_list.go
+++ b/api/provider/aws/environment_list.go
@@ -1,15 +1,11 @@
 package aws
 
 import (
-	"strings"
-
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/service/ecs"
 	"github.com/quintilesims/layer0/common/models"
 )
 
 func (e *EnvironmentProvider) List() ([]models.EnvironmentSummary, error) {
-	clusterNames, err := e.listClusterNames()
+	clusterNames, err := listClusterNames(e.AWS.ECS, e.Config.Instance())
 	if err != nil {
 		return nil, err
 	}
@@ -29,25 +25,6 @@ func (e *EnvironmentProvider) List() ([]models.EnvironmentSummary, error) {
 	}
 
 	return summaries, nil
-}
-
-func (e *EnvironmentProvider) listClusterNames() ([]string, error) {
-	output, err := e.AWS.ECS.ListClusters(&ecs.ListClustersInput{})
-	if err != nil {
-		return nil, err
-	}
-
-	clusterNames := []string{}
-	for _, arn := range output.ClusterArns {
-		// cluster arn format: arn:aws:ecs:region:012345678910:cluster/name
-		clusterName := strings.Split(aws.StringValue(arn), "/")[1]
-
-		if hasLayer0Prefix(e.Config.Instance(), clusterName) {
-			clusterNames = append(clusterNames, clusterName)
-		}
-	}
-
-	return clusterNames, nil
 }
 
 func (e *EnvironmentProvider) populateSummariesTags(summaries []models.EnvironmentSummary) error {
--- a/api/provider/aws/load_balancer_update.go
+++ b/api/provider/aws/load_balancer_update.go
@@ -1,9 +1,6 @@
 package aws
 
 import (
-	"strings"
-
-	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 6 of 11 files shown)


common/models/task_summary.go
```diff
@@ -5,4 +5,5 @@ type TaskSummary struct {
 	TaskName        string `json:"task_name"`
 	EnvironmentID   string `json:"environment_id"`
 	EnvironmentName string `json:"environment_name"`
+	OperatingSystem string `json:"operating_system"`
```
zpatrick: We don't need this field

api/provider/aws/task_list.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/quintilesims/layer0/common/config"
+	"github.com/quintilesims/layer0/common/errors"
+	"github.com/quintilesims/layer0/common/models"
+)
+
+const MAX_TASK_IDS = 100
+
+func (t *TaskProvider) List() ([]models.TaskSummary, error) {
+	clusterNames, err := t.listClusterNames()
```
zpatrick: This would be a good candidate for a shared function in `common.go`, since we know `ListEnvironments` uses this function

api/provider/aws/task_list.go
```diff
@@ -0,0 +1,151 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	taskCopies := map[string][]*ecs.Task{}
+	for _, clusterName := range clusterNames {
+		environmentID := delLayer0Prefix(t.Config.Instance(), clusterName)
```
zpatrick: I don't think this is correct. You should use the fully-qualified environment id in aws calls.

api/provider/aws/task_list.go
```diff
@@ -0,0 +1,151 @@
...
+
+		running := "RUNNING"
+		tasks, err := t.listTasksByStatus(&running, &environmentID)
```
zpatrick: We should use the most "simple" input types as arguments to helper functions. We want helper function to hide the ugliness of the aws api calls, so these should just be `string` inputs, not `*string` inputs.

zpatrick: [nit] perhaps rename `tasks` to `runningTasks`, so it's consistent with `stoppedTasks`

api/provider/aws/task_list.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+
+		stopped := "STOPPED"
+		stoppedTasks, err := t.listTasksByStatus(&stopped, &environmentID)
+		if err != nil {
+			return nil, err
+		}
+
+		taskARNs := append(tasks, stoppedTasks...)
+		if len(taskARNs) > 0 {
```
zpatrick: Couple things:

* You're putting aws api calls into the "master" function - this should be placed in a helper function
* I understand this is copied over from the previous implementation. However, it's a bad implementation that needs to be re-analyzed with a critical eye. A huge red flag is we are doing a `Describe` call during a `List`. We should avoid that as much as possible. 
* The notion of `taskCopies` can go away, as it was carryover from old implementations of tasks. We have all the info we need in the `listTasksByStatus`. We shouldn't need to make any more api calls after that.

api/provider/aws/task_list.go
```diff
@@ -0,0 +1,151 @@
...
+						}
+
+						taskCopies[*task.StartedBy] = append(taskCopies[*task.StartedBy], task)
+					}
+				}
+			}
+		}
+	}
+
+	getModel := func(tasks []*ecs.Task) (*models.Task, error) {
```
zpatrick: Seeing a lot of carryover from the old implementation. 

I think this "master" function should really only look something like this (psuedo code of course):

```
clusterNames, err := listClusterNames(...)

summaries := []*models.TaskSummary{}
for _, clusterName := range clusterNames {
    runningTaskARNs, err := listByStatus(clusterName, "RUNNING")
    stoppedTaskARNs, err := listByStatus(clusterName, "STOPPED")

    for _, taskARN := range append(runningTaskARNs, stoppedTaskARNs...) {
        summary := models.TaskSummary { ... }
        summaries = append(summaries, &summary)
    }
}

if err := t.populateTaskSummaries(summaries); err != nil { ... }

return summaries, nil
```

api/provider/aws/task_list.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+
+		ret = append(ret, output.Tasks...)
+	}
+
+	return ret, nil
+
+}
+
+func (t *TaskProvider) listTasksByStatus(status *string, environmentID *string) ([]*string, error) {
```
zpatrick: Good choice for a helper function :+1: 

Some nitpick items: 
* arguments should be as simple as possible, so use `string` args and not `*string`.
* follow the existing pattern for making aws api calls ([for example](https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/environment_create.go#L133)). The current pattern is much more readable .

api/provider/aws/task_read.go
```diff
@@ -0,0 +1,11 @@
+package aws
+
+import (
+	"github.com/quintilesims/layer0/common/models"
+)
+
+func (t *TaskProvider) Read(taskID string) (*models.Task, error) {
+	fqEnvironmentID := addLayer0Prefix(t.Config.Instance(), taskID)
```
zpatrick: you cannot fetch the environment id from the task id. This will probably be done via tag lookup

api/provider/aws/task_list.go
jparsons04: This and the previous comment came after I started to consider just returning the cluster ARNs from AWS rather than getting their names. Do you have a preference here whether I should use the prefix-included cluster name or the ARN?

zpatrick: I'd recommend moving the [existing function](https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/environment_list.go#L34) into `common.go` and just using that

api/provider/aws/common.go
```diff
@@ -86,3 +89,24 @@ func deleteEntityTags(tagStore tag.Store, entityType, entityID string) error {
 
 	return nil
 }
+
+func listClusterNames(ecsapi ecsiface.ECSAPI, instance string) ([]string, error) {
+	input := &ecs.ListClustersInput{}
+
+	output, err := ecsapi.ListClusters(input)
```
zpatrick: This should use [ListClustersPages](https://github.com/aws/aws-sdk-go/blob/master/service/ecs/api.go#L1511):

```
clusterNames = []string{}
fn := func(output *ListClustersOutput, lastPage bool) bool {
    for _, arn := range output.ClusterArns { ... }

    return !lastPage
}

if err :=  e.AWS.ECS.ListClustersPages(&ecs.ListClustersInput{}, fn); err != nil {
    return nil, err
}

return clusterNames, nil
```

api/provider/aws/task_list.go
```diff
@@ -0,0 +1,75 @@
+package aws
+
+import (
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/quintilesims/layer0/common/models"
+)
+
+func (t *TaskProvider) List() ([]*models.TaskSummary, error) {
+	clusterNames, err := listClusterNames(t.AWS.ECS, t.Config.Instance())
+
```
zpatrick: remove line

api/provider/aws/task_list.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	summaries := []*models.TaskSummary{}
+	for _, clusterName := range clusterNames {
+		runningTaskARNs, err := t.listTasksByStatus(clusterName, "RUNNING")
```
zpatrick: You always need to do `if err != nil {}` check after a function returns an error.
This swallows any errors returned on line 17.

api/provider/aws/task_list.go
```diff
@@ -0,0 +1,75 @@
...
+
+		if err != nil {
+			return nil, err
+		}
+
+		environmentID := delLayer0Prefix(t.Config.Instance(), clusterName)
+
+		for _, taskARN := range append(runningTaskARNs, stoppedTaskARNs...) {
+			summary := models.TaskSummary{
+				TaskID:        *taskARN,
```
zpatrick: `taskARN` != `taskID`. Also, use safe deferences from the aws package - I believe for strings it is `aws.StringValue(*string)`

api/provider/aws/task_list.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	return summaries, nil
+
+}
+
+func (t *TaskProvider) listTasksByStatus(clusterName, status string) ([]*string, error) {
+	input := &ecs.ListTasksInput{}
+	input.SetCluster(clusterName)
+	input.SetDesiredStatus(status)
```
zpatrick: need to do `input.Validate() `

zpatrick: This should be [ListTasksPages](https://github.com/aws/aws-sdk-go/blob/master/service/ecs/api.go#L2264)

api/provider/aws/task_list.go
```diff
@@ -0,0 +1,75 @@
...
+
+	tasks, err := t.AWS.ECS.ListTasks(input)
```
zpatrick: Similar to how we do `listClusterNames`, it would be nice to have this helper function convert the `[]*string` of taskARNs to a `[]string` of task ids.

e.g. `listTaskIDs(clusterName, status string) ([]string, error)`

api/provider/aws/task_list.go
```diff
@@ -0,0 +1,75 @@
...
+	clusterNames, err := listClusterNames(t.AWS.ECS, t.Config.Instance())
+
+	if err != nil {
+		return nil, err
+	}
+
+	summaries := []*models.TaskSummary{}
+	for _, clusterName := range clusterNames {
+		runningTaskARNs, err := t.listTasksByStatus(clusterName, "RUNNING")
+		stoppedTaskARNs, err := t.listTasksByStatus(clusterName, "STOPPED")
```
zpatrick: Instead of using a string, we could use the constants provided in aws-sdk-go: https://github.com/aws/aws-sdk-go/blob/master/service/ecs/api.go#L9159

`stoppedTaskARNs, err := t.listTasksByStatus(clusterName, ecs.DesiredStatusStopped)`

api/provider/aws/task_list.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func (t *TaskProvider) populateTaskSummaries(summaries []*models.TaskSummary) error {
+	tags, err := t.TagStore.SelectByType("task")
+	if err != nil {
+		return err
+	}
+
+	for i, summary := range summaries {
+		if tag, ok := tags.WithID(summary.TaskID).WithKey("task_name").First(); ok {
```
zpatrick: key is `name`, we also don't use an `environment_name` key. 
You should be able to copy+paste the logic from [here](https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/load_balancer_list.go#L53)

api/provider/aws/task_list.go
jparsons04: I'm not sure that Validate() is implemented in the sdk for either ListTasksInput or ListTasksPages?
```
$ grep -r "Validate" . | grep Task | grep ecs
./vendor/github.com/aws/aws-sdk-go/service/ecs/api.go:func (s *DeregisterTaskDefinitionInput) Validate() error {
./vendor/github.com/aws/aws-sdk-go/service/ecs/api.go:func (s *DescribeTaskDefinitionInput) Validate() error {
./vendor/github.com/aws/aws-sdk-go/service/ecs/api.go:func (s *DescribeTasksInput) Validate() error {
./vendor/github.com/aws/aws-sdk-go/service/ecs/api.go:func (s *RegisterTaskDefinitionInput) Validate() error {
./vendor/github.com/aws/aws-sdk-go/service/ecs/api.go:func (s *RunTaskInput) Validate() error {
./vendor/github.com/aws/aws-sdk-go/service/ecs/api.go:func (s *StartTaskInput) Validate() error {
./vendor/github.com/aws/aws-sdk-go/service/ecs/api.go:func (s *StopTaskInput) Validate() error {
./vendor/github.com/hashicorp/terraform/builtin/providers/aws/resource_aws_ecs_task_definition.go:
```
Validate() is there for the above providers, but doesn't appear to be there for List* methods. Unless I'm overlooking something else, I don't think I can simply invoke input.Validate().

zpatrick: oh weird, nvm then

jparsons04:
Are we concerned about the fact that I'm passing a slice of pointers to models.TaskSummary into the populateTaskSummaries helper function? Other entities I've seen with this pattern appear not to do that.

You had earlier suggested that I should append a summaries slice in preparation of populating the tags with
`summaries = append(summaries, &summary)` which leads me to believe I should be passing slices of pointers to it. Is there any justification for the departure here for Tasks?

zpatrick:
> Is there any justification for the departure here for Tasks?

I'd prefer consistency, so the non-pointer slice is the way to go IMO, I must have suggested the `append` line w/o looking at the existing pattern first.

api/provider/aws/common.go
```diff
@@ -86,3 +89,25 @@ func deleteEntityTags(tagStore tag.Store, entityType, entityID string) error {
 
 	return nil
 }
+
+func listClusterNames(ecsapi ecsiface.ECSAPI, instance string) ([]string, error) {
```
zpatrick: :+1: nice

api/provider/aws/task_list.go
```diff
@@ -0,0 +1,94 @@
...
+
+func (t *TaskProvider) List() ([]models.TaskSummary, error) {
+	clusterNames, err := listClusterNames(t.AWS.ECS, t.Config.Instance())
+	if err != nil {
+		return nil, err
+	}
+
+	summaries := []models.TaskSummary{}
+	for _, clusterName := range clusterNames {
+		runningTaskARNs, err := t.listTaskIDs(clusterName, ecs.DesiredStatusRunning)
```
zpatrick: looks like you forgot to update variable names `TaskARNs` -> `TaskIDs`

api/provider/aws/task_list.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	return summaries, nil
+}
+
+func (t *TaskProvider) listTaskIDs(clusterName, status string) ([]string, error) {
+	taskIDs := []string{}
+	fn := func(output *ecs.ListTasksOutput, lastPage bool) bool {
+		for _, arn := range output.TaskArns {
+			// task arn format: arn:aws:ecs:region:012345678910:task-definition/taskID:revision
```
zpatrick: This is an arn for a `task definition`, not a `task`. 

The  [AWS API Docs](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DescribeTasks.html) show task arns to be in the format:  `"arn:aws:ecs:us-east-1:012345678910:task/1dc5c17a-422b-4dc4-b493-371970c6c4d6"`

We should be able to get away with just: 
```
taskID := strings.Split(aws.StringValue(arn), "/")[1]
```

api/provider/aws/task_list.go
```diff
@@ -0,0 +1,94 @@
...
+	input.SetDesiredStatus(status)
+
+	if err := t.AWS.ECS.ListTasksPages(input, fn); err != nil {
+		return nil, err
+	}
+
+	return taskIDs, nil
+}
+
+func (t *TaskProvider) populateTaskSummaries(summaries []models.TaskSummary) error {
```
zpatrick: :+1: nice

api/provider/aws/task_list.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	return summaries, nil
+}
+
+func (t *TaskProvider) listTaskIDs(clusterName, status string) ([]string, error) {
+	taskIDs := []string{}
+	fn := func(output *ecs.ListTasksOutput, lastPage bool) bool {
+		for _, arn := range output.TaskArns {
+			// task arn format: arn:aws:ecs:region:012345678910:task/aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee
```
zpatrick: [nit] `arn:aws:ecs:region:012345678910:task/taskID`

PR #351, opened 2017-09-18
API Scaler and Dispatcher

zpatrick:
Adds a `scaler` package and an `EnvironmentScaler` skeleton in the `provider/aws` package. 
The `scaler` package handles the logic around scaling environments on a timer (every N minutes), and on demand (e.g. when creating a service). 

Manually trigger a CreateTask or CreateService request. The scaler should indicate it is running via logs. 

**Checklist**
- [x] Unit tests


api/job/dispatcher.go
```diff
@@ -6,6 +6,8 @@ import (
 )
 
 const (
+	// dynamodb allows a burt-read every 5 minutes
```
zpatrick: @sesh-kebab 
> That's a significant increase. Would it be better make to this configurable? Also why such a high default (relative to the previous values of 5 seconds).

api/provider/aws/job_runner.go
```diff
@@ -27,19 +28,19 @@ func NewJobRunner(
 	l provider.LoadBalancerProvider,
 	s provider.ServiceProvider,
 	t provider.TaskProvider,
-	store job.Store,
```
zpatrick: @sesh-kebab 
> Would you want to update the job status to failed here?

I've updated the runner to return a `string` representing the job's `Result`. Now this particular package doesn't even have access to the job store, so that should make the separation of responsibility more clear.

api/provider/aws/job_runner.go
```diff
@@ -68,189 +69,196 @@ func (r *JobRunner) Run(j models.Job) error {
...
+		}
 
-	return r.jobStore.SetJobResult(jobID, task.TaskID)
+		return task.TaskID, nil, false
+	})
 }
 
-func (r *JobRunner) deleteDeploy(jobID, deployID string) error {
-	return r.deployProvider.Delete(deployID)
+func (r *JobRunner) deleteDeploy(jobID, deployID string) (string, error) {
```
zpatrick: @sesh-kebab 
> Why don't deletion jobs update the Job Store?

We only populate the `result string` field when we create things -> it's the entity id of the new object.

api/provider/aws/job_runner.go
```diff
@@ -68,189 +69,196 @@ func (r *JobRunner) Run(j models.Job) error {
...
+				return "", err
 			}
 		}
 	}
 
-	return catchAndRetry(15, func() (shouldRetry bool, err error) {
+	return catchAndRetry(time.Minute*15, func() (result string, err error, shouldRetry bool) {
+		log.Printf("[DEBUG] [JobRunner] Deleting environment %s", environmentID)
 		if err := r.environmentProvider.Delete(environmentID); err != nil {
-			switch err := err.(type) {
```
zpatrick: @sesh-kebab 
> Indicates leaky abstraction. Can this retry logic be encapsulated within the environment provider itself?

Ignoring error capturing, just doing "blind" retries to avoid abstraction leaking

api/controllers/swagger.go
```diff
@@ -219,6 +226,31 @@ func (s *SwaggerController) ServeSwaggerSpec(c *fireball.Context) (fireball.Resp
 					},
 				},
 			},
+			"/task": map[string]swagger.Method{
```
zpatrick: ~ note that this was added in for testing. I know it's not complete yet ~

api/provider/aws/job_runner.go
zpatrick: I'd like to avoid that if possible. We had mixed retry logic before between the job runner and backend. I think keeping the retry logic isolated to one area of the code makes things much clearer.

api/scaler/dispatcher.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (d *Dispatcher) ScheduleRun(environmentID string) {
+	if timer, ok := d.schedule[environmentID]; ok {
+		log.Printf("[DEBUG] [ScalerDispatcher] Pushing back scheduled run for environment %s", environmentID)
+		timer.Stop()
+	}
+
+	log.Printf("[DEBUG] [ScalerDispatcher] Scaling environment %s in %v", environmentID, SCALE_GRACE_PERIOD)
+	d.schedule[environmentID] = time.AfterFunc(SCALE_GRACE_PERIOD*timeMultiplier, func() {
+		if _, ok := d.schedule[environmentID]; ok {
```
zpatrick: does this block make sense? perhaps it warrants a comment

common/models/task.go
```diff
@@ -1,15 +1,39 @@
...
+	TaskID          string `json:"task_id"`
+	TaskName        string `json:"task_name"`
+	EnvironmentID   string `json:"environment_id"`
+	EnvironmentName string `json:"environment_name"`
+	DeployID        string `json:"deploy_id"`
+	DeployName      string `json:"deploy_name"`
+	DeployVersion   int    `json:"deploy_version"`
+
+	// todo: fix copies field
+	//Copies       []TaskCopy `json:"copies"`
```
zpatrick: We should re-visit this at some point. This is a very leaky abstraction in the model. I think it may be better to change this to something much more layer0-centric, e.g. a slice of objects that have the container name and exit code?

```
type Task struct {
    ...
    Things []TaskThing
}

type TaskThing struct {
    ContainerName string
    ExitCode int
    // meta?
}
```

api/scaler/dispatcher.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"github.com/quintilesims/layer0/api/provider"
+)
+
+const (
+	SCALE_GRACE_PERIOD = time.Second * 15
+)
+
+var (
+	timeMultiplier time.Duration = 1
```
tlake: What's the reason for doing
```
var (
		timeMultiplier time.Duration = 1
	)
```
instead of
```	
timeMultiplier := time.Duration(1)
```
? Not necessarily a change or a nit, just wondering what you know that I don't. :)

api/scaler/dispatcher_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func setTimeMultiplier(m int) func() {
+	timeMultiplier = time.Duration(m)
+	return func() { timeMultiplier = 1 }
+}
+
+func TestDispatcherScheduleRun(t *testing.T) {
+	defer setTimeMultiplier(0)()
```
tlake: This is a cool/cute move. :+1:

api/scaler/dispatcher.go
```diff
@@ -0,0 +1,78 @@
...
+func NewDispatcher(e provider.EnvironmentProvider, s Scaler) *Dispatcher {
+	return &Dispatcher{
+		environmentProvider: e,
+		scaler:              s,
+		schedule:            map[string]*time.Timer{},
+	}
+}
+
+func (d *Dispatcher) ScheduleRun(environmentID string) {
+	if timer, ok := d.schedule[environmentID]; ok {
```
diemonster: I feel like we're mixing our terminology with AWS' terminology a bit regard nouns and verbs. AWS uses the term 'scale' when applied to `cluster`s, but from a user/client perspective, `service`s are the entity that is scaled, not `environment`. `Environment` instead uses `size`, (which may or may not be "correct").

common/models/task.go
tlake: Transforming what's currently just the output from ECS into a model more in-line with the rest of our implementation sounds like a fine idea to me. :+1:

api/scaler/dispatcher.go
zpatrick: I honestly didn't know you could do `timeMultiplier := time.Duration(1)` for package scoped variables. I thought it had to be `var x = y` or `var x type = y`

api/scaler/dispatcher_test.go
zpatrick: hashicorp++ (I stole it form them)

api/scaler/dispatcher.go
zpatrick: hmm, I'm not sure what a better alternative to "scale an environment" would be though?

The action we want is to get the exact right amount of ec2 instances in a cluster. Calibrate?

diemonster: It seems like, in general, we make things more confusing by attempting to rebrand AWS terms. `Scale` makes a lot of sense here because that's exactly what's happening AND it coincides with AWS' terminology. We should consider using a new verb for `service`s --- possibly just `AutoScale` to properly differentiate. This will probably have more implications for when we finally get around to implementing Application Load Balancers

common/models/task.go
diemonster: Is there an object we can borrow from the AWS SDK? Otherwise, would it be overkill to add this model to our models package?

zpatrick: My point is to move away from that. We basically do that currently, and it's a leaky abstraction that doesn't map well to how layer0 does tasks.

PR #352, opened 2017-09-19
Service provider: `Update`

tlake:
Adds `Update` to service provider.

**NOTE**
This PR contains some unfinished service provider work from Seshi. I will take responsibility for continuing this work going forward, now that he has other things to worry about.


api/provider/aws/service_update.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/quintilesims/layer0/common/models"
+)
+
+func (s *ServiceProvider) Update(req models.UpdateServiceRequest) error {
+	serviceID := req.ServiceID
+	deployID := req.DeployID
+	fqServiceID := addLayer0Prefix(s.Config.Instance(), serviceID)
+
+	serviceTags, err := s.TagStore.SelectByTypeAndID("service", serviceID)
```
zpatrick: I think all of this logic should go to a helper function in `common.go`:

```
func lookupEntityEnvironmentID(store tag.Store, entityType, entityID string) (string, error) { ... }
```

this would be used by services, tasks, and loadbalancers

api/provider/aws/service_update.go
```diff
@@ -0,0 +1,70 @@
...
+
+	environmentID := tag.Value
+	clusterName := addLayer0Prefix(s.Config.Instance(), environmentID)
+
+	service, err := s.readService(clusterName, serviceID)
+	if err != nil {
+		return err
+	}
+
+	var serviceScaleCount *int64
```
zpatrick: I think we should follow a more general pattern for all updates:
```
if req.OptionA != nil {
    updateOptionA(...)
}

if req.OptionB != nil {
    updateOptionB(...)
}

...

return nil
```

In that regard, I think we have a cleaner, more readable solution (even though it's a bit longer) by splitting out `updateServiceScale` and `updateServiceTaskDefinition` into different helper functions. 

That also lets us avoid calling `Read()` inside this function.

api/provider/aws/service_update.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	if err := s.updateService(clusterName, fqServiceID, deployID, serviceScaleCount); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (s *ServiceProvider) updateService(cluster, service string, taskDefinition *string, desiredCount *int64) error {
```
zpatrick: [nit] let's be very clear on parameters - that's bitten us in the past: `clusterName, serviceName, taskDefinitionID string`. 

Also, try to wrap aws-sdk-go's "ugliness" in helper functions, using `string` for args is much easier on the caller than `*string`.

api/provider/aws/service_update.go
```diff
@@ -0,0 +1,70 @@
...
+	input := &ecs.UpdateServiceInput{}
+	input.SetCluster(cluster)
```
zpatrick: assuming we change the params to `string`, aws has helper functions to convert to string pointers:
```
input.SetCluster(aws.String(clusterName))
input.SetService(aws.String(serviceName))
```

common/models/update_service_request.go
```diff
@@ -1,5 +1,44 @@
...
+	DeployID          *string `json:"deploy_id"`
+	ServiceScaleCount *int    `json:"service_scale_count"`
+}
+
+func (u UpdateServiceRequest) Validate() error {
+	if u.ServiceID == "" {
+		return fmt.Errorf("ServiceID must be specified")
+	}
+
+	if u.DeployID != nil {
```
zpatrick: `if u.DeployID != nil && *u.DeployID == "" {}`

- same for `u.ServiceScaleCount`

common/models/update_service_request.go
```diff
@@ -1,5 +1,44 @@
...
+	"fmt"
+
+	swagger "github.com/zpatrick/go-plugin-swagger"
+)
+
 type UpdateServiceRequest struct {
-	DeployID string `json:"deploy_id"`
+	ServiceID         string  `json:"service_id"`
+	DeployID          *string `json:"deploy_id"`
+	ServiceScaleCount *int    `json:"service_scale_count"`
```
zpatrick: [nit] `ServiceScaleCount` seems like a lot - can we just do `Scale`?

api/provider/aws/service_update.go
tlake: We're okay with unnecessarily making two `s.AWS.ECS.UpdateService(input)` calls when a user wants to update both scale and deploy at once? It doesn't seem like that would cause any issues, but I've tried to keep an eye on how many API calls we make since Jonathan et alia have been having rate limit problems.

Then again, I'm already making two AWS calls; one for reading, and one for updating. Stripping out the reading call in favor adding a _potential_ second update call is an improvement. 

Cool, I'm onboard. :+1:

api/provider/aws/service_update.go
tlake: ✅

api/provider/aws/service_update.go
zpatrick: The rate of aws calls is a good thing to keep in mind, but the create/update/delete calls pale in comparison to read and list calls.

api/provider/aws/service_update.go
tlake: TBH, we don't need pointers at all; they were just my solution to checking whether to add a field to the input, but with two separate update functions, actual values are what we want.

✅

api/provider/aws/service_update.go
tlake: ✅

api/provider/aws/service_update.go
tlake: ✅

common/models/update_service_request.go
tlake: ✅

common/models/update_service_request.go
tlake: ✅

api/provider/aws/common.go
```diff
@@ -9,6 +9,21 @@ import (
...
+func lookupEntityEnvironmentID(store tag.Store, entityType, entityID string) (string, error) {
+	tags, err := store.SelectByTypeAndID(entityType, entityID)
+	if err != nil {
+		return "", err
+	}
+
+	tag, ok := tags.WithKey("environment_id").First()
+	if !ok {
+		// is this canonical error handling?
+		return "", fmt.Errorf("Cannot resolve environment_id for %s '%s'", entityType, entityID)
```
zpatrick: We use `common/errors` when we want to capture meaningful error codes.

Speaking of,  this will likely be one of the first helper functions called for many of the `Read()` functions. 
If the entity does not exist, we won't capture that error code correctly. That will break the terraform provider since it checks for specific `errors.<Entity>DoesNotExist` error codes.

Perhaps we can check if `len(tags) == 0`? We would probably need a new helper function in `common/errors` that can do a  `switch` on `entityType string` and convert it to the proper error code.

api/provider/aws/service_update.go
```diff
@@ -0,0 +1,71 @@
...
+
+	fqEnvironmentID := addLayer0Prefix(s.Config.Instance(), environmentID)
+	clusterName := fqEnvironmentID
+
+	fqServiceID := addLayer0Prefix(s.Config.Instance(), serviceID)
+	serviceName := fqServiceID
+
+	if req.DeployID != nil {
+		taskDefinitionID := *req.DeployID
+		if err := s.updateServiceTaskDefinition(clusterName, serviceName, taskDefinitionID); err != nil {
```
zpatrick: The [docs](https://github.com/aws/aws-sdk-go/blob/master/service/ecs/api.go#L8927) show that the `TaskDefinition` needs to be in `family:revision` format. With our new deploy implementation, Layer0 deploy ids are just randomly generated strings; their `family` and `revision` properties are stored as tags, under `name` and `version`, respectively. 
This needs an extra step converting a layer0 deploy ID to the format `family:revision` - or, perhaps even better, we always pass around those 2 arguments are to separate things, e.g.:
```
 updateServiceTaskDefinition(clusterName, serviceName, taskDefinitionFamily string, taskDefinitionRevision int)
```

common/models/update_service_request.go
```diff
@@ -1,5 +1,40 @@
...
+	"fmt"
+
+	swagger "github.com/zpatrick/go-plugin-swagger"
+)
+
 type UpdateServiceRequest struct {
-	DeployID string `json:"deploy_id"`
+	ServiceID string  `json:"service_id"`
+	DeployID  *string `json:"deploy_id"`
+	Scale     *int    `json:"service_scale_count"`
```
zpatrick: need to update the struct tag

tlake: Oop, thanks :forehead_slap:

✅

api/provider/aws/common.go
tlake: I had blanked on the pattern before, but I think I was looking for something more like this:
```
    tag, ok := tags.WithKey("environment_id").First(); ok {
        return tag.Value, nil
    }

    // handle error here
```

api/provider/aws/service_update.go
tlake: ✅

api/provider/aws/common.go
tlake: @zpatrick Do we need to check `len(tags) == 0`? Is `!ok` not enough to result in a DNE error?

I've updated to `return "", errors.Newf(errors.EnvironmentDoesNotExist, "Could not resolve environment ID for %s '%s'", entityType, entityID)` on `!ok`, but maybe that's wrong. Should we be reporting that the entity doesn't exist instead of the environment?

api/controllers/service.go
```diff
@@ -78,3 +79,16 @@ func (s *ServiceController) ListServices(c *fireball.Context) (fireball.Response
 	return fireball.NewJSONResponse(200, summaries)
 
 }
+
+func (s *ServiceController) UpdateService(c *fireball.Context) (fireball.Response, error) {
```
zpatrick: Can you add the swagger endpoint please?

api/provider/aws/common.go
```diff
@@ -7,8 +7,48 @@ import (
...
 	"github.com/quintilesims/layer0/api/tag"
+	"github.com/quintilesims/layer0/common/errors"
 )
 
+func lookupEntityEnvironmentID(store tag.Store, entityType, entityID string) (string, error) {
+	tags, err := store.SelectByTypeAndID(entityType, entityID)
+	if err != nil {
+		return "", err
+	}
+
```
zpatrick: This is what I meant by my earlier comment:
```
tags, err := store.SelectByTypeAndID(...)
...

// if there are 0 tags for the specified type and id, I think it's reasonable to assume that entity does not exist
if len(tags) == 0 {
    // we will need a new function in common/errors to match 'entityType' to an error code, 
    // e.g. "environment" = errors.EnvironmentDoesNotExist
    return "", errors.NewEntityDoesNotExistError(entityType, entityID)
}

if tag, ok := tags.WithKey(...)

// we don't really have a meaningful error code  for this scenario (other than `UnexpectedError`), 
// so just use a basic error here
return "", fmt.Errorf("Could not resolve ...")
```

api/provider/aws/common.go
```diff
@@ -7,8 +7,48 @@ import (
...
+	if tag, ok := tags.WithKey("environment_id").First(); ok {
+		return tag.Value, nil
+	}
+
+	return "", errors.Newf(errors.EnvironmentDoesNotExist, "Could not resolve environment ID for %s '%s'", entityType, entityID)
+}
+
+func lookupTaskDefinitionFamily(store tag.Store, deployID string) (string, error) {
```
zpatrick: Can we merge these? Also, since we are looking up in our tags database, should the verbiage be in "layer0" terms? 
 `func lookupDeployNameAndVersion(store tag.Store, deployID string) (string, string, error)`

api/provider/aws/common.go
```diff
@@ -7,8 +7,48 @@ import (
...
+	tags, err := store.SelectByTypeAndID("deploy", deployID)
+	if err != nil {
+		return "", err
+	}
+
```
zpatrick: will probably want another `if len(tags) == 0 { return errors.Newf(errors.DeployDoesNotExist, "") }`

zpatrick: However, outside of that scenario, I'm not so confident in returning a `DeployDoesNotExist` error when we are missing only one particular tag.

api/provider/aws/service_update.go
```diff
@@ -0,0 +1,85 @@
...
+
+	return nil
+}
+
+func (s *ServiceProvider) updateServiceTaskDefinition(clusterName, serviceName, taskDefinitionFamily, taskDefinitionRevision string) error {
+	input := &ecs.UpdateServiceInput{}
+	input.SetCluster(clusterName)
+	input.SetService(serviceName)
+
+	taskDefinition := fmt.Sprintf("%s:%d", taskDefinitionFamily, taskDefinitionRevision)
```
zpatrick: `taskDefintionRevision` is a string, so this should be `%s:%s`

api/provider/aws/common.go
tlake: Aaaaah, I see - so both scenarios are useful here. :+1:

api/provider/aws/service_update.go
tlake: ✅

api/controllers/service.go
tlake: ✅

api/provider/aws/common.go
tlake: For sure.

✅

api/provider/aws/common.go

api/provider/aws/common.go

api/provider/aws/common.go
```diff
@@ -7,8 +7,48 @@ import (
...
+	tags, err := store.SelectByTypeAndID(entityType, entityID)
+	if err != nil {
+		return "", err
+	}
+
+	if tag, ok := tags.WithKey("environment_id").First(); ok {
+		return tag.Value, nil
+	}
+
+	return "", errors.Newf(errors.EnvironmentDoesNotExist, "Could not resolve environment ID for %s '%s'", entityType, entityID)
```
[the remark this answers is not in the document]

api/provider/aws/service_list.go
```diff
@@ -0,0 +1,170 @@
...
+
+import (
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/quintilesims/layer0/common/models"
+)
+
+func (s *ServiceProvider) List() ([]models.ServiceSummary, error) {
```
zpatrick: is this what was already in seshi's branch? if so, let's merge anyway for now. but still will need to be revisited.

common/errors/error.go
```diff
@@ -32,3 +32,22 @@ func (s *ServerError) Model() models.ServerError {
 		Message:   s.Err.Error(),
 	}
 }
+
+func ResolveErrorByEntityType(entityType string) (ErrorCode, error) {
```
zpatrick: [nit] this makes the ux for the caller weird, I think it should be handled similar to how errors in `fmt.Printf` are handled.
Also, given we know the caller's scenario, I think it should be something like:
```
func NewEntityDoesNotExistError(entityType, entityID string) *ServerError {
    switch entityType {
    case: "deploy": 
        return Newf(DeployDoesNotExist, "Deploy '%s' does not exist", entityID)
    ...
    default:
        Newf(UnexpectedError, "Entity (type='%s') '%s' does not exist", entityType, entityID) 
    }
}
```

That way, the caller can be like:
```
if len(tags) == 0 {
    return "", "", errors.NewEntityDoesNotExistError(entityType, entityID)
}
```

api/provider/aws/service_list.go
tlake: Yeah - our two branches seemed to be a little intertwined, but with Seshi being a new dad, I expect to take responsibility for all of service, so I can continue his work going forward.

common/errors/error.go
tlake: Yeah, that's better

✅

PR #354, opened 2017-09-20
Loadbalancer Update

zpatrick:
Adds the `Update` method to loadbalancer. 

Manually via swagger


jparsons04:
I should note that I also see this:

```
$ l0 job list
...
1505945640883415517                (Unknown JobType=0!)  (Unknown JobStatus=0!)  0001-01-01 00:00:00
```

zpatrick:
The `SerializationException` is an unrelated issue, but has been fixed in another branch i'm working on anyways. 

Don't use the CLI until we have finished the refactor, since we have dramatic changes to the API that the CLI has addressed yet.

PR #355, opened 2017-09-20
 Refactor Tasks

zpatrick:
This updates some of the underlying mechanisms of tasks due to the random-id generation done by the AWS API. 

Our current implementation basically does a "read all" request anytime we need to reference a particular task. This is very costly in terms of API limits, and (I think) can be improved quite a bit. This change set includes two main changes:
* Using the `startedBy` field when creating tasks, and then using that same field to filter tasks during `ListTasks`. 
* Adding an `arn` field to a task's tags once it has been created. This will be used to lookup the task's arn prior to running `DescribeTask`. 

TODO:
 * Unit Tests


api/provider/task.go
```diff
@@ -3,8 +3,9 @@ package provider
...
 
 type TaskProvider interface {
-	Create(req models.CreateTaskRequest) (*models.Task, error)
-	Delete(taskID string) error
-	List() ([]models.TaskSummary, error)
+	// todo: all creates should return a string
+	Create(req models.CreateTaskRequest) (string, error)
 	Read(taskID string) (*models.Task, error)
+	List() ([]models.TaskSummary, error)
+	Delete(taskID string) error
```
tlake: [nit] I think we decided to make these alphabetical, not CRUD-ordered (made the same comment on Josh's WIP PR)

api/provider/aws/task_read.go
```diff
@@ -1,9 +1,132 @@
...
 	"github.com/quintilesims/layer0/common/models"
 )
 
 func (t *TaskProvider) Read(taskID string) (*models.Task, error) {
-	return nil, nil
+	environmentID, err := lookupEntityEnvironmentID(t.TagStore, "task", taskID)
+	if err != nil {
+		return nil, err
+	}
+	fqEnvironmentID := addLayer0Prefix(t.Config.Instance(), environmentID)
```
tlake: [super duper nit} blank line after closing bracket

api/provider/aws/task_read.go
```diff
@@ -1,9 +1,90 @@
...
+func (t *TaskProvider) readTask(clusterName, taskARN string) (*ecs.Task, error) {
+	input := &ecs.DescribeTasksInput{}
+	input.SetCluster(clusterName)
+	input.SetTasks([]*string{aws.String(taskARN)})
+
+	if err := input.Validate(); err != nil {
+		return nil, err
+	}
+
+	// todo: catch does ntoe xist error
```
jparsons04: When you work on this, unfortunately DescribeTasks() also does not have a notion of Task not found: http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DescribeTasks.html#API_DescribeTasks_Errors. The closest equivalent you can use from this is `ClusterNotFoundException`

api/provider/aws/common.go
```diff
@@ -68,6 +68,13 @@ func getLoadBalancerRoleName(loadBalancerID string) string {
 	return fmt.Sprintf("%s-lb", loadBalancerID)
 }
 
+func taskFamilyRevisionFromARN(taskARN string) (string, string) {
```
diemonster: [nit] Since I'm not sure if this is called by anything external to Tasks yet, I have to ask: does this function have to live in `common.go`?

api/provider/aws/common.go
```diff
@@ -68,6 +68,13 @@ func getLoadBalancerRoleName(loadBalancerID string) string {
...
+	// task definition arn format: 'arn:aws:ecs:region:account:task-definition/family:version
```
diemonster: Appreciate comments like these 👍

api/provider/aws/task_read.go
```diff
@@ -1,9 +1,132 @@
...
```
diemonster: [nit] typo

api/provider/aws/task_read.go
```diff
@@ -1,9 +1,132 @@
...
+	output, err := t.AWS.ECS.DescribeTasks(input)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(output.Failures) > 0 {
+		// todo: catch does not exist eror rhere?
```
diemonster: [nit] typo

api/provider/aws/task_create.go
```diff
@@ -1,7 +1,90 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/quintilesims/layer0/common/models"
+)
+
+func (t *TaskProvider) Create(req models.CreateTaskRequest) (string, error) {
+	taskID := generateEntityID(req.TaskName)
+	fqEnvironmentID := addLayer0Prefix(t.Config.Instance(), req.EnvironmentID)
+
+	deployName, deployVersion, err := lookupDeployNameAndVersion(t.TagStore, req.DeployID)
```
diemonster: Possibly a controversial opinion, but i'd rather `lookupDeployNameAndVersion` returned a `models.Deploy`

api/provider/aws/task_create.go
```diff
@@ -1,7 +1,90 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	clusterName := fqEnvironmentID
+	startedBy := t.Config.Instance()
```
diemonster: what does `startedBy` mean?

diemonster: Sorry, I missed the comment in the MR originally...
"Using the startedBy field when creating tasks, and then using that same field to filter tasks during ListTasks."

diemonster: this variable name could be better... possibly a better would be more explicit like `taskStartedBy`

api/provider/aws/common.go
zpatrick: I expect it will with deploys and services, but since that isn't the case yet, I'll move it

api/provider/aws/task_read.go
zpatrick: :+1:

api/provider/task.go
zpatrick: :+1:

api/provider/aws/task_create.go
zpatrick: we made this function because we need to do operations on the deploy's name and version (for aws calls); returning `models.Deploy` would be very odd here.

api/provider/aws/task_create.go
zpatrick: that variable matches what is made in the aws call. This is a RoT instance where we explicitly set a new variable to show the relationships between layer0 and AWS concepts.

api/provider/aws/task_list_test.go
```diff
@@ -17,7 +17,13 @@ func TestTask_populateTaskSummaries(t *testing.T) {
...
 			EntityType: "task",
 			Key:        "name",
-			Value:      "tsk1",
+			Value:      "tname1",
+		},
+		{
+			EntityID:   "t1",
+			EntityType: "task",
+			Key:        "arn",
+			Value:      "arn1",
```
jparsons04: I get that this is mocking the ARN in the test, but I wonder if we should be using the true form of an ARN here when we test it so that we're testing closer to the actual values that will be passed into the function

`arn:aws:ecs:<region>:<aws_account_id>:container-instance/aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee`

api/provider/aws/task_list_test.go
```diff
@@ -26,16 +32,16 @@ func TestTask_populateTaskSummaries(t *testing.T) {
...
-			Value:      "ename1",
+			Value:      "tname2",
 		},
 		{
 			EntityID:   "t2",
 			EntityType: "task",
-			Key:        "name",
-			Value:      "tsk2",
+			Key:        "arn",
+			Value:      "arn2",
```
jparsons04: Same comment as above.

api/provider/aws/task_list_test.go
zpatrick: given we don't do any string manipulation on the task arns, this is functionally equivalent while being easier to read

jparsons04:
On commit [b422b95](https://github.com/quintilesims/layer0/pull/355/commits/b422b95d4b19af2eb4da3ac05dd6a6884355af80), I'm getting unit test errors against api/controllers/task_test.go in `TestGetTask`:

```
api $ go test ./...
# github.com/quintilesims/layer0/api/controllers
controllers/task_test.go:81: cannot use 5 (type int) as type string in field value
controllers/task_test.go:82: unknown field 'DesiredCount' in struct literal of type models.Task
controllers/task_test.go:85: unknown field 'PendingCount' in struct literal of type models.Task
controllers/task_test.go:86: unknown field 'RunningCount' in struct literal of type models.Task
?   	github.com/quintilesims/layer0/api	[no test files]
FAIL	github.com/quintilesims/layer0/api/controllers [build failed]
ok  	github.com/quintilesims/layer0/api/job	0.006s
?   	github.com/quintilesims/layer0/api/job/mock_job	[no test files]
?   	github.com/quintilesims/layer0/api/provider	[no test files]
ok  	github.com/quintilesims/layer0/api/provider/aws	0.009s
?   	github.com/quintilesims/layer0/api/provider/mock_provider	[no test files]
ok  	github.com/quintilesims/layer0/api/scaler	0.013s
ok  	github.com/quintilesims/layer0/api/tag	0.009s
```

Task.DeployVersion is now a string instead of an int.
Task.DesiredCount, Task.PendingCount, and Task.RunningCount were removed from the model.

PR #356, opened 2017-09-21
274: Task Provider, Delete()

jparsons04:
Implementation of Delete() for Task Provider

Unit tests

**Checklist**
- [x] Unit tests

closes #
Continuing work toward #274


jparsons04:
Just to be explicit about the immediate work, the core of this PR is to implement Delete() for the Task Provider, however, I've taken the time to fold in the recent PR that was accepted into api-refactor and the work being done in the task-refactor named branch.

api/controllers/task_test.go
```diff
@@ -20,11 +20,11 @@ func TestCreateTask(t *testing.T) {
 	controller := NewTaskController(mockTaskProvider, mockJobStore)
 
 	req := models.CreateTaskRequest{
-		ContainerOverrides: []models.ContainerOverride{},
-		Copies:             1,
-		DeployID:           "deploy_id",
-		EnvironmentID:      "env_id",
-		TaskName:           "task_name",
+		// TODO: Fix Copies field
+		//Copies:             1,
```
zpatrick: remove copies field (ditto with below)

api/provider/aws/job_runner.go
```diff
@@ -208,15 +208,7 @@ func (r *JobRunner) deleteEnvironment(jobID, environmentID string) (string, erro
 }
 
 func (r *JobRunner) deleteLoadBalancer(jobID, loadBalancerID string) (string, error) {
-	return catchAndRetry(time.Minute*15, func() (result string, err error, shouldRetry bool) {
```
zpatrick: can you keep this please

api/provider/aws/task_delete.go
```diff
@@ -1,5 +1,49 @@
...
+	}
+
+	if err := deleteEntityTags(t.TagStore, "task", taskID); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (t *TaskProvider) stopTask(clusterName, taskArn string) error {
```
zpatrick: `taskARN` (aws-sdk-go uses improper casing)

api/provider/aws/task_delete.go
```diff
@@ -1,5 +1,49 @@
...
+	input := &ecs.StopTaskInput{}
+	input.SetCluster(clusterName)
+	input.SetTask(taskArn)
+
+	if err := input.Validate(); err != nil {
+		return err
+	}
+
+	if _, err := t.AWS.ECS.StopTask(input); err != nil {
+		if err, ok := err.(awserr.Error); ok && err.Code() == "ClusterNotFoundException" {
```
zpatrick: Don't catch errors unless you plan on doing something meaningful with them. 

This should catch if the task does not exist, and `return nil`, since we want delete operations to be idempotent

api/provider/aws/common.go
```diff
@@ -20,7 +20,7 @@ func lookupEntityEnvironmentID(store tag.Store, entityType, entityID string) (st
 	}
 
 	if len(tags) == 0 {
-		return "", "", errors.NewEntityDoesNotExistError(entityType, entityID)
+		return "", errors.NewEntityDoesNotExistError(entityType, entityID)
```
tlake: Thanks for catching my mistake here. :smile_skin-tone-sheepish:

api/provider/aws/task_create.go
```diff
@@ -1,7 +1,88 @@
...
+
+func (t *TaskProvider) Create(req models.CreateTaskRequest) (string, error) {
+	taskID := generateEntityID(req.TaskName)
+	fqEnvironmentID := addLayer0Prefix(t.Config.Instance(), req.EnvironmentID)
+
+	clusterName := fqEnvironmentID
+	startedBy := t.Config.Instance()
+	// todo: deployName, deployVersion, err := lookupDeployNameAndVersion(...)
+
+	taskDefinitionFamily := "l0-zpatrick-test"
```
tlake: I assume `taskDefinitionFamily` and `taskDefinitionVersion` here are just placeholders for future function calls?

api/provider/aws/task_delete.go
```diff
@@ -1,5 +1,49 @@
...
+		return err
+	}
+	fqEnvironmentID := addLayer0Prefix(t.Config.Instance(), environmentID)
+
+	taskARN, err := t.lookupTaskARN(taskID)
+	if err != nil {
+		return err
+	}
+
+	clusterName := fqEnvironmentID
```
tlake: [super nit] To group similar concepts/work/actions/objects together, we can move this line up to be just under `fqEnvironmentID := ...`.

api/provider/aws/task_delete.go
jparsons04: Unfortunately, a Task not found error is not among the error codes returned from `StopTask()`:

http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_StopTask.html#API_StopTask_Errors

It seemed that the most specific error I could catch given the choices was the ClusterNotFoundException. I guess given this, I should just simply check for a non-nil `err` from the call to `StopTask()` and return it, otherwise return `nil`?

api/provider/aws/task_create.go
zpatrick: fwiw: this is pulled from my working branch. will be fixed in future pr

jparsons04: That is my assumption as well. I'm keeping an eye at the moment on progress being made on the task-refactor branch for what will come to replace these.

api/provider/aws/task_delete.go
jparsons04: Or perhaps just changing the return to `nil` due to its idempotency?

api/provider/aws/task_read.go
```diff
@@ -1,9 +1,90 @@
...
+		return nil, err
+	}
+	fqEnvironmentID := addLayer0Prefix(t.Config.Instance(), environmentID)
+
+	taskARN, err := t.lookupTaskARN(taskID)
+	if err != nil {
+		return nil, err
+	}
+
+	clusterName := fqEnvironmentID
```
tlake: Again, this could be moved up with `fqEnvironmentID := ...` to avoid fragmenting logic.

api/provider/task.go
```diff
@@ -3,8 +3,9 @@ package provider
...
 
 type TaskProvider interface {
-	Create(req models.CreateTaskRequest) (*models.Task, error)
-	Delete(taskID string) error
-	List() ([]models.TaskSummary, error)
+	// todo: all creates should return a string
+	Create(req models.CreateTaskRequest) (string, error)
 	Read(taskID string) (*models.Task, error)
+	List() ([]models.TaskSummary, error)
+	Delete(taskID string) error
```
tlake: [super nit] I think last week or so we had decided in favor of alphabetized ordering over CRUD ordering

jparsons04: Yes, we did. I'll make sure it's alphabetical here.

api/provider/aws/task_delete.go
zpatrick: You need to actually force this codepath to execute and see what type of error is returned by aws-sdk-go. 

Some examples: 
* [Without specific error codes (good)](https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/environment_delete.go#L60)
* [With specific error code (better) ](https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/environment_delete.go#L98)

api/provider/aws/task_delete.go
tlake: Actually, ignore this comment - our RoT sets up final arguments for helper functions just above the function calls; this looked strange to me because there's only one var being set up here.

api/provider/aws/task_read.go
tlake: Again, ignore this comment.

api/provider/aws/task_delete.go
jparsons04: Ok, given this and previous discussion about the lack of a Task Not Found, it appears the only option I have to force the codepath would be

```
if _, err := t.AWS.ECS.StopTask(input); err != nil {
    if err, ok := err.(awserr.Error); ok && err.Code() == "ClusterNotFoundException" {
        return nil
    }
}
```

api/provider/aws/task_delete.go
```diff
@@ -1,5 +1,50 @@
...
+	input := &ecs.StopTaskInput{}
+	input.SetCluster(clusterName)
+	input.SetTask(taskARN)
+
+	if err := input.Validate(); err != nil {
+		return err
+	}
+
+	if _, err := t.AWS.ECS.StopTask(input); err != nil {
+		if err, ok := err.(awserr.Error); ok && strings.Contains(err.Message(), "The referenced task was not found") {
```
zpatrick: usually do a shorter check on string match: `task was not found`

api/provider/aws/task_delete_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/quintilesims/layer0/api/tag"
+	"github.com/quintilesims/layer0/common/models"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTask_deleteEntityTags(t *testing.T) {
+	tagStore := tag.NewMemoryStore()
+	task := NewTaskProvider(nil, tagStore, nil)
+
+	tags := models.Tags{
```
zpatrick: should add a tag for the task's `arn`

PR #357, opened 2017-09-21
274: Service provider: Read

tlake:
Adds the `Read` operation to the Service Provider.

- Unit tests
- Run API locally
    - confirm compilation
    - hit http://localhost:9090/api/?url=/swagger.json with a browser to confirm Swagger functionality

**Checklist**
- [x] Unit tests
- ~System tests (if applicable)~
- ~Documentation (if applicable)~

-----

Apologies for the difficulty in reviewing this PR; the subject matter is both the current changes as well as a post-merge review of Seshi's work. You might not need to look at all of them, but in an attempt to ease this headache, the files immediately relevant to `Service Provider: Read()` are these:

```
common/models/
    - deployment.go
    - service.go

api/controllers/
    - service.go
    - service_test.go
    - swagger.go

api/job/
    - types.go

api/provider/
    - service.go

api/provider/aws/
    - common.go
    - service.go
    - service_read.go
```

If you have comments on parts of the code that aren't in the diff, I guess just make them here in the PR's conversation like so?

`api/provider/aws/service_read.go:Read()`:
 - `model` should be `&models.Service{}`, not `&models.SillyGoose{}`

### DIFF
--- a/api/provider/aws/README.md
+++ b/api/provider/aws/README.md
@@ -248,8 +248,16 @@ func (e *EntityProvider) readResourceB(args) (*aws.ResourceB, error) {
     // assuming `Describe()` returns `(ECSOutput, err)`
     // and `ECSOutput.ResourceBs` is `[]*ResourceB`
 	output, err := e.AWS.ECS.Describe(input)
-	if err != nil || len(output.ResourceBs) == 0 {
-        return nil, errors.Newf(errors.<Entity>DoesNotExist, "<Entity> '%s' does not exist")
+	if err != nil {
+        if err, ok := err.(awserr.Error); ok && err.Code() == "<Entity>NotFoundException" {
+            return nil, errors.Newf(errors.<Entity>DoesNotExist, "<Entity> '%s' does not exist", <entity>ID)
+        }
+
+        return nil, err
+    }
+
+    if len(output.ResourceBs) == 0 {
+        return nil, errors.Newf(errors.<Entity>DoesNotExist, "<Entity> '%s' does not exist", <entity>ID)
     }
 
 	return output.ResourceBs[0], nil
--- a/api/provider/aws/README.md
+++ b/api/provider/aws/README.md
@@ -248,12 +248,8 @@ func (e *EntityProvider) readResourceB(args) (*aws.ResourceB, error) {
     // assuming `Describe()` returns `(ECSOutput, err)`
     // and `ECSOutput.ResourceBs` is `[]*ResourceB`
 	output, err := e.AWS.ECS.Describe(input)
-	if err != nil {
-		return nil, err
-	}
-
-    if len(output.ResourceBs) == 0 {
-        return nil, fmt.Errorf("ResourceB not found")
+	if err != nil || len(output.ResourceBs) == 0 {
+        return nil, errors.Newf(errors.<Entity>DoesNotExist, "<Entity> '%s' does not exist")
     }
 
 	return output.ResourceBs[0], nil
--- a/api/provider/aws/README.md
+++ b/api/provider/aws/README.md
@@ -234,6 +234,9 @@ func (e *EntityProvider) readResourceA(args) (*aws.ResourceA, error) {
 	return output.ResourceA, nil
 }
 
+// if the helper function's AWS call returns a slice of objects,
+// and we expect `len(slice) == 0` or `len(slice) == 1`, standardize
+// on a pattern like this:
 func (e *EntityProvider) readResourceB(args) (*aws.ResourceB, error) {
 	input := &aws.Input{}
 	input.FieldA(args)
@@ -242,12 +245,18 @@ func (e *EntityProvider) readResourceB(args) (*aws.ResourceB, error) {
 		return nil, err
 	}
 
+    // assuming `Describe()` returns `(ECSOutput, err)`
+    // and `ECSOutput.ResourceBs` is `[]*ResourceB`
 	output, err := e.AWS.ECS.Describe(input)
 	if err != nil {
 		return nil, err
 	}
 
-	return output.ResourceB, nil
+    if len(output.ResourceBs) == 0 {
+        return nil, fmt.Errorf("ResourceB not found")
+    }
+
+	return output.ResourceBs[0], nil
 }
 
 func (e *EntityProvider) populateModelTags(entityID string, model *models.Entity) error {
--- a/api/provider/aws/service_read.go
+++ b/api/provider/aws/service_read.go
@@ -1,10 +1,9 @@
 package aws
 
 import (
-	"fmt"
-
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/quintilesims/layer0/common/errors"
 	"github.com/quintilesims/layer0/common/models"
 )
 
@@ -16,36 +15,54 @@ func (s *ServiceProvider) Read(serviceID string) (*models.Service, error) {
 
 	fqEnvironmentID := addLayer0Prefix(s.Config.Instance(), environmentID)
 	clusterName := fqEnvironmentID
-	service, err := s.readService(clusterName, serviceID)
+	ecsService, err := s.readService(clusterName, serviceID)
 	if err != nil {
 		return nil, err
 	}
 
-	model := &models.Service{
-		DesiredCount: aws.Int64Value(service.DesiredCount),
-		RunningCount: aws.Int64Value(service.RunningCount),
-		PendingCount: aws.Int64Value(service.PendingCount),
-	}
-
-	for _, deploy := range service.Deployments {
+	var deployments []models.Deployment
+	for _, deploy := range ecsService.Deployments {
 		deployID := aws.StringValue(deploy.TaskDefinition)
+
 		//todo: convert deployid to layer0 deploy id
 
-		deploy := models.Deployment{
-			DeploymentID: aws.StringValue(deploy.Id),
-			Created:      aws.TimeValue(deploy.CreatedAt),
-			Updated:      aws.TimeValue(deploy.UpdatedAt),
-			Status:       aws.StringValue(deploy.Status),
-			PendingCount: aws.Int64Value(deploy.PendingCount),
-			RunningCount: aws.Int64Value(deploy.RunningCount),
-			DesiredCount: aws.Int64Value(deploy.DesiredCount),
-			DeployID:     deployID,
+		d := models.Deployment{
+			Created:       aws.TimeValue(deploy.CreatedAt),
+			DeployID:      deployID,
+			DeployName:    "", // tag
+			DeployVersion: "", // tag
+			DesiredCount:  int(aws.Int64Value(deploy.DesiredCount)),
+			PendingCount:  int(aws.Int64Value(deploy.PendingCount)),
+			RunningCount:  int(aws.Int64Value(deploy.RunningCount)),
+			Status:        aws.StringValue(deploy.Status),
+			Updated:       aws.TimeValue(deploy.UpdatedAt),
 		}
 
-		model.Deployments = append(model.Deployments, deploy)
+		deployments = append(deployments, d)
 	}
 
-	if err := s.updateWithTagInfo(model, serviceID); err != nil {
+	var loadBalancerID string
+	if len(ecsService.LoadBalancers) != 0 {
+		loadBalancer := ecsService.LoadBalancers[0]
+		loadBalancerName := aws.StringValue(loadBalancer.LoadBalancerName)
+		fqLoadBalancerID := loadBalancerName
+		loadBalancerID = delLayer0Prefix(s.Config.Instance(), fqLoadBalancerID)
+	}
+
+	model := &models.Service{
+		Deployments:      deployments,
+		DesiredCount:     int(aws.Int64Value(ecsService.DesiredCount)),
+		EnvironmentID:    environmentID,
+		EnvironmentName:  "", // tag
+		LoadBalancerID:   loadBalancerID,
+		LoadBalancerName: "", // tag
+		PendingCount:     int(aws.Int64Value(ecsService.PendingCount)),
+		RunningCount:     int(aws.Int64Value(ecsService.RunningCount)),
+		ServiceID:        serviceID,
+		ServiceName:      "", // tag
+	}
+
+	if err := s.populateModelTags(serviceID, model); err != nil {
 		return nil, err
 	}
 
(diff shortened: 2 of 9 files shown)


api/provider/aws/service_read.go
```diff
@@ -9,17 +9,23 @@ import (
...
 	service, err := s.readService(clusterName, serviceID)
 	if err != nil {
 		return nil, err
 	}
 
-	model.DesiredCount = aws.Int64Value(service.DesiredCount)
-	model.RunningCount = aws.Int64Value(service.RunningCount)
-	model.PendingCount = aws.Int64Value(service.PendingCount)
+	model := &models.Service{
+		DesiredCount: aws.Int64Value(service.DesiredCount),
```
zpatrick: I may be alone here, if so disregard, but I think it makes more sense to have the `*Count` fields in `models.Service` be `int` instead of `int64`.

api/provider/aws/service_read.go
```diff
@@ -9,17 +9,23 @@ import (
...
+		RunningCount: aws.Int64Value(service.RunningCount),
+		PendingCount: aws.Int64Value(service.PendingCount),
+	}
```
zpatrick: [line 35] `models.Deployment` should not have the `DeploymentID` field. That's a leaked abstraction.

zpatrick: [line 48] please use the standard name in regards to  ` s.updateWithTagInfo` -> `populateModelTags`

api/provider/aws/service_read.go
```diff
@@ -58,9 +64,9 @@ func (s *ServiceProvider) readService(clusterName, serviceID string) (*ecs.Servi
 		return nil, err
 	}
 
-	for _, service := range output.Services {
-		return service, nil
+	if len(output.Services) == 0 {
+		return nil, fmt.Errorf("ECS service '%s' in cluster '%s' does not exist", serviceID, clusterName)
```
zpatrick: It's important we return specific error codes when entities do not exist. This is because the terraform plugin checks for that code specifically. 

```
return nil, errors.Newf(errors.ServiceDoesNotExist, "Service %s does not exist", serviceID)
```

api/provider/aws/service_read.go
zpatrick: that helper function should exist in this file. Also, `ListTags()` should not be using that function.  It should have a special function that is optimized for list instead of read (current implementation makes N calls to the db. Optimized one can make only 2 or 3).

api/provider/aws/service_read.go
```diff
@@ -9,17 +9,23 @@ import (
...
+	clusterName := fqEnvironmentID
 	service, err := s.readService(clusterName, serviceID)
 	if err != nil {
 		return nil, err
 	}
 
-	model.DesiredCount = aws.Int64Value(service.DesiredCount)
-	model.RunningCount = aws.Int64Value(service.RunningCount)
-	model.PendingCount = aws.Int64Value(service.PendingCount)
+	model := &models.Service{
```
zpatrick: you forgot to add service id, environment id, loadbalancer id

api/provider/aws/README.md
```diff
@@ -242,12 +245,18 @@ func (e *EntityProvider) readResourceB(args) (*aws.ResourceB, error) {
 		return nil, err
 	}
 
+    // assuming `Describe()` returns `(ECSOutput, err)`
+    // and `ECSOutput.ResourceBs` is `[]*ResourceB`
 	output, err := e.AWS.ECS.Describe(input)
 	if err != nil {
 		return nil, err
```
zpatrick: `return nil, errors.Newf(errors.<Entity>DoesNotExist, "<Entity> %s does not exist", <entity>ID)`

api/provider/aws/service_read.go
tlake: Nah, I also prefer `int`s here. ✅

api/provider/aws/service_read.go
tlake: ✅

api/provider/aws/service_read.go
tlake: ✅

api/provider/aws/service_read.go
tlake: ✅

api/provider/aws/README.md
tlake: ✅

api/provider/aws/service_read.go
```diff
@@ -1,45 +1,68 @@
...
+	if err != nil {
+		return nil, err
+	}
 
-	for _, deploy := range service.Deployments {
+	var deployments []models.Deployment
+	for _, deploy := range ecsService.Deployments {
 		deployID := aws.StringValue(deploy.TaskDefinition)
+
 		//todo: convert deployid to layer0 deploy id
```
zpatrick: shouldn't this be handled in this pr?

api/provider/aws/service_read.go
```diff
@@ -1,45 +1,68 @@
...
+		loadBalancerName := aws.StringValue(loadBalancer.LoadBalancerName)
+		fqLoadBalancerID := loadBalancerName
+		loadBalancerID = delLayer0Prefix(s.Config.Instance(), fqLoadBalancerID)
+	}
+
+	model := &models.Service{
+		Deployments:      deployments,
+		DesiredCount:     int(aws.Int64Value(ecsService.DesiredCount)),
+		EnvironmentID:    environmentID,
+		EnvironmentName:  "", // tag
```
zpatrick: I think calling out which fields get tags is a step in the right direction. I'm hoping we can think of a way to make the relationship even clearer, but that is out of scope for this pr.

api/provider/aws/service_read.go
```diff
@@ -54,13 +77,72 @@ func (s *ServiceProvider) readService(clusterName, serviceID string) (*ecs.Servi
 	})
 
 	output, err := s.AWS.ECS.DescribeServices(input)
+	if err != nil || len(output.Services) == 0 {
```
zpatrick: This reads nice, but unfortunately the `err != nil` case isn't specific enough, and this will lead to us mislabeling errors. We should only return a `ServiceDoesNotExist` if/when we catch the specific `awserr.Error` with the right code or message. 

```
if err != nil {
    if err, ok := err.(*awserr.Error); ok && err.Code == "xxx" {
        return nil, errors.Newf(...)
    }

    return err
}

if len(output.Services) == 0 {
    return nil, errors.Newf(..)
}

return output.Services[0], nil
```

api/provider/aws/service_read.go
```diff
@@ -54,13 +77,72 @@ func (s *ServiceProvider) readService(clusterName, serviceID string) (*ecs.Servi
...
 
-	for _, service := range output.Services {
-		return service, nil
+	if tag, ok := tags.WithKey("name").First(); ok {
+		model.ServiceName = tag.Value
 	}
 
-	return nil, fmt.Errorf("ecs service '%s' in cluster '%s' does not exist", serviceID, clusterName)
+	if model.EnvironmentID != "" {
+		tags, err := s.TagStore.SelectByTypeAndID("environment", model.EnvironmentID)
```
zpatrick: I think the "environment" and "loadbalancer" blocks should live together:

```
if tag, ok := tags.WithKey("environment_id").First(); ok {
    environmentID := tag.Value
    model.EnvironmentID = environmentID

    tags, err := s.TagStore.SelectByTypeAndID(...)
```

api/provider/aws/service_read.go
```diff
@@ -54,13 +77,72 @@ func (s *ServiceProvider) readService(clusterName, serviceID string) (*ecs.Servi
...
+
+		if tag, ok := tags.WithKey("name").First(); ok {
+			deploy.DeployName = tag.Value
+		}
+
+		if tag, ok := tags.WithKey("version").First(); ok {
+			deploy.DeployVersion = tag.Value
+		}
+
+		deployments = append(deployments, deploy)
```
zpatrick: This seems odd, I assume it's a carryover from the previous implementation. Can't we just update these in place instead of re-creating all of the deployments?

api/provider/aws/service_read.go
tlake: Yeah, for sure

api/provider/aws/service_read.go
tlake: Aaah, gotcha.

✅

api/provider/aws/service_read.go
tlake: ✅

api/provider/aws/service_read.go
tlake: ✅

api/provider/aws/service_read.go
```diff
@@ -77,13 +80,38 @@ func (s *ServiceProvider) readService(clusterName, serviceID string) (*ecs.Servi
...
 	return output.Services[0], nil
 }
 
+func (s *ServiceProvider) lookupDeployIDFromTaskDefinitionARN(taskDefinitionARN string) (string, error) {
+	tags, err := s.TagStore.SelectByType("deploy")
+	if err != nil {
+		return "", err
+	}
+
+	if len(tags) == 0 {
```
zpatrick: this block isn't necessary

api/provider/aws/service_read.go
```diff
@@ -77,13 +80,38 @@ func (s *ServiceProvider) readService(clusterName, serviceID string) (*ecs.Servi
...
+		return "", errors.Newf(errors.DeployDoesNotExist, "No deploys exist")
+	}
+
+	if tag, ok := tags.WithValue(taskDefinitionARN).First(); ok {
```
zpatrick: should be `tags.WithKey("arn").WithValue(taskDefinitionARN).First()`

api/provider/aws/service_read.go
```diff
@@ -124,25 +152,20 @@ func (s *ServiceProvider) populateModelTags(serviceID string, model *models.Serv
...
-		tags, err := s.TagStore.SelectByTypeAndID("deploy", deploy.DeployID)
+	for _, d := range model.Deployments {
+		tags, err := s.TagStore.SelectByTypeAndID("deploy", d.DeployID)
 		if err != nil {
 			return err
 		}
 
-		if tag, ok := tags.WithKey("name").First(); ok {
-			deploy.DeployName = tag.Value
+		if tag, ok := tags.WithKey("deploy_name").First(); ok {
```
zpatrick: nothing exists with the key `deploy_name` as far as I'm aware. I think you meant `"name"`?

api/provider/aws/service_read.go
tlake: ✅

api/provider/aws/service_read.go
tlake: ✅

api/provider/aws/service_read.go
tlake: Oop, yeah, my mistake.

✅

api/provider/aws/service_read.go
tlake: ✅

api/provider/aws/service_read.go
```diff
@@ -1,48 +1,67 @@
...
-	if err := s.updateWithTagInfo(model, serviceID); err != nil {
+	var loadBalancerID string
+	if len(ecsService.LoadBalancers) != 0 {
+		loadBalancer := ecsService.LoadBalancers[0]
+		loadBalancerName := aws.StringValue(loadBalancer.LoadBalancerName)
+		fqLoadBalancerID := loadBalancerName
+		loadBalancerID = delLayer0Prefix(s.Config.Instance(), fqLoadBalancerID)
+	}
+
+	model, err := s.newServiceModel(environmentID, loadBalancerID, serviceID)
```
diemonster: I made a similar comment in the WIP PR for this pattern, but I'll ask here as well: do the other CRUD operations _not_ need to call `newServiceModel`? If they do, we need to change the name for the functions in this pattern or generalize them more. Also not convinced that putting them directly in the various CRUD files makes sense versus (in this case) `service.go`.

tlake: `newServiceModel()` should only be relevant for Read, just like the prior `populateModelTags()` was only relevant for Read. List, then, would replace `populateSummariesTags()` with something like `newServiceSummary()`, and that would likewise only be relevant to List. Since other operation functions don't call these functions, I don't think they belong in `service.go`, but should stay in their respective `service_<operation>.go` files.

api/provider/aws/README.md
```diff
@@ -242,12 +245,14 @@ func (e *EntityProvider) readResourceB(args) (*aws.ResourceB, error) {
...
 	}
 
+    // assuming `Describe()` returns `(ECSOutput, err)`
+    // and `ECSOutput.ResourceBs` is `[]*ResourceB`
 	output, err := e.AWS.ECS.Describe(input)
-	if err != nil {
-		return nil, err
-	}
+	if err != nil || len(output.ResourceBs) == 0 {
```
zpatrick: This needs to be updated to reflect the changes we addressed earlier.

api/provider/aws/service_read.go
```diff
@@ -55,12 +74,91 @@ func (s *ServiceProvider) readService(clusterName, serviceID string) (*ecs.Servi
...
+	}
+
+	if len(output.Services) == 0 {
+		return nil, errors.Newf(errors.ServiceDoesNotExist, "Service '%s' does not exist", serviceID)
+	}
+
+	return output.Services[0], nil
+}
+
+func (s *ServiceProvider) lookupDeployIDFromTaskDefinitionARN(taskDefinitionARN string) (string, error) {
```
zpatrick: This should be in `common.go` since `task.Read()` will be using this.

api/provider/aws/service_read.go
```diff
@@ -55,12 +74,91 @@ func (s *ServiceProvider) readService(clusterName, serviceID string) (*ecs.Servi
...
+	if err != nil {
+		return nil, err
 	}
 
-	return nil, fmt.Errorf("ecs service '%s' in cluster '%s' does not exist", serviceID, clusterName)
+	if tag, ok := tags.WithKey("name").First(); ok {
+		model.ServiceName = tag.Value
+	}
+
+	if environmentID != "" {
```
zpatrick: I don't think this block is required. It should be safe to assume `environmentID` will never be empty string (unlike loadBalancerID)

api/provider/aws/service_read_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+import (
+	"testing"
+
+	"github.com/quintilesims/layer0/api/tag"
+	"github.com/quintilesims/layer0/common/models"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestService_lookupDeployIDFromTaskDefinition(t *testing.T) {}
```
zpatrick: WIP?

tlake: Oh crap! Yeah, amidst all our debating, I completely forgot that I hadn't finished this yet.

api/provider/aws/service_read.go
tlake: ✅

api/provider/aws/service_read_test.go
tlake: ✅

api/provider/aws/README.md
tlake: ✅

api/provider/aws/service_read.go
tlake: ✅

zpatrick:
:+1: LGTM

Issue #338, opened 2017-08-31
Run `terraform init` on l0-setup init


zpatrick:
The `init` command of `l0-setup` should run `terraform init` after running [`terraform get`](https://github.com/quintilesims/layer0/blob/develop/setup/instance/init.go#L44). 
This is required for [v10+ of Terraform](https://www.terraform.io/upgrade-guides/0-10.html).

PR #358, opened 2017-09-25
Added terraform init to l0-setup init

nehayward:
This adds the "terraform init" to the l0-setup.

Clone branch, cd to layer0/setup and run `go run main.go init <instance name>`
You should see the terraform success message then

**Checklist**
- [ ] Unit tests

**Concerns**
screen shot 2017-09-25 at 3 32 45 pm
- I've removed the color in the output, but I'm wondering if this should be muted too since it's confusing that it tells the user they can run terraform plan.
- No download indicator when running the `terraform get` command

closes #
#338

### DIFF
--- a/setup/terraform/terraform.go
+++ b/setup/terraform/terraform.go
@@ -39,6 +39,10 @@ func (t *Terraform) Get(dir string) error {
 	return t.run(dir, "get", "-update")
 }
 
+func (t *Terraform) Init(dir string) error {
+	return t.run(dir, "init", "-no-color")
+}
+
 func (t *Terraform) Output(dir, key string) (string, error) {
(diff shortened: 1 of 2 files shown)


setup/terraform/terraform.go
```diff
@@ -39,6 +39,10 @@ func (t *Terraform) Get(dir string) error {
 	return t.run(dir, "get", "-update")
 }
 
+func (t *Terraform) Init(dir string) error {
+	return t.run(dir, "init", "-no-color")
```
zpatrick: [nit] I don't think we need to explicitly set `-no-color` here, it makes it seem like things will break if someone were to remove it.

PR #362, opened 2017-09-26
Use 'makeEntityModel' function helpers

zpatrick:

### DIFF
--- a/api/provider/aws/load_balancer_read.go
+++ b/api/provider/aws/load_balancer_read.go
@@ -18,7 +18,12 @@ func (l *LoadBalancerProvider) Read(loadBalancerID string) (*models.LoadBalancer
 		return nil, err
 	}
 
-	ports := make([]models.Port, len(loadBalancer.ListenerDescriptions))
+	model, err := l.newModel(loadBalancerID)
+	if err != nil {
+		return nil, err
+	}
+
+	model.Ports = make([]models.Port, len(loadBalancer.ListenerDescriptions))
 	for i, description := range loadBalancer.ListenerDescriptions {
 		port := models.Port{
 			ContainerPort: aws.Int64Value(description.Listener.InstancePort),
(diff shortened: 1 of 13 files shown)


tlake:
So this pattern really isn't a very big departure from our current one, as I understand it; it boils down to just a change in sequence.

### Current sequence:
- make AWS call
- instantiate model and populate it with values from the AWS call
- pass model into `populateModelTags`
   - get "key" values from model and perform tag lookups
   - update the model with values from the tag database

### New sequence:
- make AWS call
- pass relevant values used for tag lookups into `newModel`
   - perform tag lookups, instantiate model with tag-relevant values
   - return partial model
- populate model with values from AWS call

So what we gain from the new proposal is explicitness in which fields are used in a tag lookup, not modifying the state of an existing model in a helper function, and a helper function that actually returns something that's not an error. It also keeps the helper function small and focused, and doesn't create a huge function signature in order to handle the complete population of a model inside the helper function.

I like it.

zpatrick:
@tlake yep, exactly right. I think `service.Read` will be an psuedo-outlier depending on how we want to populate the `Deployments` field, but I think every other entity read/list helper can follow this pattern.

tlake:
@zpatrick I was thinking the same thing with 2 helpers. 👍

api/provider/aws/load_balancer_read.go
```diff
@@ -80,10 +76,14 @@ func (l *LoadBalancerProvider) describeLoadBalancer(loadBalancerName string) (*e
 	return output.LoadBalancerDescriptions[0], nil
 }
 
-func (l *LoadBalancerProvider) populateModelTags(loadBalancerID string, model *models.LoadBalancer) error {
+func (l *LoadBalancerProvider) newModel(loadBalancerID string) (*models.LoadBalancer, error) {
```
diemonster: These functions are super specific to their use case in this particular CRUD operation, but have such a general name. In the `Create()` portion, we'd also need to call a `newModel` function, wouldn't we?

PR #365, opened 2017-09-27
Add GoDoc compliant comments for exported methods in provider

jparsons04:
Adds comments for the following exported methods in provider:
* Deploy's Create(), List(), and Read()
* Task's Create(), Delete(), List(), and Read()
* Load Balancer's Create(), Delete(), List(), and Read()
* Environment's Create(), Delete(), List(), Read(), and Update()
* Service's List(), Read(), and Update()

Are the comments clear? Grammar/spell check. Does it align with Go's style guidelines?

**Checklist**
~- [ ] Unit tests~


api/provider/aws/deploy_read.go
```diff
@@ -9,6 +9,9 @@ import (
 	"github.com/quintilesims/layer0/common/models"
 )
 
+// Read returns a Deploy based on the provided Deploy ID. The Deploy ID is used to look up
```
diemonster: This one gave me slight pause --- seems like more context around what a 'Deploy' is would be useful if this is going into GoDoc style docs

api/provider/aws/task_delete.go
```diff
@@ -7,6 +7,8 @@ import (
 	"github.com/aws/aws-sdk-go/service/ecs"
 )
 
+// Delete stops a Task using the specified Task ID. The user's active cluster (Environment) is used as a filter
```
diemonster: Would we want to specify that a Task is an ECS Task?

api/provider/aws/task_read.go
```diff
@@ -9,6 +9,8 @@ import (
 	"github.com/quintilesims/layer0/common/models"
 )
 
+// Read returns a Task based on the provided Task ID. The user's active cluster (Environment) is used as a filter
```
diemonster: Ditto re: clarification on `Task`. Is it ours or ECS' concept?

api/provider/aws/deploy_read.go
jparsons04: Should we then make a comment in the file `common/models/deploy.go` above the `struct` definition of Deploy to provide this context?

api/provider/aws/task_delete.go
jparsons04: Yes, this should be explicitly specified.

api/provider/aws/task_read.go
jparsons04: This is our concept, since we are [returning a models.Task](https://github.com/quintilesims/layer0/blob/aws-provider-deploy/api/provider/aws/task_read.go#L35). I'll make this adjustment.

api/provider/aws/task_create.go
```diff
@@ -8,6 +8,9 @@ import (
 	"github.com/quintilesims/layer0/common/models"
 )
 
+// Create runs a Task using the specified Create Task Request. The Create Task Request
+// contains the name of the Task, the Environment ID, and the Deploy ID. The ECS Task Definition
+// family and version are derived from the Deploy ID.
```
zpatrick: This isn't true. There isn't a logical mapping of deploy id -> task def family/verson

api/provider/aws/task_list.go
```diff
@@ -6,6 +6,9 @@ import (
 	"github.com/quintilesims/layer0/common/models"
 )
 
+// List retrieves a list of Task ARNs from the user's active cluster (Environment) from ECS
```
zpatrick: ditto

api/provider/aws/task_read.go
```diff
@@ -9,6 +9,8 @@ import (
 	"github.com/quintilesims/layer0/common/models"
 )
 
+// Read returns a models.Task based on the provided Task ID. The user's active cluster
```
zpatrick: ditto

api/provider/aws/task_delete.go
```diff
@@ -7,6 +7,8 @@ import (
 	"github.com/aws/aws-sdk-go/service/ecs"
 )
 
+// Delete stops an ECS Task using the specified Task ID. The user's active cluster (Environment) is used as a filter
```
zpatrick: There is no such thing as an "active cluster (Environment)". Layer0 has 0-to-many environments.

api/provider/aws/task_create.go
jparsons04: I based the last sentence on this part of the code in `api-refactor`: https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/task_create.go#L22

Deploy ID is passed into `lookupDeployNameAndVersion()` and returns a Deploy Name and Deploy Version. Unless this code flow is now different?

api/provider/aws/task_delete.go
jparsons04: Is it more accurate to say then that it retrieves a list of Task ARNs from all environments in the user's l0 instance?

api/provider/aws/task_create.go
zpatrick: sorry, I was a little cart-before-the-horse with my comment. See https://github.com/quintilesims/layer0/pull/349/files#diff-f550f783d7a4fe285b3c373cc0eef78fR15. 

At this point it's a little bit in-flight, but I think we are going to end up doing something like:
```
deployID = randomHash
familyName = req.DeployName
revision/deploy version  = determined by aws
```

api/provider/aws/task_delete.go
zpatrick: Yep! perfect

api/provider/aws/task_create.go
jparsons04: If that is the case, then I feel that the [CreateTaskRequest](https://github.com/quintilesims/layer0/blob/api-refactor/common/models/create_task_request.go) model also needs to be updated to account for extracting out a req.DeployName since it currently does not and I don't see that reflected in the PR you linked to me.

api/provider/aws/deploy_read.go
diemonster: :thinking: I don't think so, otherwise we'd be inclined to add comments to all (or most) of the models, which seems a bit heavy-handed at this point. If anyone has thoughts on that, I'm all 👂 s.

jparsons04: I think what I'm going to do here is follow what I did for Task Read() and say "Read returns a models.Deploy based on the provided Deploy ID." rather than just "returns a Deploy". That way it is clearer to the reader that I'm returning a Deploy as defined within the `models` package. I'll make some minor adjustments to the explanation as well.

diemonster:
@jparsons04 would you prefer to merge this as-is, or add the remaining entity's comment with this same PR?

jparsons04:
I'm going to add LoadBalancer, Environment, and Service. So let me take a crack at those before we merge.

api/provider/aws/environment_update.go
```diff
@@ -6,6 +6,10 @@ import (
 	"github.com/quintilesims/layer0/common/models"
 )
 
+// Update updates an ECS Cluster using the specified Update Environment Request, and
```
diemonster: [nit] I don't think we need to reiterate what is visible in the function signature unless it's not reasonably apparent. "Returns an error" is specifically what made me think of that here --- if we say that, would we need to be more explicit about _when_ it returns an error? Just seems like redundant information.

jparsons04: Stating that it returns an error is probably an overkill. I must admit that it felt weird typing that out explicitly, but I wasn't sure if I needed to for consistency's sake.

jparsons04:
I went ahead and removed the references.

api/provider/aws/environment_delete.go
```diff
@@ -9,6 +9,8 @@ import (
 	"github.com/aws/aws-sdk-go/service/ecs"
 )
 
+// Delete deletes an ECS Cluster using the specified Environment ID. The associated
+// EC2 Launch Configuration and Auto Scaling Group are also removed as part of the process.
```
zpatrick: also the security group

api/provider/aws/environment_create.go
```diff
@@ -14,8 +14,14 @@ import (
 	"github.com/quintilesims/layer0/common/models"
 )
 
-// todo: ensure envirnment name is unique
+// Create creates an ECS Cluster using the specified Create Environment Request, and
+// returns a models.Environment. The Create Environment Request contains the name of the
+// Environment, the instance type and user data for the Launch Configuration, the minimum
+// size of the Cluster's Auto Scaling Group, and the Operating System and AMI ID used
+// in the Launch Configuration. The EC2 Launch Configuration and Auto Scaling Group are
+// created before the Cluster is created.
```
zpatrick: forgetting the security group

api/provider/aws/environment_update.go
zpatrick: on that note, listing the models' fields seems overkill IMO as well

api/provider/aws/service_update.go
```diff
@@ -7,6 +7,10 @@ import (
 	"github.com/quintilesims/layer0/common/models"
 )
 
+// Update updates an ECS Service using the specified Update Service Request.
```
nehayward: [nit] Maybe to avoid repetition.
"Update the ECS Service using the specified Update Service Request"
or
"Update is used to update the ECS Service"

api/provider/aws/environment_update.go
```diff
@@ -6,6 +6,10 @@ import (
 	"github.com/quintilesims/layer0/common/models"
 )
 
+// Update updates an ECS Cluster using the specified Update Environment Request.
```
nehayward: [nit] same here.
"Update the ECS Cluster using the specified Update Environment Request."

api/provider/aws/load_balancer_create.go
```diff
@@ -13,6 +13,17 @@ import (
 	"github.com/quintilesims/layer0/common/models"
 )
 
+// Create creates an Elastic Load Balancer using the specified Create
```
nehayward: [nit] same here.
"Create an Elastic Load Balancer using the specified Create"

api/provider/aws/service_update.go
jparsons04: I will adopt your changes so that it's not so repetitive. But for reference, I modeled my sentence after golang docs that I know have this awkward repetitiveness:
https://golang.org/pkg/fmt/#Scan
https://golang.org/pkg/net/#UnixConn.Close
https://golang.org/pkg/bufio/#Reader.Read
etc.

nehayward: Okay gotcha. Thanks.

PR #366, opened 2017-09-27
274: Service provider: List

tlake:
Adds the `List` operation to the Service Provider.

- Unit tests
- Run API locally
    - confirm compilation
    - hit http://localhost:9090/api/?url=/swagger.json with a browser to confirm Swagger functionality 

**Checklist**
- [x] Unit tests
- ~System tests (if applicable)~
- ~Documentation (if applicable)~

-----

Apologies for the difficulty in reviewing this PR; the subject matter is both the current changes as well as a post-merge review of Seshi's work. You might not need to look at all of them, but in an attempt to ease this headache, the files immediately relevant to `Service Provider: List()` are these:

```
common/models/
    - service_summary.go

api/controllers/
    - service.go
    - service_test.go
    - swagger.go

api/provider/
    - service.go

api/provider/aws/
    - common.go
    - common_test.go
    - service.go
    - service_list.go
    - service_list_test.go
```

### DIFF
--- a/api/controllers/swagger.go
+++ b/api/controllers/swagger.go
@@ -284,6 +284,16 @@ func (s *SwaggerController) ServeSwaggerSpec(c *fireball.Context) (fireball.Resp
 						},
 					},
 				},
+				"get": {
+					Summary: "List all Services",
+					Tags:    []string{"Service"},
+					Responses: map[string]swagger.Response{
+						"200": {
+							Description: "An array of services",
+							Schema:      swagger.NewObjectSliceSchema("Service"),
+						},
+					},
+				},
 				"put": {
 					Summary: "Update a Service",
 					Tags:    []string{"Service"},
--- a/api/provider/aws/common.go
+++ b/api/provider/aws/common.go
@@ -122,6 +122,19 @@ func lookupDeployNameAndVersion(store tag.Store, deployID string) (string, strin
 	return deployName, deployVersion, nil
 }
 
+func lookupServiceIDFromServiceARN(store tag.Store, serviceARN string) (string, error) {
+	tags, err := store.SelectByType("service")
+	if err != nil {
+		return "", err
+	}
+
+	if tag, ok := tags.WithKey("arn").WithValue(serviceARN).First(); ok {
+		return tag.EntityID, nil
+	}
+
+	return "", errors.Newf(errors.ServiceDoesNotExist, "Failed to find service with ARN '%s'", serviceARN)
+}
+
 func getEnvironmentSGName(environmentID string) string {
 	return fmt.Sprintf("%s-env", environmentID)
 }
--- a/api/provider/aws/common.go
+++ b/api/provider/aws/common.go
@@ -122,19 +122,6 @@ func lookupDeployNameAndVersion(store tag.Store, deployID string) (string, strin
 	return deployName, deployVersion, nil
 }
 
-func lookupServiceIDFromServiceARN(store tag.Store, serviceARN string) (string, error) {
-	tags, err := store.SelectByType("service")
-	if err != nil {
-		return "", err
-	}
-
-	if tag, ok := tags.WithKey("arn").WithValue(serviceARN).First(); ok {
-		return tag.EntityID, nil
-	}
-
-	return "", errors.Newf(errors.ServiceDoesNotExist, "Failed to find service with ARN '%s'", serviceARN)
-}
-
 func getEnvironmentSGName(environmentID string) string {
 	return fmt.Sprintf("%s-env", environmentID)
 }
--- a/api/provider/aws/service_list.go
+++ b/api/provider/aws/service_list.go
@@ -1,10 +1,9 @@
 package aws
 
 import (
-	"strings"
-
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/quintilesims/layer0/common/errors"
 	"github.com/quintilesims/layer0/common/models"
 )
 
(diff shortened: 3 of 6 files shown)


api/provider/aws/common.go
```diff
@@ -69,6 +69,19 @@ func lookupDeployNameAndVersion(store tag.Store, deployID string) (string, strin
 	return deployName, deployVersion, nil
 }
 
+func lookupServiceIDFromServiceARN(store tag.Store, serviceARN string) (string, error) {
```
zpatrick: We shouldn't do this pattern for services. It is only used for tasks and deploys since they don't have a logical mapping of Layer0 ID -> AWS ID.

api/provider/aws/service_list.go
```diff
@@ -1,110 +1,81 @@
...
-
-	for {
-		output, err := s.AWS.ECS.ListServices(input)
-		if err != nil {
-			return nil, err
+func (s *ServiceProvider) listServiceARNs() ([]string, error) {
+	var serviceARNs []string
+	fn := func(output *ecs.ListServicesOutput, lastPage bool) bool {
+		for _, serviceARN := range output.ServiceArns {
+			serviceARNs = append(serviceARNs, aws.StringValue(serviceARN))
```
zpatrick: You need to filter these by layer0 instance

api/provider/aws/service_list.go
```diff
@@ -1,110 +1,81 @@
...
+	for i, serviceID := range serviceIDs {
+		models[i].ServiceID = serviceID
+
+		if tag, ok := serviceTags.WithID(serviceID).WithKey("name").First(); ok {
+			models[i].ServiceName = tag.Value
+		}
 
-	for {
-		output, err := s.AWS.ECS.ListClusters(input)
+		environmentID, err := lookupEntityEnvironmentID(s.TagStore, "service", serviceID)
```
zpatrick: I don't think we need this helper function. In each case it would be used (as far as I've seen), we already have the tags that contain the `entity_id` tag. It's way faster to iterate over the tags we already have than to burden the db with another select operation.

api/provider/aws/service_list.go
```diff
@@ -1,110 +1,81 @@
...
+		if err != nil {
+			return nil, err
 		}
-	}
 
-	return clusterArns, nil
-}
+		tag, ok := environmentTags.WithKey("name").First()
+		if !ok {
+			return nil, errors.Newf(errors.EnvironmentDoesNotExist, "Could not resolve name for environment with id '%s'", environmentID)
```
zpatrick: This does not mean the environment does not exist. We don't error when we can't find name tags.

api/controllers/swagger.go
```diff
@@ -235,6 +235,16 @@ func (s *SwaggerController) ServeSwaggerSpec(c *fireball.Context) (fireball.Resp
 				},
 			},
 			"/service": {
+				"get": {
```
jparsons04: `POST /service` for Create Service needs to be included in this entity block (based on the routes `api/controllers/service.go`)

tlake: It will be once Create Service exists

api/provider/aws/service_list.go
tlake: ✅

api/provider/aws/service_list.go
tlake: ✅

api/provider/aws/service_list.go
tlake: ✅

api/provider/aws/common.go

tlake:
@jparsons04 Re: Swagger router for Deploys: Those should be added as part of Deploy work, correct?

api/controllers/swagger.go
```diff
@@ -33,7 +33,7 @@ func (s *SwaggerController) ServeSwaggerSpec(c *fireball.Context) (fireball.Resp
 	spec := swagger.Spec{
 		SwaggerVersion: "2.0",
 		Host:           c.Request.Host,
-		Schemes:        []string{"https"},
+		Schemes:        []string{"http", "https"},
```
diemonster: [nit] I made this change as well, but i think we should let this be configureable via env vars or something... pretty big security risk to have swagger open via HTTP if we forget to remove this part

tlake: Yeah, great point 👍

✅

api/provider/aws/service_list_test.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	serviceIDs := []string{"svc_id_0", "svc_id_1"}
+	results, err := service.makeServiceSummaryModels(serviceIDs)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	assert.Len(t, results, 2)
+	assert.Equal(t, "svc_id_0", results[0].ServiceID)
```
diemonster: seems like you could could iterate over `tags` instead of all these individual `assert` calls

tlake:
That makes sense to me, and I've started a branch for just such a thing. :)

api/provider/aws/service_list.go
```diff
@@ -9,102 +9,76 @@ import (
...
-		}
+	serviceARNs, err := s.listServiceARNsForClusterNames(clusterNames)
+	if err != nil {
+		return nil, err
+	}
 
-		clusterServices[clusterArn] = serviceArns
+	serviceIDs := make([]string, len(serviceARNs))
+	for i, serviceARN := range serviceARNs {
+		fqServiceID := strings.Split(serviceARN, "/")[1]
```
zpatrick: [nit] can you throw a comment right here showing a service ARN looks like?

api/provider/aws/service_list.go
```diff
@@ -9,102 +9,76 @@ import (
...
 
-			summary := models.ServiceSummary{
-				ServiceID:       service.ServiceID,
-				ServiceName:     service.ServiceName,
-				EnvironmentID:   service.EnvironmentID,
-				EnvironmentName: service.EnvironmentName,
+func (s *ServiceProvider) listServiceARNsForClusterNames(clusterNames []string) ([]string, error) {
+	var serviceARNs []string
+	for _, clusterName := range clusterNames {
+		fn := func(output *ecs.ListServicesOutput, lastPage bool) bool {
```
zpatrick: [super nit] we don't need to re-create `fn` for each environment. Also, it may help this function read a bit better to declare `fn` outside the for loop.

api/provider/aws/service_list_test.go
zpatrick: maybe something like:
```
results, err := service.makeServiceSummaryModels(serviceIDs)
...
expected := []models.ServiceSummary{
    { ServiceID: "xx", ServiceName: "yyy", ...},
    { ... },
}

assert.Equal(t, expected, result)
```

api/provider/aws/service_read.go
```diff
@@ -65,24 +67,25 @@ func (s *ServiceProvider) Read(serviceID string) (*models.Service, error) {
...
+func (s *ServiceProvider) readService(clusterName, fqServiceID string) (*ecs.Service, error) {
 	input := &ecs.DescribeServicesInput{}
 	input.SetCluster(clusterName)
 	input.SetServices([]*string{
-		aws.String(serviceID),
+		aws.String(fqServiceID),
 	})
 
 	output, err := s.AWS.ECS.DescribeServices(input)
+
```
zpatrick: remove extra line

api/provider/aws/service_read.go
```diff
@@ -65,24 +67,25 @@ func (s *ServiceProvider) Read(serviceID string) (*models.Service, error) {
 	return model, nil
 }
 
-func (s *ServiceProvider) readService(clusterName, serviceID string) (*ecs.Service, error) {
+func (s *ServiceProvider) readService(clusterName, fqServiceID string) (*ecs.Service, error) {
```
zpatrick: I get the rationale for using `fqServiceID` here, but that is inconsistent with what we have already implemented - which is to use ecs/aws verbiage for helper functions that wrap aws api calls. In this case, the ecs verbiage is `serviceID`.

If we want to change that, it needs to be an all-or-nothing thing for consistencies sake.

api/provider/aws/service_list.go
```diff
@@ -9,102 +9,76 @@ import (
...
-			s.populateSummariesTags(service, ecsServiceID)
+	return s.makeServiceSummaryModels(serviceIDs)
+}
 
-			summary := models.ServiceSummary{
-				ServiceID:       service.ServiceID,
-				ServiceName:     service.ServiceName,
-				EnvironmentID:   service.EnvironmentID,
-				EnvironmentName: service.EnvironmentName,
+func (s *ServiceProvider) listServiceARNsForClusterNames(clusterNames []string) ([]string, error) {
```
zpatrick: [nit] the helper `listClusterTaskARNs` already exists, maybe rename this to `listClusterServiceARNs`

api/provider/aws/service_list_test.go
tlake: Yeah, this is a nice pattern. :+1: 

:white_check_mark:

api/provider/aws/service_list.go
tlake: :white_check_mark:

api/provider/aws/service_list.go
tlake: Good point. :white_check_mark:

api/provider/aws/service_list.go
tlake: :white_check_mark:

api/provider/aws/service_list.go
```diff
@@ -9,102 +9,79 @@ import (
...
+		return nil, err
+	}
 
-	for _, clusterArn := range clusterArns {
-		clusterName := strings.Split(aws.StringValue(clusterArn), ":")[5]
-		clusterName = strings.Replace(clusterName, "cluster/", "", -1)
+	serviceIDs := make([]string, len(serviceARNs))
+	for i, serviceARN := range serviceARNs {
+		// sample service ARN:
+		// arn:aws:ecs:us-west-2:856306994068:task-definition/l0-tlakedev-guestbook-dpl:84
```
zpatrick: that's not a service arn

tlake: Yep, correct. Silly me. :white_check_mark:

api/provider/aws/service_list.go
```diff
@@ -12,102 +12,79 @@ import (
...
+	serviceARNs, err := s.listClusterServiceARNs(clusterNames)
+	if err != nil {
+		return nil, err
+	}
 
-	for _, clusterArn := range clusterArns {
-		clusterName := strings.Split(aws.StringValue(clusterArn), ":")[5]
-		clusterName = strings.Replace(clusterName, "cluster/", "", -1)
+	serviceIDs := make([]string, len(serviceARNs))
+	for i, serviceARN := range serviceARNs {
```
zpatrick: Doing this logic within `listClusterServiceARNs` and changing that to `listClusterServiceNames` seems to be more inline with other entity.Lists (e.g. environment's `listClusterNames`)

tlake: From what I can tell in looking at `environment_list.go` and `load_balancer_list.go`, only the one line from this `for` loop would go into `listClusterServiceNames()`. Trimming off the Layer0 prefix and creating IDs from Names seems to consistently be done here in the master function.

zpatrick: ah, good call

api/provider/aws/service_list.go
```diff
@@ -12,102 +12,78 @@ import (
...
 
-		clusterArns = append(clusterArns, output.ClusterArns...)
-		input.NextToken = output.NextToken
-
-		if output.NextToken == nil {
-			break
+		if tag, ok := environmentTags.WithKey("name").First(); ok {
+			models[i].EnvironmentName = tag.Value
 		}
 	}
```
zpatrick: you need to populate the load balancer id and name tag

api/provider/aws/service_list_test.go
```diff
@@ -0,0 +1,72 @@
...
+			Value:      "svc_name_1",
+		},
+		{
+			EntityID:   "svc_id_1",
+			EntityType: "service",
+			Key:        "environment_id",
+			Value:      "env_id_1",
+		},
+		{
+			EntityID:   "env_id_1",
```
zpatrick: ditto - load balancer and name tag

api/provider/aws/service_list.go
zpatrick: ignore me

api/provider/aws/service_list_test.go
zpatrick: ignore me

api/provider/aws/service_list.go
tlake: xD ✅

api/provider/aws/service_list_test.go
tlake: xD ✅

PR #369, opened 2017-09-28
Add service delete to aws package

nehayward:
This will add the service delete functionality to aws package.

I've done a smoke test by creating a service with existing l0 cli and then deleting service with updated api through running it locally and sending a command via curl

```bash
curl -X DELETE "http://localhost:9090/service/deleteb1ddd8" -H "accept: application/json" -H "content-type: application/json"
```

**Checklist**
- [x] Unit tests

** Notes **
nil


api/provider/aws/service_delete_test.go
```diff
@@ -0,0 +1,38 @@
...
+		},
+	}
+
+	for _, tag := range tags {
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	// TODO: Add delete service test.
```
jparsons04: I think here, you can make a call to the common helper function `deleteServiceTags()` and pass in `tagStore, "service", serviceID` where serviceID would be something like `serviceID := "sID"`, then make assertions after the call that the tags database has removed the entity. Once service Create() has been implemented, if Create() inserts other `models.Tags` into the tags db, you can add them to `tags` later and test to make sure that all associated `models.Tags` are removed.

jparsons04: Sorry, I mean the common function `deleteEntityTags`, not `deleteServiceTags()`

api/provider/aws/service_delete.go
```diff
@@ -1,5 +1,65 @@
...
 
-func (s *ServiceProvider) Delete(ServiceID string) error {
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/service/ecs"
+)
+
+// Delete a specified service within a cluster.
+func (s *ServiceProvider) Delete(serviceID string) error {
```
diemonster: It looks like we're missing this behavior, is it no longer needed?
https://github.com/quintilesims/layer0/blob/develop/api/backend/ecs/service_manager.go#L130

api/provider/aws/service_delete.go
```diff
@@ -1,5 +1,65 @@
...
+	input := &ecs.DeleteServiceInput{}
+	input.SetCluster(clusterID)
+	input.SetService(serviceID)
+
+	if err := input.Validate(); err != nil {
+		return fmt.Errorf("Error failed to validate: %e", err)
+	}
+
+	if _, err := s.AWS.ECS.DeleteService(input); err != nil {
+		return fmt.Errorf("Error deleting service: %e", err)
```
tlake: We have a pattern here in other `delete<Entity>` functions where we return the AWS error unless the error indicates that the entity was not found; if we're trying to delete something and we can't find it, mission accomplished anyway because it's not there one way or another.

Here are a couple of examples from `environment_delete.go`:
```
// deleteCluster()
	if _, err := e.AWS.ECS.DeleteCluster(input); err != nil {
		if err, ok := err.(awserr.Error); ok && err.Code() == "ClusterNotFoundException" {
			return nil
		}

		return err
	}
```
```
// deleteLC()
	if _, err := e.AWS.AutoScaling.DeleteLaunchConfiguration(input); err != nil {
		if err, ok := err.(awserr.Error); ok && strings.Contains(err.Message(), "Launch configuration name not found") {
			return nil
		}

		return err
	}
```

api/provider/aws/service_delete.go
```diff
@@ -1,5 +1,65 @@
...
+	input.DesiredCount = &desiredCount
+	input.SetCluster(clusterID)
+	input.SetService(serviceID)
+
+	if err := input.Validate(); err != nil {
+		return fmt.Errorf("Error failed to validate: %e", err)
+	}
+
+	if _, err := s.AWS.ECS.UpdateService(input); err != nil {
+		return fmt.Errorf("Error scaling service: %e", err)
```
tlake: Just pass the AWS error up here as well

api/provider/aws/service_delete.go
diemonster: needs to take an environmentID as a param as well

api/provider/aws/service_delete.go
```diff
@@ -1,5 +1,65 @@
...
+func (s *ServiceProvider) Delete(serviceID string) error {
+	desiredCount := int64(0)
+
+	environmentID, err := lookupEntityEnvironmentID(s.TagStore, "service", serviceID)
+	if err != nil {
+		return err
+	}
+	fqServiceID := addLayer0Prefix(s.Config.Instance(), serviceID)
+	fqClusterID := addLayer0Prefix(s.Config.Instance(), environmentID)
+
```
diemonster: somewhere in here, we need to replicate the behavior of gathering the taskArns for a given service ala https://github.com/quintilesims/layer0/blob/develop/api/backend/ecs/service_manager.go#L115

(if the intention is to simplify that behavior somehow, lmk)

api/provider/aws/service_delete.go
nehayward: I noticed pattern when looking at the code and in the current state, the error doesn't get encountered when I did my smoke testing when there isn't a service. 

This is why I decided to include the error. Is there another part of the api that checks if service exists?

api/provider/aws/service_delete.go
```diff
@@ -1,5 +1,65 @@
...
+	if err := s.scaleService(fqServiceID, fqClusterID, desiredCount); err != nil {
+		return fmt.Errorf("Failed to scale service: %e", err)
```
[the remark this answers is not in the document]
nehayward: Okay I wasn't sure if I should boiler plate it a bit, I'll change it.

api/provider/aws/service_delete.go
zpatrick: the top level function's done take the environmentID as a param

api/provider/aws/service_delete.go
nehayward: Okay I wasn't sure about that, can we discuss it. I'm not sure it's necessary looking at the aws sdk, but again I'm new to all this, so I could be wrong.

https://docs.aws.amazon.com/sdk-for-go/api/service/ecs/#ECS.DeleteService

"Deletes a specified service within a cluster. You can delete a service if you have no running tasks in it and the desired task count is zero. If the service is actively maintaining tasks, you cannot delete it, and you must update the service to a desired task count of zero. For more information, see UpdateService."

api/provider/aws/service_delete.go
```diff
@@ -1,5 +1,65 @@
...
+// Delete a specified service within a cluster.
+func (s *ServiceProvider) Delete(serviceID string) error {
+	desiredCount := int64(0)
+
+	environmentID, err := lookupEntityEnvironmentID(s.TagStore, "service", serviceID)
+	if err != nil {
+		return err
+	}
+	fqServiceID := addLayer0Prefix(s.Config.Instance(), serviceID)
+	fqClusterID := addLayer0Prefix(s.Config.Instance(), environmentID)
```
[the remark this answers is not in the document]
zpatrick: let's be clear with var names here:
```
clusterName := addLayer0Prefix(s.Config.Instance(), environmentID)
```

api/provider/aws/service_delete.go
nehayward: So keep this line https://github.com/quintilesims/layer0/blob/51fc6a03f3dffbb5ac38dc81cc94f1f271e8cc0c/api/provider/aws/service_delete.go#L13

api/provider/aws/service_delete.go
```diff
@@ -1,5 +1,65 @@
...
+	}
+
+	if err := deleteEntityTags(s.TagStore, "service", serviceID); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (s *ServiceProvider) scaleService(serviceID, clusterID string, desiredCount int64) error {
```
zpatrick: one of the "rules of thumb" is to hide ugliness caused by `aws-sdk-go` - using an `int64` where an `int` should be is a pretty common one.

api/provider/aws/service_delete_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+	"github.com/quintilesims/layer0/api/tag"
+	"github.com/quintilesims/layer0/common/models"
+)
+
+func TestService_deleteService(t *testing.T) {
+
+	tagStore := tag.NewMemoryStore()
+	serviceProvider := NewServiceProvider(nil, tagStore, nil)
+	_ = serviceProvider
```
zpatrick: ?

api/provider/aws/service_delete.go
zpatrick: does it throw an error at all?

tlake: Oh - if AWS doesn't throw an error when you tell it to delete a service that doesn't exist, then we don't need to look for such an error. Whether we do or don't have logic to check for DNE errors, I think we'd still want to bubble up any other AWS error that might appear.

api/provider/aws/service_delete.go
zpatrick: If I remember correctly, we may not need to do that anymore. I think that was implemented as a workaround for ecs's unreliable api for deletes at the time. I believe that is fixed now.

api/provider/aws/service_delete.go
nehayward: Okay so probably just return an error then, because if an error occurs, it means something else went wrong when attempting a delete, like a task running.

api/provider/aws/service_delete.go
zpatrick: yep

api/provider/aws/service_delete_test.go
nehayward: @jparsons04 thanks, I'll take a look at this.

api/provider/aws/service_delete_test.go
nehayward: This was just there so compiler didn't throw an error, I'm wondering how testing should be setup. Looking for feedback here.

api/provider/aws/service_delete.go
nehayward: So should I just cast it to a int64 then for the sdk

nehayward:
@zpatrick 

Here is the error when you try and delete an non existent task. It looks like the job catches it.

`2017/09/28 18:08:50 [ERROR] [JobWorker 2]: Failed to run job 1506640129315432054: ServerError (code='ServiceDoesNotExist') Service 'deleteb1ddd8' does not exist`

api/provider/aws/service_delete_test.go
tlake: Our current pattern in these test files is to write unit tests for the helper functions that don't make AWS calls; thus, we're not writing unit tests for the master `Delete()` function, nor the helper functions that reach out to AWS. As things stand right now, `Delete()` doesn't have any new helper functions that would need to be tested (both `scaleService()` and `deleteService()` call AWS).

api/provider/aws/service_delete.go
tlake: What we've been doing is to keep the types in the models, master functions, and non-AWS helper functions simple. Then we'd cast as needs require within the AWS helper functions as we build `input` objects.

api/provider/aws/service_delete.go
diemonster: obvs it would be better to avoid adding needless code, but without a system test or full smoke testing, this will be a hard thing to confirm as fixed

api/provider/aws/service_delete.go
```diff
@@ -1,5 +1,131 @@
...
+	return taskARNs, nil
+}
+
+func (s *ServiceProvider) stopARNTasks(clusterName, serviceID string) error {
+	service, err := s.readService(clusterName, serviceID)
+	if err != nil {
+		return err
+	}
+
+	taskARNs, err := s.getARNTasks(clusterName, serviceID, service.Deployments)
```
diemonster: part of the pattern that we've been trying to adopt is to strongly avoid chaining helper functions. In other words, call all the helpers from the primary `Delete()` function and give the results of those functions clear variable names

nehayward: Got it, I'll change that.

api/provider/aws/service_delete.go
```diff
@@ -1,5 +1,136 @@
...
+	"github.com/aws/aws-sdk-go/service/ecs"
+)
+
+// Delete a specified service within a cluster.
+func (s *ServiceProvider) Delete(serviceID string) error {
+	desiredCount := 0
+
+	environmentID, err := lookupEntityEnvironmentID(s.TagStore, "service", serviceID)
+	if err != nil {
+		if strings.Contains(err.Error(), "ServiceDoesNotExist") {
```
zpatrick: We should check by code here:
```
if err, ok := (errors.ServerError); ok && err.Code == errors.ServiceDoesNotExist { ... }
```

api/provider/aws/service_delete.go
```diff
@@ -1,5 +1,136 @@
...
+	}
+
+	if err := deleteEntityTags(s.TagStore, "service", serviceID); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (s *ServiceProvider) getARNTasks(clusterName, serviceID string,
```
zpatrick: 1 line per param when doing multi-line params

api/provider/aws/service_delete.go
```diff
@@ -1,5 +1,136 @@
...
+			return nil, err
+		}
+
+		taskARNs = append(taskARNs, tasks.TaskArns...)
+	}
+	return taskARNs, nil
+}
+
+func (s *ServiceProvider) stopARNTasks(clusterName, serviceID string, taskARNs []*string) error {
+	for i := range taskARNs {
```
zpatrick:
```
for _, taskARN := range taskARNS { 
    ...
}
```

api/provider/aws/service_delete.go
```diff
@@ -1,5 +1,136 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		taskARNs = append(taskARNs, tasks.TaskArns...)
+	}
+	return taskARNs, nil
+}
+
+func (s *ServiceProvider) stopARNTasks(clusterName, serviceID string, taskARNs []*string) error {
```
zpatrick: awkward function name, should use something like `stopServiceTasks`

api/provider/aws/service_delete.go
```diff
@@ -1,5 +1,136 @@
...
+		taskARN := taskARNs[i]
+		inputTask := &ecs.StopTaskInput{}
+		inputTask.SetCluster(clusterName)
+		inputTask.SetTask(*taskARN)
+
+		if err := inputTask.Validate(); err != nil {
+			return err
+		}
+
+		_, err := s.AWS.ECS.StopTask(inputTask)
```
zpatrick: always put errors in the smallest scope possible - which means always use in-line `if` statements when you are only checking an error returned by a function, e.g. `if _, err := a.AWS.ECS.StopTask(input); err != nil { ... `

api/provider/aws/service_delete.go
```diff
@@ -1,5 +1,136 @@
...
+
+		tasks, err := s.AWS.ECS.ListTasks(input)
+		if err != nil {
+			return nil, err
+		}
+
+		taskARNs = append(taskARNs, tasks.TaskArns...)
+
+		input.SetDesiredStatus(stopped)
+		tasks, err = s.AWS.ECS.ListTasks(input)
```
zpatrick: [nit] please stick to standard naming conventions w/ `aws-sdk-go` calls:
```
input := AWSThingInput{}
...
output, err := aws.DoThing(input)
...

things := output.Things
```

api/provider/aws/service_delete.go
```diff
@@ -1,5 +1,136 @@
...
+	}
+	return taskARNs, nil
+}
+
+func (s *ServiceProvider) stopARNTasks(clusterName, serviceID string, taskARNs []*string) error {
+	for i := range taskARNs {
+		taskARN := taskARNs[i]
+		inputTask := &ecs.StopTaskInput{}
+		inputTask.SetCluster(clusterName)
+		inputTask.SetTask(*taskARN)
```
zpatrick: always use safe de-reference helpers from `aws-sdk-go`; for strings, this is `aws.StringValue(taskARN)`.
Its fine to make this conversion a little bit ugly for aws calls IMO:
```
input.SetTask(aws.StringValue(taskARN))
```

api/provider/aws/service_delete.go
```diff
@@ -1,5 +1,136 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (s *ServiceProvider) scaleService(clusterName, serviceID string, desiredCount int) error {
+	input := &ecs.UpdateServiceInput{}
+	awsDesiredCount := int64(desiredCount)
```
zpatrick: this is awkward, can just be a one-liner: `input.SetDesiredCount(int64(desiredCount))`

api/provider/aws/service_delete.go
```diff
@@ -1,5 +1,136 @@
...
+	input.SetCluster(clusterName)
+	input.SetService(serviceID)
+
+	if err := input.Validate(); err != nil {
+		return err
+	}
+
+	if _, err := s.AWS.ECS.UpdateService(input); err != nil {
+		return err
+	}
```
zpatrick: [nit] add newline after closing bracket

api/provider/aws/service_delete.go
```diff
@@ -1,5 +1,136 @@
...
+func (s *ServiceProvider) deleteService(clusterName, serviceID string) error {
+	input := &ecs.DeleteServiceInput{}
+	input.SetCluster(clusterName)
+	input.SetService(serviceID)
+
+	if err := input.Validate(); err != nil {
+		return err
+	}
+
+	if _, err := s.AWS.ECS.DeleteService(input); err != nil {
```
zpatrick: this needs to check for the `ServiceDoesNotExist` error from `aws-sdk-go`

api/provider/aws/service_delete.go
zpatrick: Remove this function, move [listClusterTaskARNs](https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/task_list.go#L29) to `common.go`. Update it to take a `status string` argument, where the callers should use the string constants `ecs.DesiredStatusRunning` or `ecs.DesiredStatusStopped` instead of just a string like  `"STOPPED"`. The reason for this new param is that we don't want to list tasks that are already stopped in this scenario.

api/provider/aws/service_delete.go
```diff
@@ -1,5 +1,130 @@
...
+
+	clusterName := addLayer0Prefix(s.Config.Instance(), environmentID)
+	fqServiceID := addLayer0Prefix(s.Config.Instance(), serviceID)
+
+	service, err := s.readService(clusterName, fqServiceID)
+	if err != nil {
+		return err
+	}
+
+	taskARNs, err := s.getServieTasks(clusterName, service.Deployments)
```
zpatrick: typo

zpatrick: should't we change this to `getServiceRunningTasks` because we don't need to stop tasks that are already stopped?

api/provider/aws/service_delete.go
```diff
@@ -1,5 +1,130 @@
...
+	}
+
+	return taskARNs, nil
+}
+
+func (s *ServiceProvider) stopServiceTasks(clusterName, serviceID string, taskARNs []string) error {
+	for _, taskARN := range taskARNs {
+		input := &ecs.StopTaskInput{}
+		input.SetCluster(clusterName)
+		input.SetTask(aws.StringValue(&taskARN))
```
zpatrick: this is redundant, can be:
```
input.SetTask(taskARN)
```

api/provider/aws/service_delete.go
```diff
@@ -1,5 +1,130 @@
...
+		}
+
+		taskARNs = append(taskARNs, clusterTaskARNsStopped...)
+		taskARNs = append(taskARNs, clusterTaskARNsRunning...)
+	}
+
+	return taskARNs, nil
+}
+
+func (s *ServiceProvider) stopServiceTasks(clusterName, serviceID string, taskARNs []string) error {
```
zpatrick: I don't think we use the `serviceID` param for this function

api/provider/aws/service_delete.go
```diff
@@ -1,5 +1,130 @@
...
+	input := &ecs.DeleteServiceInput{}
+	input.SetCluster(clusterName)
+	input.SetService(serviceID)
+
+	if err := input.Validate(); err != nil {
+		return err
+	}
+
+	if _, err := s.AWS.ECS.DeleteService(input); err != nil {
+		if err, ok := err.(awserr.Error); ok && err.Code() == errors.ServiceDoesNotExist.String() {
```
zpatrick: This is mixing the `aws-sdk-go` error codes and our error codes in `layer0/common/errors`. They are two very different things and shouldn't be mixed w/ each other. Any overlap between the two is purely coincidental. 

I don't know the exact error code returned by ECS, but this should look something like:
```
if err, ok := err.(awserr.Error); ok && err.Code() == "SomeECSErrorCode" {
```

api/provider/aws/service_delete_test.go
```diff
@@ -0,0 +1,43 @@
...
+		},
+	}
+
+	for _, tag := range tags {
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	for _, tag := range tags {
```
zpatrick: This doesn't simulate how the call is actually made by the client. We only call `deleteEntityTags()` once in the "real" code, and that should be reflected here.

api/provider/aws/service_delete.go
```diff
@@ -1,5 +1,130 @@
...
+	}
+
+	return nil
+}
+
+func (s *ServiceProvider) getServiceActiveTasks(clusterName string, deployments []*ecs.Deployment) ([]string, error) {
+	taskARNs := []string{}
+
+	for _, deployment := range deployments {
+		clusterTaskARNsStopped, err := listClusterTaskARNs(s.AWS.ECS, clusterName, aws.StringValue(deployment.Id), ecs.DesiredStatusStopped)
```
zpatrick: Need to remove fetching the stopped tasks

zpatrick:
@nick have you tested this manually?

api/provider/aws/service_delete.go
nehayward: Okay, this was carried over from the legacy, so I wasn't sure why it was there before.

nehayward:
@zpatrick 

Yes multiple times. The only thing I was unsure was testing listClusterTaskARNs in the task list. I tried some manually testing, but wasn't sure how to keep a task active. It may need to be tested some more.

zpatrick:
@nehayward fwiw:  testing against the api service itself can be useful. You can always run `l0-setup apply` afterwards to restore the api to its running state.

nehayward:
@zpatrick can you elaborate, I've just been run the main.go in api, and testing calls there. What's the workflow you're suggesting.

diemonster:
LGTM after the conflicts are resolved

PR #370, opened 2017-09-28
Task Logs

zpatrick:
Adds the task logs method

Swagger

**Checklist**
- [x] Unit tests


diemonster:
This covers the case of gathering logs for a task's underlying container, but we also need to figure out how we can implement gathering additional log data for the ECS API itself, and whether that should be included within `service logs` or `task logs` or some other command: http://docs.aws.amazon.com/AmazonECS/latest/developerguide/logging-using-cloudtrail.html. 

It sounds to me like the ECS API logs will require us to sort data in CloudTrail and correlate that to a task or service by `userIdentity` to make them useful.

Alternatively, we could say screw it and try to figure out a way to dump this all to Splunk ;)

diemonster:
Moving the ECS API log discussion to here https://github.com/quintilesims/layer0/issues/371

zpatrick:
@diemonster : let's try to get this and [service logs](https://github.com/quintilesims/layer0/pull/372) merged before I begin work on adding/fixing the service model to have that metadata info

api/controllers/swagger.go
```diff
@@ -316,6 +317,24 @@ func (s *SwaggerController) ServeSwaggerSpec(c *fireball.Context) (fireball.Resp
...
+					Tags:    []string{"Task"},
+					Parameters: []swagger.Parameter{
+						swagger.NewStringPathParam("id", "ID of the task to describe", true),
+						swagger.NewIntQueryParam("tail", "The number of lines from the end to return", false),
+						swagger.NewStringQueryParam("start", "The start of the time range to fetch logs (format YYYY-MM-DD HH:MM)", false),
+						swagger.NewStringQueryParam("end", "The end of the time range to fetch logs (format YYYY-MM-DD HH:MM)", false),
+					},
+					Responses: map[string]swagger.Response{
+						"200": {
+							Description: "The task's logs",
```
diemonster: [nit] should be `tasks'`

api/controllers/task.go
```diff
@@ -69,6 +82,46 @@ func (t *TaskController) GetTask(c *fireball.Context) (fireball.Response, error)
...
+		if v == "" {
+			return time.Time{}, nil
+		}
+
+		return time.Parse(TIME_LAYOUT, v)
+	}
+
+	start, err := parseTime(query.Get("start"))
+	if err != nil {
+		return nil, errors.Newf(errors.InvalidRequest, "Invalid time: start must be in format YYYY-MM-DD HH:MM")
```
diemonster: Why are we doing request verification in the `controller` layer?

api/controllers/task_test.go
```diff
@@ -111,6 +113,54 @@ func TestGetTask(t *testing.T) {
...
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockTaskProvider := mock_provider.NewMockTaskProvider(ctrl)
+	mockJobStore := mock_job.NewMockStore(ctrl)
+	controller := NewTaskController(mockTaskProvider, mockJobStore)
+
+	logFiles := []models.LogFile{
+		{
+			ContainerName: "apline",
```
diemonster: [nit] typo

api/provider/aws/logging.go
```diff
@@ -0,0 +1,165 @@
...
+	logGroupName string,
+	taskARNs []string,
+	tail int,
+	start time.Time,
+	end time.Time,
+	fn func(logStreamName string, events []*cloudwatchlogs.OutputLogEvent),
+) error {
+	taskIDMatch := generateTaskIDMap(taskARNs)
+
+	logStreams, err := describeLogStreams(cloudWatchLogsAPI, logGroupName)
```
diemonster: [nit] I don't want to be too dogmatic about the patterns we've adopted in the `provider` layer, but it seems like we could do something similar here in regards to how we call helper functions and there would be some benefit to doing so. For example, `GetLogsFromTaskARNs()` could call the various helpers rather than having the helpers call each other --- I've found this to be a lot more readable.

api/provider/aws/logging.go
```diff
@@ -0,0 +1,165 @@
...
+	"github.com/quintilesims/layer0/common/models"
+)
+
+const (
+	// DescribeLogStreams is throttled after five transactions per second.
+	// With 50 streams/transaction, 1000 gives a reasonable streams:time ratio
+	MAX_DESCRIBE_STREAMS_COUNT = 1000
+)
+
+func GetLogsFromTaskARNs(
```
diemonster: This seems like a difficult function to work with from the caller's perspective

api/provider/aws/task_logs.go
```diff
@@ -0,0 +1,17 @@
+package aws
+
+import (
+	"time"
+
+	"github.com/quintilesims/layer0/common/models"
+)
+
+func (t *TaskProvider) Logs(taskID string, tail int, start, end time.Time) ([]models.LogFile, error) {
```
diemonster: I understanding splitting the tough log logic into a separate file with `logging.go` so it can be re-used by multiple entities, but unless I'm missing something major here, I don't see tests for any of that logic. Seems unwise given how much string manipulation we're doing in that code, etc.

api/controllers/task.go
zpatrick: we do all request verification in the controller layer

api/controllers/swagger.go
zpatrick: I don't think it is actually - it's the logs belonging to a single task (as opposed to the containers' logs)

api/provider/aws/logging.go
zpatrick: I think it's about as simple as one can make it given we have all of these tail/start/end params. 
If you look at the `taskProvider.Logs()` caller, it's actually quite simple IMO.

api/controllers/task.go
diemonster: I'm not catching were we delineate request verification versus something else. For example, what's the difference between request verification here and this? https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/environment_create.go#L38

api/controllers/swagger.go
diemonster: gotcha, disregard

api/provider/aws/logging.go
diemonster: taking a step back, do we even need all of the start/tail/end params?

zpatrick: yeah, the `log` command becomes pretty unusable without at least one of them - think about doing `l0 service logs api` when your api has been up for a month

api/provider/aws/task_logs.go
zpatrick: these are wrappers for aws api calls, which should be tested in system tests

api/controllers/task.go
zpatrick: That example is an outlier due to the nature of that particular parameter.
Regardless of whatever verification is done in the `controller` layer for the environment's os, we would still need to throw an error if we couldn't match the specified os to a user_data template in the `provider` layer.

api/provider/aws/logging.go
diemonster: I'm not suggesting we don't need at least one of these params, just questioning if we need all of them. "Complex" date range filtering tasks are better done within CloudWatch itself or, in the future, Splunk. On the cli, all I would want is a `start` param to specify how far back the logs should potentially go.

api/provider/aws/task_logs.go
diemonster: In some cases, the helpers have a decent amount of logic surrounding the API calls. The "pagination" logic in `getLogEvents` could be split out, for example.  And `generateTaskIDMap` is an potentially testable function.

api/controllers/task.go
diemonster: OK, I see what you're saying re: where the error is thrown, makes sense. 👍

api/provider/aws/logging.go
zpatrick: well, it's already been implemented and it works, so i don't see why we should take it out.

api/provider/aws/task_logs.go
zpatrick: Right now, all of the existing api helper functions w/ pagination do so within the helper function itself, and they aren't being unit tested. This would break the RoT and make it really inconsistent what is unit tested (and, in my opinion, be poor design).  So far, we've been testing helpers that only touch layer0 concepts, e.g. tags. 

However, if we do decide certain api calls should be tested, we need to do it on a consistent basis. I don't think this PR is in that scope. 

In regards to `generateTaskIDMap`, I just don't think that unit test would be very useful
from the perspective of "unit tests should give you confidence in your code". It's such a small function, I don't think it needs one.

diemonster: I agree that adding more tests here would break the pattern we have in place, and is therefore not in the scope of this PR. I made a separate issue to discuss that https://github.com/quintilesims/layer0/issues/374

diemonster:
@zpatrick what has been covered by manual testing at this point? All functionality?

PR #372, opened 2017-10-03
Service logs

zpatrick:
DEPENDS ON: https://github.com/quintilesims/layer0/pull/370

Adds the service logs method

Swagger

**Checklist**
- [ ] Unit tests


api/controllers/logging.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"net/url"
+	"strconv"
+	"time"
+
+	"github.com/quintilesims/layer0/common/errors"
+)
+
+func parseLoggingQuery(query url.Values) (int, time.Time, time.Time, error) {
```
diemonster: Shouldn't we test this?

zpatrick: yep, it's covered in the [service test](https://github.com/quintilesims/layer0/blob/service-logs/api/controllers/service_test.go#L138) and [task test](https://github.com/quintilesims/layer0/blob/service-logs/api/controllers/task_test.go#L143)

PR #376, opened 2017-10-03
Deploy Provider: List only Active Task Definitions

jparsons04:
Deploy List() will now explicitly only return active Task Definitions and Task Definition Families.

Manually test Deploy methods with Swagger
Unit tests

**Checklist**
- [x] Unit tests

### DIFF
--- a/api/controllers/deploy_test.go
+++ b/api/controllers/deploy_test.go
@@ -113,6 +113,11 @@ func TestListDeploys(t *testing.T) {
 			Version:    "1",
 		},
 		{
+			DeployID:   "d1",
+			DeployName: "deploy1",
+			Version:    "2",
+		},
+		{
 			DeployID:   "d2",
 			DeployName: "deploy2",
 			Version:    "1",
--- a/api/provider/aws/deploy_list.go
+++ b/api/provider/aws/deploy_list.go
@@ -6,9 +6,9 @@ import (
 	"github.com/quintilesims/layer0/common/models"
 )
 
-// List retrieves a list of Task Definition ARNs from ECS and returns a list of
-// Deploy summaries. A Deploy summary consists of the Deploy ID, Deploy name,
-// and Version.
+// List retrieves a list of Task Definition ARNs from ECS and returns a list of Deploy Summaries.
+// A Deploy Summary consists of the Deploy ID, Deploy name, and Version. Only active
+// Task Definitions will be listed.
 func (d *DeployProvider) List() ([]models.DeploySummary, error) {
 	taskDefinitionARNs, err := d.listTaskDefinitionARNs()
 	if err != nil {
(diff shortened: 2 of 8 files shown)


api/controllers/deploy.go
```diff
@@ -70,7 +70,13 @@ func (d *DeployController) GetDeploy(c *fireball.Context) (fireball.Response, er
 }
 
 func (d *DeployController) ListDeploys(c *fireball.Context) (fireball.Response, error) {
-	summaries, err := d.DeployProvider.List()
+	var all bool
+
+	if c.PathVariables["all"] != "" {
+		all = true
```
zpatrick: Lets use https://golang.org/pkg/strconv/#ParseBool

api/controllers/deploy_test.go
```diff
@@ -120,7 +121,7 @@ func TestListDeploys(t *testing.T) {
 	}
 
 	mockDeployProvider.EXPECT().
-		List().
+		List(all).
```
zpatrick: You need to add this query param to the fireball.Context (see task logs test) and test when all=true

api/provider/aws/deploy_list.go
```diff
@@ -34,8 +36,12 @@ func (d *DeployProvider) listTaskDefinitionARNs() ([]string, error) {
 	familyPrefix := addLayer0Prefix(d.Config.Instance(), "")
 	input := &ecs.ListTaskDefinitionFamiliesInput{}
 	input.SetFamilyPrefix(familyPrefix)
-	// TODO: Revisit how Inactive and Active Task Definitions might want to be returned to the client
-	input.SetStatus(ecs.TaskDefinitionFamilyStatusActive)
+	if all {
```
zpatrick: [nit] try to void if/else blocks. Set status active and then only have the “if” block for when all=true

api/controllers/deploy_test.go
jparsons04: :white_check_mark: 

I made this adjustment and added a new test, but I've added a note in the test that the unit test can't actually test properly filtering out Active/Inactive Task Definitions and would have to be done through a system test. As it stands, we're just testing the controller's ability to handle all=false and all=true.

jparsons04:
Something I'm concerned about with this implementation is the fact that we are allowing a client to list all inactive Task Definitions and Task Definition Families, but List() currently will only return those Deploy Summaries that [have a match in the tags db](https://github.com/quintilesims/layer0/blob/api-refactor-deploy-provider-list-all-jparsons/api/provider/aws/deploy_list.go#L90).

In my specific case, I have about 11k inactive Task Definitions (this is normal per [AWS's note on this doc page](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DeregisterTaskDefinition.html)), so when I do an `all=true` query for List, the process for me can take upwards of 30 seconds, but then I get back a single Deploy that I created through Swagger.

jparsons04:
In this case, no, because we're not including the new query parameter, so the swagger controller doesn't change.

```             "/deploy": map[string]swagger.Method{
                 "get": {
                     Summary: "List all Deploys",
                     Tags:    []string{"Deploy"},
                     Responses: map[string]swagger.Response{
                         "200": {
                             Description: "An array of deploys",
                             Schema:      swagger.NewObjectSliceSchema("Deploy"), 
                         },
                     },
                 },```

jparsons04:
Likewise, the Deploy controller doesn't change because it's making the same call to `DeployProvider.List()` as it always has.

PR #378, opened 2017-10-04
Repair 'task_list.go'

tlake:
It seems that part of `task_list.go` got messed up in a merge somewhere; most apparent, `makeTaskSummaryModels()` had been overwritten with an outdated `populateSummariesFromTaskARNs()`. This is an attempt to fix it. WARNING: I'm not particularly familiar with this part of the codebase, so it's very possible that I've overlooked something; but the tests pass now, which they don't currently do in `api-refactor`.

- `cd layer0/api && go test ./...`
- Confirm if desired behavior is intact
- Make sure that everything looks right

**Checklist**
- ~[ ] Unit tests~

closes #

### DIFF
--- a/api/provider/aws/task_list.go
+++ b/api/provider/aws/task_list.go
@@ -31,7 +31,7 @@ func (t *TaskProvider) List() ([]models.TaskSummary, error) {
 		taskARNs = append(taskARNs, clusterTaskARNsRunning...)
 	}
 
-	summaries, err := t.populateSummariesFromTaskARNs(taskARNs)
+	summaries, err := t.makeTaskSummaryModels(taskARNs)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


api/provider/aws/task_list.go
```diff
@@ -17,29 +18,42 @@ func (t *TaskProvider) List() ([]models.TaskSummary, error) {
...
+			return nil, err
+		}
 	}
 
-	return summaries, nil
+	return taskARNs, nil
 }
 
-func (t *TaskProvider) populateSummariesFromTaskARNs(taskARNs []string) ([]models.TaskSummary, error) {
+func (t *TaskProvider) makeTaskSummaryModels(taskARNs []string) ([]models.TaskSummary, error) {
```
nehayward: I think this is the only line that needs to change, since listClusterTaskARNs was moved to commons and does the same thing.

PR #379, opened 2017-10-04
274: Service provider: Create

tlake:
Implements the Create operation for the Service Provider.

- Confirm that the api unit tests pass
    - `go test ./...` from `layer0/api/`
- Confirm functionality through Swagger
    - `go run main.go` from `layer0/api/`
    - Use http://localhost:9090/api/?url=/swagger.json to create a service, both with and without a loadbalancer
        - Make sure to convert the contents of a Dockerrun.aws.json file to base64: `openssl base64 -in <infile>`
- Confirm that other service operations function upon services created by this changeset (i.e. can you use Swagger to read a service you've created?)

**Checklist**
- [x] Unit tests

closes #

### DIFF
--- a/api/provider/aws/service_create.go
+++ b/api/provider/aws/service_create.go
@@ -1,7 +1,56 @@
 package aws
 
-import "github.com/quintilesims/layer0/common/models"
+import (
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/quintilesims/layer0/common/models"
+)
 
-func (s *ServiceProvider) Create(req models.CreateServiceRequest) (string, error) {
-	return "", nil
+func (s *ServiceProvider) Create(req models.CreateServiceRequest) (*models.Service, error) {
+	serviceID := generateEntityID(req.ServiceName)
+
+	fqDeployID := addLayer0Prefix(s.Config.Instance(), req.DeployID)
+	fqEnvironmentID := addLayer0Prefix(s.Config.Instance(), req.EnvironmentID)
+	fqServiceID := addLayer0Prefix(s.Config.Instance(), serviceID)
+	desiredCount := 1
+
+	var (
+		loadBalancerContainers []ecs.LoadBalancer
+		loadBalancerRole       string
+	)
+
+	if req.LoadBalancerID != "" {
+		fqLoadbalancerID := addLayer0Prefix(s.Config.Instance(), req.LoadBalancerID)
+		loadBalancerContainer, err := s.getLoadBalancerContainer(fqLoadBalancerID, fqDeployID)
+		if err != nil {
+			return nil, err
+		}
+
+		loadBalancerContainers = []ecs.LoadBalancer{loadBalancerContainer}
+		loadBalancerRole = fmt.Sprintf("%s-lb", fqLoadBalancerID)
+	}
+
+	if err := s.createService(); err != nil {
+		return nil, err
+	}
+
+	return s.Read(serviceID)
+}
+
+func (s *ServiceProvider) createService(cluster, serviceName, taskDefinition, loadBalancerRole string, loadBalancers []ecs.LoadBalancer) error {
+	input := &ecs.CreateServiceInput{}
+	input.SetCluster(cluster)
+	input.SetDesiredCount(aws.Int64(desiredCount))
+	input.SetServiceName(serviceName)
+	if len(loadbalancers) > 0 {
+		input.SetLoadBalancers
+
+	if err := input.Validate(); err != nil {
+		return err
+	}
+
+	if _, err := s.AWS.ECS.CreateService(input); err != nil {
+		return err
+	}
+
+	return nil
 }
--- a/api/provider/aws/service_create.go
+++ b/api/provider/aws/service_create.go
@@ -69,6 +69,10 @@ func (s *ServiceProvider) Create(req models.CreateServiceRequest) (*models.Servi
 		return nil, err
 	}
 
+	if err := s.createTags(serviceID, req.ServiceName, req.EnvironmentID); err != nil {
+		return nil, err
+	}
+
 	return s.Read(serviceID)
 }
 
@@ -89,6 +93,31 @@ func (s *ServiceProvider) createService(desiredCount int, cluster, serviceName, 
 
 	if _, err := s.AWS.ECS.CreateService(input); err != nil {
 		return err
+	}
+
+	return nil
+}
+
+func (s *ServiceProvider) createTags(serviceID, serviceName, environmentID string) error {
+	tags := []models.Tag{
+		{
+			EntityID:   serviceID,
+			EntityType: "service",
+			Key:        "name",
+			Value:      serviceName,
+		},
+		{
+			EntityID:   serviceID,
+			EntityType: "service",
+			Key:        "environment_id",
+			Value:      environmentID,
+		},
+	}
+
+	for _, tag := range tags {
+		if err := s.TagStore.Insert(tag); err != nil {
+			return err
+		}
 	}
 
 	return nil
--- a/api/provider/aws/service_create.go
+++ b/api/provider/aws/service_create.go
@@ -3,36 +3,28 @@ package aws
 import (
 	"fmt"
 
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ecs"
 	"github.com/quintilesims/layer0/common/models"
 )
 
 func (s *ServiceProvider) Create(req models.CreateServiceRequest) (*models.Service, error) {
-	var (
-		cluster          string
-		desiredCount     int
-		loadBalancer     *ecs.LoadBalancer
-		loadBalancerRole string
-		serviceName      string
-		taskDefinition   string
-	)
-
 	fqEnvironmentID := addLayer0Prefix(s.Config.Instance(), req.EnvironmentID)
-	cluster = fqEnvironmentID
+	cluster := fqEnvironmentID
 
-	desiredCount = 1
+	desiredCount := 1
 
 	serviceID := generateEntityID(req.ServiceName)
 	fqServiceID := addLayer0Prefix(s.Config.Instance(), serviceID)
-	serviceName = fqServiceID
+	serviceName := fqServiceID
 
-	taskDefinitionARN, err := s.lookupTaskDefinitionARNFromDeployID(req.DeployID)
+	taskDefinitionARN, err := lookupTaskDefinitionARNFromDeployID(s.TagStore, req.DeployID)
 	if err != nil {
 		return nil, err
 	}
 
-	taskDefinition = taskDefinitionARN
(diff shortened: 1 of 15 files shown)


api/provider/aws/service_create.go
```diff
@@ -1,8 +1,137 @@
...
-import "github.com/quintilesims/layer0/common/models"
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/quintilesims/layer0/common/models"
+)
 
 func (s *ServiceProvider) Create(req models.CreateServiceRequest) (*models.Service, error) {
+	var (
```
zpatrick: this block is redundant. Also it's a pattern for C, not Go

api/provider/aws/service_create.go
```diff
@@ -1,8 +1,137 @@
...
+	serviceID := generateEntityID(req.ServiceName)
+	fqServiceID := addLayer0Prefix(s.Config.Instance(), serviceID)
+	serviceName = fqServiceID
+
+	taskDefinitionARN, err := s.lookupTaskDefinitionARNFromDeployID(req.DeployID)
+	if err != nil {
+		return nil, err
+	}
+
+	taskDefinition = taskDefinitionARN
```
zpatrick: `taskDefinitionARN` is more exact - why rename/replace here?

api/provider/aws/service_create.go
```diff
@@ -1,8 +1,137 @@
...
+
+	if req.LoadBalancerID != "" {
+		fqLoadBalancerID := addLayer0Prefix(s.Config.Instance(), req.LoadBalancerID)
+		loadBalancerDescription, err := describeLoadBalancer(s.AWS.ELB, fqLoadBalancerID)
+		if err != nil {
+			return nil, err
+		}
+
+		taskDefinitionDescription, err := describeTaskDefinition(s.AWS.ECS, taskDefinitionARN)
```
zpatrick: _this_ is a `taskDefinition`

api/provider/aws/service_create.go
```diff
@@ -1,8 +1,137 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		var loadBalancerContainer ecs.LoadBalancer
+		for _, containerDefinition := range taskDefinitionDescription.ContainerDefinitions {
+			for _, portMapping := range containerDefinition.PortMappings {
+				for _, listenerDescription := range loadBalancerDescription.ListenerDescriptions {
+					listener := listenerDescription.Listener
+					if *listener.InstancePort == *portMapping.ContainerPort {
```
zpatrick: need safe de-reference

api/provider/aws/service_create.go
```diff
@@ -1,8 +1,137 @@
...
+							ContainerName:    containerDefinition.Name,
+							ContainerPort:    portMapping.ContainerPort,
+							LoadBalancerName: loadBalancerDescription.LoadBalancerName,
+						}
+					}
+				}
+			}
+		}
+
+		loadBalancer = &loadBalancerContainer
```
zpatrick: ditto - weird var rename

api/provider/aws/service_create.go
```diff
@@ -1,8 +1,137 @@
...
+	return s.Read(serviceID)
+}
+
+func (s *ServiceProvider) createService(desiredCount int, cluster, serviceName, taskDefinition, loadBalancerRole string, loadBalancer *ecs.LoadBalancer) error {
+	input := &ecs.CreateServiceInput{}
+	input.SetCluster(cluster)
+	input.SetDesiredCount(int64(desiredCount))
+	input.SetServiceName(serviceName)
+	input.SetTaskDefinition(taskDefinition)
+	if loadBalancer != nil && loadBalancerRole != "" {
```
zpatrick: Should probably just check `if loadBalancer != nil` ...checking the role as well seems overkill

api/provider/aws/service_create.go
```diff
@@ -1,8 +1,137 @@
...
+	for _, tag := range tags {
+		if err := s.TagStore.Insert(tag); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (s *ServiceProvider) lookupTaskDefinitionARNFromDeployID(deployID string) (string, error) {
```
zpatrick: this can probably live right next to `lookupDeployIDFromTaskDefinitionARN`

api/provider/aws/service_create.go
```diff
@@ -1,8 +1,137 @@
...
+	}
+
+	if _, err := s.AWS.ECS.CreateService(input); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (s *ServiceProvider) createTags(serviceID, serviceName, environmentID string) error {
```
zpatrick: you need to add `load_balancer_id` tag

api/provider/aws/service_create_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+func TestService_createTags(t *testing.T) {
+	tagStore := tag.NewMemoryStore()
+	service := NewServiceProvider(nil, tagStore, nil)
+
+	if err := service.createTags("svc_id", "svc_name", "env_id"); err != nil {
+		t.Fatal(err)
+	}
+
+	expected := models.Tags{
```
zpatrick: ditto - need to add `load_balancer_id` tag

api/provider/aws/service_create_test.go
```diff
@@ -0,0 +1,64 @@
...
+		},
+		{
+			EntityID:   "svc_id",
+			EntityType: "service",
+			Key:        "environment_id",
+			Value:      "env_id",
+		},
+	}
+
+	for _, tag := range expected {
```
zpatrick: `assert.Equal(t, expected, tagStore.Tags())`

api/provider/aws/service_create.go
```diff
@@ -1,8 +1,137 @@
...
+	}
+
+	if err := s.createTags(serviceID, req.ServiceName, req.EnvironmentID); err != nil {
+		return nil, err
+	}
+
+	return s.Read(serviceID)
+}
+
+func (s *ServiceProvider) createService(desiredCount int, cluster, serviceName, taskDefinition, loadBalancerRole string, loadBalancer *ecs.LoadBalancer) error {
```
zpatrick: [nit] this order of params is kind of weird - one would expected cluster, serviceName, etc. to go first

api/provider/aws/service_create.go
tlake: Yeah, I was going back and forth on visibility of import vars vs not-redundancy. Didn't know it was a C pattern, so TIL.

✅

api/provider/aws/service_create.go
tlake: Just an attempt to make variable names line up between the master function and the helper function. I'm happy to not rename. ✅

api/provider/aws/service_create.go
```diff
@@ -1,8 +1,137 @@
...
+
+		taskDefinitionDescription, err := describeTaskDefinition(s.AWS.ECS, taskDefinitionARN)
+		if err != nil {
+			return nil, err
+		}
+
+		var loadBalancerContainer ecs.LoadBalancer
+		for _, containerDefinition := range taskDefinitionDescription.ContainerDefinitions {
+			for _, portMapping := range containerDefinition.PortMappings {
+				for _, listenerDescription := range loadBalancerDescription.ListenerDescriptions {
```
diemonster: is it possible to break up this nesting of `for` loops?

api/provider/aws/service_create.go
diemonster: Possibly a red herring, but I've seen some occasions where a loadbalancer isn't always in sync with its' role --- sort of 'eventual consistency' behavior, particularly after a `delete`.

api/provider/aws/service_create.go

api/provider/aws/service_create.go
tlake: @zpatrick should we prefer
```cluster, servcieName, taskDefinition string, desiredCount int, loadBalancerRole string, loadBalancer *ecs.LoadBalancer
```
where `string` is specified twice,
```
cluster, serviceName, taskDefinition, loadBalancerRole string, desiredCount int, loadBalancer *ecs.LoadBalancer
```
where the `loadBalancer` and `loadBalancerRole` concepts are separated, or
```
cluster, serviceName, taskDefinition, loadBalancerRole string, loadBalancer *ecs.LoadBalancer, desiredCount int
```
where we don't maintain an increasing amount of complexity in parameters as we read left to right?

api/provider/aws/service_create.go
tlake: Ack, right

✅

api/provider/aws/service_create_test.go
tlake: ✅

api/provider/aws/service_create.go
zpatrick: I think the first reads the best

api/provider/aws/service_create.go
tlake: I'm actually not sure we can (I mean, we _can_, but I think it will be way more of a hassle). I had started this comment thinking that maybe we could generate a list of load balancer ports and a list of container ports and then just find a match, but then I realized that we would need the corresponding `containerDefinition.Name`. So I guess we could generate a list of something like
```
type ContainerPort struct {
    Name *string
    Port *int64
}
```
That would save us one level of nesting, and we would still need
```
for _, container := range containerPorts {
    for _, listenerDescription := range loadBalancerDescription.ListenerDescriptions {
        . . .
    }
}
```

I'm happy to entertain alternative ideas if other folks have them, but I don't have one ready at the moment. For reference, [here](https://github.com/quintilesims/layer0/blob/1ce0a511a6566d3e972201fd7c9523042c2b2511/api/backend/ecs/service_manager.go#L234) is the relevant section of the old implementation (this `getLoadBalancer()` function is called by `GetService()` directly above it).

api/provider/aws/service_create.go
tlake: Cool, thanks :+1:

✅

api/provider/aws/service_create.go

api/provider/aws/service_create_test.go
tlake: ✅

api/provider/aws/service_create.go

api/provider/aws/service_create.go
tlake: Now using AWS methods. ✅

api/provider/aws/service_create.go
tlake: @zpatrick ✅ 

@diemonster Is that something that needs to be addressed in `create` then, if it tends to happen after a `delete`?

tlake:
Before we merge this, let's get https://github.com/quintilesims/layer0/pull/366 squared away first.

PR #380, opened 2017-10-05
Add admin endpoints to api and swagger

nehayward:
Add admin endpoints to new api-refactor.

May need to add a unit test.go

**Checklist**
- [ ] Unit tests

closes #

### DIFF
--- a/api/controllers/admin.go
+++ b/api/controllers/admin.go
@@ -1,22 +1,20 @@
 package controllers
 
 import (
-	"os"
-	"strings"
-
-	"github.com/quintilesims/layer0/api/job"
 	"github.com/quintilesims/layer0/common/config"
 	"github.com/quintilesims/layer0/common/models"
 	"github.com/zpatrick/fireball"
 )
 
 type AdminController struct {
-	JobStore job.Store
+	Config  config.APIConfig
+	Version string
 }
 
-func NewAdminController(j job.Store) *AdminController {
+func NewAdminController(c config.APIConfig, version string) *AdminController {
 	return &AdminController{
-		JobStore: j,
+		Config:  c,
+		Version: version,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


api/controllers/admin.go
```diff
@@ -0,0 +1,69 @@
...
+	"strings"
+
+	"github.com/quintilesims/layer0/api/job"
+	"github.com/quintilesims/layer0/common/config"
+	"github.com/quintilesims/layer0/common/models"
+	"github.com/zpatrick/fireball"
+)
+
+type AdminController struct {
+	JobStore job.Store
```
diemonster: If you take my advice elsewhere, add `	Config   config.APIConfig` here

api/controllers/admin.go
```diff
@@ -0,0 +1,69 @@
...
+			},
+		},
+	}
+}
+
+func (a *AdminController) GetConfig(c *fireball.Context) (fireball.Response, error) {
+	publicSubnets := []string{}
+	privateSubnets := []string{}
+
+	VPCID := os.Getenv(config.ENVVAR_AWS_VPC)
```
diemonster: I'm not if this is breaking some rules unknown to me with the controller pattern we have, but I think it would be better to inject `config` into `*AdminController` --- then you can get these values a bit easier ala `a.Config.SomeValue()`

api/controllers/admin.go
```diff
@@ -0,0 +1,69 @@
...
+		{
+			Path: "/admin/health",
+			Handlers: fireball.Handlers{
+				"GET": a.GetHealth,
+			},
+		},
+	}
+}
+
+func (a *AdminController) GetConfig(c *fireball.Context) (fireball.Response, error) {
```
diemonster: We should probably test this function

api/controllers/admin.go
zpatrick: This doesn't need a jobStore

common/models/api_config.go
```diff
@@ -1,8 +1,22 @@
...
 	VPCID          string   `json:"vpc_id"`
 	PublicSubnets  []string `json:"public_subnets"`
 	PrivateSubnets []string `json:"private_subnets"`
 }
+
+func (a APIConfig) Definition() swagger.Definition {
+	return swagger.Definition{
+		Type: "object",
+		Properties: map[string]swagger.Property{
+			"prefix":          swagger.NewStringProperty(),
```
zpatrick: We should add `Version string`

common/config/api.go
```diff
@@ -100,6 +101,7 @@ type APIConfig interface {
 	DynamoJobTable() string
 	DynamoTagTable() string
 	LogGroupName() string
+	Prefix() string
```
zpatrick: Remove this - we've standardized on `Instance()` instead of `Prefix()`

common/config/api.go
```diff
@@ -202,3 +204,7 @@ func (c *ContextAPIConfig) PrivateSubnets() []string {
 func (c *ContextAPIConfig) LogGroupName() string {
 	return c.C.String(FLAG_AWS_LOG_GROUP_NAME)
 }
+
+func (c *ContextAPIConfig) Prefix() string {
```
zpatrick: ditto - remove

common/models/api_config.go
```diff
@@ -1,8 +1,23 @@
...
+	Version        string   `json:"version"`
 	PublicSubnets  []string `json:"public_subnets"`
 	PrivateSubnets []string `json:"private_subnets"`
 }
+
+func (a APIConfig) Definition() swagger.Definition {
+	return swagger.Definition{
+		Type: "object",
+		Properties: map[string]swagger.Property{
+			"prefix":          swagger.NewStringProperty(),
```
zpatrick: ditto - use `instance` in place of `prefix`

nehayward:
Should I add testing?

PR #382, opened 2017-10-05
add max retries to aws config

zpatrick:
Sets aws max retries to 15


zpatrick:
Yes, we used the `go-decorator` pattern to do retries because the default wasn't enough. Turns out we can just use this method to achieve the same thing

PR #383, opened 2017-10-06
Client refactor

zpatrick:
Cleanup & Update the client package. 

**Checklist**
- [ ] Unit tests

closes #


client/deploy.go
```diff
@@ -1,46 +1,45 @@
...
-func (c *APIClient) CreateDeploy(name string, content []byte) (*models.Deploy, error) {
-	req := models.CreateDeployRequest{
-		DeployName: name,
-		Dockerrun:  content,
-	}
-
-	var deploy *models.Deploy
-	if err := c.Execute(c.Sling("deploy").Post("").BodyJSON(req), &deploy); err != nil {
-		return nil, err
+func (c *APIClient) CreateDeploy(req models.CreateDeployRequest) (string, error) {
```
diemonster: This is an interesting change. What's the advantage of not passing back the entity models?

client/environment_test.go
```diff
@@ -5,164 +5,133 @@ import (
...
+		InstanceSize:     "size",
+		MinClusterCount:  1,
+		UserDataTemplate: []byte("user_data"),
+		OperatingSystem:  "os",
+		AMIID:            "ami",
+	}
 
-		var req models.CreateEnvironmentRequest
-		Unmarshal(t, r, &req)
+	handler := func(w http.ResponseWriter, r *http.Request) {
```
diemonster: [nit] would we still want to call this 'handler'?

zpatrick: That matches the name of the argument we pass to `httptest.NewServer` - see https://golang.org/pkg/net/http/httptest/#NewServer

client/deploy.go
zpatrick: This reflects the change in the API where any create/update/delete returns a job id

client/environment_test.go
diemonster: 👍

sesh-kebab:
Question regarding switching to the new REST client, would you please note this in the PR with the reasoning? Pros and cons (if any)? Probably no cons as rclient looks like a pretty minimal sweet package 😉 

A comment though: If the syntax is the only advantage, I think this is rather subjective (@diemonster mentioned this earlier I believe). Personally, I find the `fmt.Sprintf` more verbose. The Fluent syntax with sling makes a lot of sense when querying (like writing LINQ queries or constructing REST requests) and I think it is subjectively better and objectively more concise. 

`sling`
```
func (c *APIClient) GetDeploy(id string) (*models.Deploy, error) {
  var deploy *models.Deploy
  if err := c.Execute(c.Sling("deploy/").Get(id), &deploy); err != nil {
    return nil, err
  }

  return deploy, nil
}
```

vs

`rclient`
```
func (c *APIClient) GetDeploy(id string) (*models.Deploy, error) {
  var deploy *models.Deploy
  path := fmt.Sprintf("deploy/%s", id)
  if err := c.Client.Get(path, &deploy); err != nil {
    return nil, err
  }

  return deploy, nil
}
```

zpatrick:
@seshi: 

### Pros and Cons of using rclient over sling
#### Cons
* The amount of work it would take to migrate from `sling` to `rclient`
* The `sling` package is battle-tested in Layer0.  The `rclient` package has only been battle-tested in d.ims.io. 
* `sling` is an arguably better name

#### Pros
* I've already done the migration work
* `rclient` has all of the features we use w/ `sling`, plus some new ones we could use. 
* `rclient` is more configurable than `sling`. The "client" object in `rclient` is composed of interfaces,  which makes it easy to add complex/custom behaviors. The "client" in `sling` is more pedantic - which forces you to make things "fit" in the `sling` paradigm (hence you have things like the [execute wrapper function](https://github.com/quintilesims/layer0/blob/develop/cli/client/api_client.go#L110)). 
* This may be personal preference, but I think "owning" as much of the stack as possible is a huge advantage. 


For like 90% of the `entity.verb` functions (e.g. `deploy.List`), like you stated, any "advantage" is just whichever syntax you prefer. I don't think there's a huge winner in either case there. 

I think `rclient` "wins" once things become complex.  As a simple example, `rclient` uses functional options to add behavior(s) to requests. 

See `service.Logs` in [sling](https://github.com/quintilesims/layer0/blob/develop/cli/client/service.go#L62) vs [rclient](https://github.com/quintilesims/layer0/blob/client-refactor/client/service.go#L52):
```
// no first-class notion of url queries in sling - need to generate the url manually
url := fmt.Sprintf("%s/logs?%s", id, query.Encode())
c.Execute(c.Sling("service/").Get(url), &logFiles)
```
vs
```
// the rclient.Query() option adds a url query to a request
path := fmt.Sprintf("/service/%s/logs", serviceID)
c.client.Get(path, &logs, rclient.Query(query))
```

Admittedly, that example only shows a small advantage.

However, the **real** complexity in this package exists in `api_client.go`. 
For each complex behavior, `rclient` is the clear winner IMO:

* Logging requests & responses: [sling](https://github.com/quintilesims/layer0/blob/develop/cli/client/api_client.go#L22) vs [rclient](https://github.com/quintilesims/layer0/blob/client-refactor/client/api_client.go#L51)
* Custom error handling: [sling](https://github.com/quintilesims/layer0/blob/develop/cli/client/api_client.go#L136) vs [rclient](https://github.com/quintilesims/layer0/blob/client-refactor/client/api_client.go#L76)
* Execute wrapper function: [sling](https://github.com/quintilesims/layer0/blob/develop/cli/client/api_client.go#L110) vs - not needed in rclient -

sesh-kebab:
@zpatrick awesome breakdown! Thanks.

I don't feel strongly either way regarding sling vs rclient. It was more of a case where I think a change like this deserves to be explicitly called out in the PR.

Although I feel like the first 1-3 & 5 points on the Pros list are somewhat moot, the comparisons with complex behavior (towards the end of your previous comment) is good.

client/interface.go
```diff
@@ -1,55 +1,47 @@
...
+	UpdateService(req models.UpdateServiceRequest) (string, error)
 
-	CreateTask(name, environmentID, deployID string, copies int, overrides []models.ContainerOverride) (string, error)
-	DeleteTask(id string) error
-	GetTask(id string) (*models.Task, error)
-	GetTaskLogs(id, start, end string, tail int) ([]*models.LogFile, error)
+	CreateTask(req models.CreateTaskRequest) (string, error)
+	DeleteTask(taskID string) (string, error)
 	ListTasks() ([]*models.TaskSummary, error)
+	ReadTask(taskID string) (*models.Service, error)
```
sesh-kebab: Should return *models.Task

zpatrick:
since we are building the cli-refactor off of these changes, I'm going to merge

PR #384, opened 2017-10-10
CLI Environment command

sesh-kebab:
Adds environment command

**Todo**
- [x] Create
- [x] Read
- [x] List
- [x] Delete
- [x] Update
- [x] Link/Unlink

**Checklist**
- [x] Unit tests

depends on #401


cli/command/deploy.go
```diff
@@ -2,7 +2,16 @@ package command
 
 import "github.com/urfave/cli"
 
-func (f *CommandFactory) Deploy() cli.Command {
+type DeployCommand struct {
+	*CommandMediator
+}
+
+func NewDeploy(m *CommandMediator) cli.Command {
```
zpatrick: We should follow best practices and have a proper NewDeployCommand constructor. Fetching the cli.Command should be a method on the struct (was deployCommand.Command(), previously).

cli/main.go
```diff
@@ -28,10 +28,10 @@ func main() {
 	app.Version = Version
 	app.Flags = config.CLIFlags()
 
-	commandFactory := command.NewCommandFactory(nil, nil, nil)
+	mediator := &command.CommandMediator{}
 	app.Commands = []cli.Command{
-		commandFactory.Deploy(),
-		commandFactory.Environment(),
+		command.NewDeploy(mediator),
+		command.NewEnvironment(mediator),
```
zpatrick: With changes stated above, would be:
```
command.NewDeployCommand(mediator).Command()
```

cli/command/environment.go
```diff
@@ -9,51 +9,59 @@ import (
 	"github.com/urfave/cli"
 )
 
-func (f *CommandFactory) Environment() cli.Command {
+type EnvironmentCommand struct {
+	*CommandMediator
```
zpatrick: Is “Mediator” the correct name for a base class?

diemonster:
I like this pattern. I don't think adding an extra struct to embed a Mediator is tedious... it's three lines of code, and it's explicit and easy to understand what's happening.

api/controllers/environment.go
```diff
@@ -88,8 +99,31 @@ func (e *EnvironmentController) UpdateEnvironment(c *fireball.Context) (fireball
 
 	if err := req.Validate(); err != nil {
 		return nil, errors.New(errors.InvalidRequest, err)
-
 	}
 
 	return createJob(e.JobStore, job.UpdateEnvironmentJob, req)
 }
+
+func (e *EnvironmentController) LinkEnvironment(c *fireball.Context) (fireball.Response, error) {
```
zpatrick: This should be handled in a separate branch/pr

cli/command/base_test.go
```diff
@@ -0,0 +1,74 @@
...
+			flagSet.Var(&slice, key, "")
+		case int:
+			flagSet.Int(key, v, "")
+		default:
+			t.Errorf("Cannot generate CLI context: unknown flag type for '%s'", key)
+		}
+	}
+
+	// add default global flags
+	flagSet.String("output", "text", "")
```
zpatrick: this should use the consts like in:
```
waitFlag := c.GlobalBool(config.FLAG_NO_WAIT)
waitTimeout := c.GlobalDuration(config.FLAG_TIMEOUT)
```

cli/command/base_test.go
```diff
@@ -0,0 +1,74 @@
...
+	tc := &TestCommandBase{
+		client:   mock_client.NewMockClient(ctrl),
+		printer:  &printer.TestPrinter{},
+		resolver: mock_resolver.NewMockResolver(ctrl),
+	}
+
+	return tc, ctrl
+}
+
+func NewJob() *models.Job {
```
zpatrick: I don't think this is necessary. Seems like this will add "magic" to tests when `"job_id"` gets thrown around it's hard to tell where it came from

cli/command/environment.go
```diff
@@ -11,5 +18,251 @@ func NewEnvironmentCommand(b *CommandBase) *EnvironmentCommand {
...
+					cli.StringFlag{
+						Name:  "os",
+						Value: "linux",
+						Usage: "specifies if the environment will run windows or linux containers",
+					},
+					cli.StringFlag{
+						Name:  "ami",
+						Usage: "specifies a custom AMI ID to use in the environment",
+					},
+					cli.BoolFlag{
```
zpatrick: this is a global flag, shouldn't be defined here

cli/command/environment.go
```diff
@@ -11,5 +18,251 @@ func NewEnvironmentCommand(b *CommandBase) *EnvironmentCommand {
...
+					},
+				},
+			},
+			{
+				Name:      "delete",
+				Usage:     "delete an environment",
+				ArgsUsage: "NAME",
+				Action:    e.delete,
+				Flags: []cli.Flag{
+					cli.BoolTFlag{
```
zpatrick: remove wait flag

cli/command/environment.go
```diff
@@ -11,5 +18,251 @@ func NewEnvironmentCommand(b *CommandBase) *EnvironmentCommand {
...
+	return e.printer.PrintEnvironmentSummaries(environmentSummaries...)
+}
+
+func (e *EnvironmentCommand) read(c *cli.Context) error {
+	args, err := extractArgs(c.Args(), "NAME")
+	if err != nil {
+		return err
+	}
+
+	environment, err := e.client.ReadEnvironment(args["NAME"])
```
zpatrick: you need to resolve multiple entity ids from `args["NAME"]`

cli/command/environment.go
```diff
@@ -11,5 +18,251 @@ func NewEnvironmentCommand(b *CommandBase) *EnvironmentCommand {
...
+	return e.printer.PrintEnvironments(environment)
+}
+
+func (e *EnvironmentCommand) update(c *cli.Context) error {
+	args, err := extractArgs(c.Args(), "NAME")
+	if err != nil {
+		return err
+	}
+
+	count, err := strconv.ParseInt(args["COUNT"], 10, 64)
```
zpatrick: a clearer converter for ints is `strconv.Atoi(args["COUNT"])`

cli/command/environment_test.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/quintilesims/layer0/client/mock_client"
+	"github.com/quintilesims/layer0/common/models"
+	"github.com/urfave/cli"
+)
+
+func TestCreateEnvironment(t *testing.T) {
+	client, _, envCommand, ctrl := initEnvCommandTest(t)
+	defer ctrl.Finish()
+
+	userData := "user_data"
```
zpatrick: we probably don't need this overly-explicit behavior outside of the `provider/aws` package

cli/command/environment_test.go
```diff
@@ -0,0 +1,145 @@
...
+		EnvironmentName:  "name",
+		InstanceSize:     "m3.large",
+		MinClusterCount:  2,
+		UserDataTemplate: []byte(userData),
+		OperatingSystem:  "linux",
+		AMIID:            "ami",
+	}
+
+	environment := &models.Environment{}
+	job := NewJob()
```
zpatrick: too much magic with `NewJob` IMO. Makes `job_id` come out of nowhere. 

This reads better and is more explicit IMO:
```
client.EXPECT().
    CreateEnvironment(req).
    Return(models.Job{JobID: "jid"}, nil)

client.EXPECT().
   ReadJob("jid").
   Return(models.Job{Result: "eid"}, nil)

client.EXPECT().
   ReadEnvironment("eid").
   Return(models.Environment{}, nil)
```

cli/command/environment_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+func TestCreateEnvironment(t *testing.T) {
+	client, _, envCommand, ctrl := initEnvCommandTest(t)
+	defer ctrl.Finish()
+
+	userData := "user_data"
+	file, close := tempFile(t, userData)
+	defer close()
+
+	createEnvRequest := models.CreateEnvironmentRequest{
```
zpatrick: [nit] been using the variable name `req` in most tests

cli/command/environment_test.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/golang/mock/gomock"
+	"github.com/quintilesims/layer0/cli/printer"
+	"github.com/quintilesims/layer0/cli/resolver/mock_resolver"
+	"github.com/quintilesims/layer0/client/mock_client"
+	"github.com/quintilesims/layer0/common/models"
+	"github.com/urfave/cli"
+)
+
+func TestCreateEnvironment(t *testing.T) {
+	client, _, envCommand, ctrl := initEnvCommandTest(t)
```
zpatrick: we don't need `initEnvCommandTest(t)` given we have that base test class. 

We should have a pattern like:
```
base, ctrl := newTestBaseCommand(t)
defer ctrl.Finish()

base.Resolver.EXPECT()...
base.Client.EXPECT()...

command := NewEnvironmentCommand(base.Base())
command.CreateEnvironment(...)
```

cli/command/environment_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+	resolver.EXPECT().
+		Resolve("environment", "name").
+		Return([]string{"id"}, nil)
+
+	client.EXPECT().
+		DeleteEnvironment("id").
+		Return(job.JobID, nil)
+
+	c := GetCLIContext(t, []string{"name"}, map[string]interface{}{"no-wait": true})
```
zpatrick: [nit] at one point we had type aliases for args and flags, and it made this function easier to read. I think we should add it back, e.g.:
```
// in something like test_helpers.go
type args []string
type flags map[string]interface{}

// then it reads better in tests
c := getCLIContext(t, args{"name"}, flags{"no-wait": true})
```

cli/command/environment_test.go
```diff
@@ -0,0 +1,145 @@
...
+		client:   mock_client.NewMockClient(ctrl),
+		printer:  &printer.TestPrinter{},
+		resolver: mock_resolver.NewMockResolver(ctrl),
+	}
+
+	baseCommand := NewEnvironmentCommand(tc.Command())
+
+	return tc.client, tc.resolver, baseCommand, ctrl
+}
+func tempFile(t *testing.T, content string) (*os.File, func()) {
```
zpatrick: should be in a more sensible file

cli/command/utils.go
```diff
@@ -16,3 +18,24 @@ func extractArgs(received []string, names ...string) (map[string]string, error)
 
 	return args, nil
 }
+
+func resolveSingleEntityID(resolver resolver.Resolver, entityType, target string) (string, error) {
```
zpatrick: we can remove this - the actual implementation is in `base.go` and already merged.

client/environment.go
```diff
@@ -52,3 +52,29 @@ func (c *APIClient) UpdateEnvironment(req models.UpdateEnvironmentRequest) (stri
 
 	return resp.JobID, nil
 }
+
+func (c *APIClient) CreateLink(sourceID, destinationID string) error {
```
zpatrick: This should either be moved out of this PR or the unit tests should be added

cli/command/base_test.go
sesh-kebab: Updated to:
```
job := &models.Job{
    JobID:  "job-id",
    Result: "entity-id",
}
```

cli/command/base_test.go
sesh-kebab: ✅

cli/command/environment_test.go
sesh-kebab: updated to 
```
job := &models.Job{
    JobID:  "job-id",
    Result: "entity-id",
}
```
✅

cli/command/utils.go
sesh-kebab: ✅

client/environment.go
sesh-kebab: agreed. ✅

api/controllers/environment.go
sesh-kebab: agreed ✅

cli/command/environment_test.go
```diff
@@ -0,0 +1,381 @@
...
+			c := GetCLIContext(t, tc.args, nil)
+
+			if err := tc.command(c); err == nil {
+				t.Fatalf("%s: error was nil!", tc.name)
+			}
+		})
+	}
+}
+
+func TestCreateEnvironment(t *testing.T) {
```
zpatrick: Seems like we're forcing a table test here. With only 2 different inputs, and the significant difference of behavior between them, seems like these should just be 2 tests. 

edit: this seems to apply to add wait vs. non-wait tests

client/environment.go
zpatrick: I assume we are going the unit test route?

cli/command/environment_test.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+}
+
+func TestEnvironmentLink_duplicateEnvironmentID(t *testing.T) {
+	_, resolver, command, ctrl := initEnvCommandTest(t)
+	defer ctrl.Finish()
+
+	resolver.EXPECT().
+		Resolve("environment", gomock.Any()).
+		Return([]string{"id1"}, nil).
```
zpatrick: :+1: clever

cli/command/base_test.go
```diff
@@ -0,0 +1,36 @@
+package command
```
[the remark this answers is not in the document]
sesh-kebab: ✅

cli/command/environment.go
```diff
@@ -13,5 +18,251 @@ func NewEnvironmentCommand(b *CommandBase) *EnvironmentCommand {
...
+	return e.printer.PrintEnvironmentSummaries(environmentSummaries...)
+}
+
+func (e *EnvironmentCommand) read(c *cli.Context) error {
+	args, err := extractArgs(c.Args(), "NAME")
+	if err != nil {
+		return err
+	}
+
+	entityIDs, err := e.resolver.Resolve("environment", args["NAME"])
```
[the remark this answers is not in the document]

cli/command/environment.go
```diff
@@ -13,5 +18,251 @@ func NewEnvironmentCommand(b *CommandBase) *EnvironmentCommand {
...
+	args, err := extractArgs(c.Args(), "NAME", "COUNT")
+	if err != nil {
+		return err
+	}
+
+	count, err := strconv.Atoi(args["COUNT"])
+	if err != nil {
+		return fmt.Errorf("'%s' is not a valid integer", args["COUNT"])
+	}
+	minClusterCount := int(count)
```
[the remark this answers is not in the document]
sesh-kebab: The previous int parser returned an int64. ✅

cli/command/environment.go
```diff
@@ -13,5 +18,251 @@ func NewEnvironmentCommand(b *CommandBase) *EnvironmentCommand {
...
+	if err != nil {
+		return err
+	}
+
+	id2, err := e.resolveSingleEntityIDHelper("environment", args["DESTINATION"])
+	if err != nil {
+		return err
+	}
+
+	if id1 == id2 {
```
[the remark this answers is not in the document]
sesh-kebab: This was somewhat explicit. Source and Destination were always a bit strange as it implies a one-way link. So I named the method scoped vars id1 and id2 to indicate that although we expose the arguments as `source` and `destination`, they are just 2 environment ids.

But I don't know if this the right place to make that distinction. Change ✅

cli/command/environment_test.go
```diff
@@ -0,0 +1,381 @@
...
+		Printer:  &printer.TestPrinter{},
+		Resolver: mock_resolver.NewMockResolver(ctrl),
+	}
+
+	envCmd := NewEnvironmentCommand(tc.Command())
+
+	return tc.Client, tc.Resolver, envCmd, ctrl
+}
+
+func tempFile(t *testing.T, content string) (*os.File, func()) {
```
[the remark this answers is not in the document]

cli/command/environment_test.go
```diff
@@ -0,0 +1,381 @@
...
+		Return([]string{"id1"}, nil).
+		Times(2)
+
+	c := GetCLIContext(t, []string{"name1", "name2"}, nil)
+	if err := command.unlink(c); err == nil {
+		t.Fatal("error was nil!")
+	}
+}
+
+func initEnvCommandTest(t *testing.T) (*mock_client.MockClient, *mock_resolver.MockResolver, *EnvironmentCommand, *gomock.Controller) {
```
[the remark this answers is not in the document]
sesh-kebab: I disagree - why repeat the same code in each test when you can encapsulate it in a single method call? The setup itself isn't relevant to the test, so minimizing the amount of setup lines seems better to me.

```
command, resolver, command, ctrl := initEnvCommandTest(t)
defer ctrl.Finish()

```
vs
```
defer ctrl.Finish()
...

// each reference would then have to be prefixed by tc.
tc.Client
tc.Resolver
```

cli/command/environment_test.go
```diff
@@ -0,0 +1,381 @@
...
+		InstanceSize:     "m3.large",
+		MinClusterCount:  2,
+		UserDataTemplate: []byte(userData),
+		OperatingSystem:  "linux",
+		AMIID:            "ami",
+	}
+
+	environment := &models.Environment{}
+	job := &models.Job{
+		JobID:  "job-id",
```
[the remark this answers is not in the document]
sesh-kebab: I don't feel strongly either way in this case as they input is mostly irrelevant but consistency is something I can get behind. Will propose `jid` in our slack channel.

cli/command/environment.go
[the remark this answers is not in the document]
zpatrick: That's a good point - I'm fine with `id1` `id2` as long as we are consistent. This is probably as good of time as any to change it. Any idea for something a little clearer?

cli/command/environment_test.go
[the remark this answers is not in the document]
zpatrick: because the point of the `BaseTestCommand` struct/pattern is encapsulation

cli/command/environment_test.go
```diff
@@ -0,0 +1,381 @@
...
+	"github.com/golang/mock/gomock"
+	"github.com/quintilesims/layer0/api/job"
+	"github.com/quintilesims/layer0/cli/printer"
+	"github.com/quintilesims/layer0/cli/resolver/mock_resolver"
+	"github.com/quintilesims/layer0/client/mock_client"
+	"github.com/quintilesims/layer0/common/models"
+	"github.com/urfave/cli"
+)
+
+func TestCreateEnvironment_userInputErrors(t *testing.T) {
```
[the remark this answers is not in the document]
sesh-kebab: I've updated the name to `TestEnvironmentCommand_userInputErrors`. I'm not trying to do more than what we had in our existing unit tests. This test isn't doing anything in addition to our existing environment command tests - just capturing multiple user input error tests as a table test.

cli/command/environment_test.go
sesh-kebab: Yes definitely agree it is forcing a table test here. I left it in to argue a good reason for it to exist as a table test though. 

All the commands that return a job have a wait/non-wait test which is essentially just the same method with 2 additional method call expectations. If someone was trying to figure out what was different between the two tests that almost look identical, they would have to scroll between the two methods and figure out how they differ.

Whereas passing in the `no-wait` flag to the test, it is immediately clear how the code differs when the `no-wait` flag is and isn't used.

Is that a convincing enough argument. I convinced myself into this.

client/environment.go
sesh-kebab: Yes. The API changes are in a separate PR #401 . Wanted to confirm whether we expect link/unlink should be returning jobs or not. ✅

cli/command/environment_test.go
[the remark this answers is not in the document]
sesh-kebab: not sure what you mean; the `BaseTestCommand` does encapsulate the mock objects but why do we need to use an instance of BaseTestCommand rather than the properties of that object directly?

cli/command/environment_test.go
zpatrick: i think this is a good candidate for a generalized helper for all wait vs. non-wait tests

sesh-kebab: @zpatrick I've added a generalized helper here. Thoughts? https://github.com/quintilesims/layer0/pull/384/files#diff-49f6506166fd008153f625177a0eaa5bR85

zpatrick: this is nice! some small things:
* a nitpick, but perhaps `testWaitHelper` is more appropriate? `testNoWaitHelper` seems to imply it's only for no-wait=true
* the inner function's middle param should be `flags Flags`. The body of this test shouldn't re-create that object, but just append to it:
```
testWaitHelper(t, func(t *testing.T, flags Flags, wait bool){
    ...
    flags["size"] = req.InstanceSize
    flags["min-count"] = req.MinClusterCount
    ...
}
```

cli/command/environment_test.go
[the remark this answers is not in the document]
zpatrick: > I don't see why returning multiple values and explicitly choosing to ignore as needed is an issue?

It's tedious to have to have to assign/ignore those variables as you work your way through writing a a unit test. 

That being said - I think consistency is more important than whatever pattern we choose. The difference it pretty small either way. I'm happy to adopt this pattern if the rest of the team is convinced it's the better option. 

To me, it seems like this is introducing a new pattern for the sake of introducing a new pattern. I don't view creating a helper function to save one line as much of a 'win',  neither do I see prefixing the mocks with `base.` as much of a 'loss'.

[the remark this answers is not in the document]
zpatrick: cc @diemonster  @tlake @jparsons04 @nehayward

client/environment.go
zpatrick: ok cool :+1: ...we should probably take that to slack

cli/command/environment_test.go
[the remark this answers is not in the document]
tlake: My $0.02: I tend to weigh readability and explicitness far heavier than economy of lines. To that end, I think that declaring the commands in each test, while a little more verbose, is a fair bit more-readily understood. I'd vote ~80% in favor of not adding these helper functions.

cli/command/environment.go
```diff
@@ -13,5 +18,250 @@ func NewEnvironmentCommand(b *CommandBase) *EnvironmentCommand {
...
+	destID, err := e.resolveSingleEntityIDHelper("environment", args["DESTINATION"])
+	if err != nil {
+		return err
+	}
+
+	if sourceID == destID {
+		return fmt.Errorf("Cannot link an environment to itself")
+	}
+
+	if err := e.client.CreateLink(sourceID, destID); err != nil {
```
zpatrick: this should be using the job model - bit since we have changes incoming there’s no reason to fix it here. Just something to remember for future

cli/command/environment_test.go
```diff
@@ -0,0 +1,312 @@
...
+import (
+	"testing"
+
+	"github.com/golang/mock/gomock"
+	"github.com/quintilesims/layer0/api/job"
+	"github.com/quintilesims/layer0/common/models"
+	"github.com/urfave/cli"
+)
+
+func TestEnvironmentCommand_userInputErrors(t *testing.T) {
```
zpatrick: From the other prs I’ve seen, looks like the pattern is one “userInputErrors” test per entity command. Also, many of these tests are missing cases (at least one for each missing arg, invalid types, etc.)

cli/command/environment_test.go
```diff
@@ -0,0 +1,312 @@
...
+		{
+			name:    "create",
+			command: command.create,
+		},
+		{
+			name:    "update",
+			command: command.update,
+		},
+		{
+			name:    "setMinCount",
```
zpatrick: This is confusing - why are setMinCount and update treated as separate commands?

cli/command/environment_test.go
```diff
@@ -0,0 +1,312 @@
...
+			if err := tc.command(c); err == nil {
+				t.Fatalf("%s: error was nil!", tc.name)
+			}
+		})
+	}
+}
+
+func TestCreateEnvironment(t *testing.T) {
+	testNoWaitCaseHelper(t, func(t *testing.T, otherFlags map[string]interface{}, wait bool) {
+		base, ctrl := newTestCommand(t)
```
zpatrick: Looks like you’re using a different wait helper than what’s in cli-refactor

cli/command/environment_test.go
```diff
@@ -0,0 +1,312 @@
...
+	})
+}
+
+func TestGetEnvironment(t *testing.T) {
+	base, ctrl := newTestCommand(t)
+	defer ctrl.Finish()
+
+	command := NewEnvironmentCommand(base.Command())
+
+	base.Resolver.EXPECT().
```
zpatrick: We should probably adjust this to ensure we call ReadEnvironment on multiple resolved ids

cli/command/environment_test.go
```diff
@@ -0,0 +1,312 @@
...
+			Status: job.Completed.String(),
+			Result: "entity_id",
+		}
+
+		base.Resolver.EXPECT().
+			Resolve("environment", "env_name").
+			Return([]string{"env_id"}, nil)
+
+		base.Client.EXPECT().
+			UpdateEnvironment(gomock.Any()).
```
zpatrick: We should check the request model here

cli/command/test_helpers.go
```diff
@@ -66,3 +68,40 @@ func getCLIContext(t *testing.T, args Args, flags Flags) *cli.Context {
...
+	}
+
+	if _, err := file.Write([]byte(content)); err != nil {
+		t.Fatal(err)
+	}
+
+	return file, func() { os.Remove(file.Name()) }
+}
+
+func testNoWaitCaseHelper(t *testing.T, FN func(*testing.T, map[string]interface{}, bool)) {
```
zpatrick: This is essentially a dup of what’s already in cli-refactor

cli/command/environment_test.go

cli/command/environment_test.go

cli/command/environment.go
```diff
@@ -13,5 +18,264 @@ func NewEnvironmentCommand(b *CommandBase) *EnvironmentCommand {
...
-	return cli.Command{}
+	return cli.Command{
+		Name:  "environment",
+		Usage: "manage layer0 environments",
+		Subcommands: []cli.Command{
+			{
+				Name:      "create",
+				Usage:     "create a new environment",
+				Action:    e.create,
+				ArgsUsage: "NAME",
```
jparsons04: [nit] We were having a discussion about the ArgsUsage `NAME` here becoming `ENVIRONMENT_NAME`. I think this is the currently agreed consensus on the discussions I saw about this.

cli/command/environment.go
```diff
@@ -13,5 +18,264 @@ func NewEnvironmentCommand(b *CommandBase) *EnvironmentCommand {
...
+					cli.StringFlag{
+						Name:  "ami",
+						Usage: "specifies a custom AMI ID to use in the environment",
+					},
+				},
+			},
+			{
+				Name:      "delete",
+				Usage:     "delete an environment",
+				ArgsUsage: "NAME",
```
jparsons04: [nit] Same as above, `ENVIRONMENT_NAME`

cli/command/environment.go
```diff
@@ -13,5 +18,264 @@ func NewEnvironmentCommand(b *CommandBase) *EnvironmentCommand {
...
+				Name:      "list",
+				Usage:     "list all environments",
+				Action:    e.list,
+				ArgsUsage: " ",
+			},
+			{
+				Name:      "read",
+				Usage:     "describe an environment",
+				Action:    e.read,
+				ArgsUsage: "NAME",
```
jparsons04: [nit] Same as above, `ENVIRONMENT_NAME`

cli/command/environment.go
```diff
@@ -13,5 +18,264 @@ func NewEnvironmentCommand(b *CommandBase) *EnvironmentCommand {
...
+			},
+			{
+				Name:      "setmincount",
+				Usage:     "set the minimum instance count for an environment cluster",
+				Action:    e.update,
+				ArgsUsage: "NAME COUNT",
```
jparsons04: [nit] Same as above, `ENVIRONMENT_NAME`

cli/command/environment.go
```diff
@@ -13,5 +18,264 @@ func NewEnvironmentCommand(b *CommandBase) *EnvironmentCommand {
...
+				Usage:     "unlinks two previously linked environments",
+				Action:    e.unlink,
+				ArgsUsage: "SOURCE DESTINATION",
+			},
+		},
+	}
+}
+
+func (e *EnvironmentCommand) create(c *cli.Context) error {
+	args, err := extractArgs(c.Args(), "NAME")
```
jparsons04: If you make the changes above, `NAME` must become `ENVIRONMENT_NAME`

cli/command/environment.go
```diff
@@ -13,5 +18,264 @@ func NewEnvironmentCommand(b *CommandBase) *EnvironmentCommand {
...
+		content, err := ioutil.ReadFile(path)
+		if err != nil {
+			return err
+		}
+
+		userData = content
+	}
+
+	req := models.CreateEnvironmentRequest{
+		EnvironmentName:  args["NAME"],
```
jparsons04: If you make the changes above, `NAME` must become `ENVIRONMENT_NAME`

cli/command/environment.go
```diff
@@ -13,5 +18,264 @@ func NewEnvironmentCommand(b *CommandBase) *EnvironmentCommand {
...
+	environmentSummaries, err := e.client.ListEnvironments()
+	if err != nil {
+		return err
+	}
+
+	return e.printer.PrintEnvironmentSummaries(environmentSummaries...)
+}
+
+func (e *EnvironmentCommand) read(c *cli.Context) error {
+	args, err := extractArgs(c.Args(), "NAME")
```
jparsons04: If you make the changes above, `NAME` must become `ENVIRONMENT_NAME`

cli/command/environment.go
```diff
@@ -13,5 +18,264 @@ func NewEnvironmentCommand(b *CommandBase) *EnvironmentCommand {
...
+	if err != nil {
+		return err
+	}
+
+	environmentIDs, err := e.resolver.Resolve("environment", args["NAME"])
```
jparsons04: If you make the changes above, `NAME` must become `ENVIRONMENT_NAME`

cli/command/environment.go
```diff
@@ -13,5 +18,264 @@ func NewEnvironmentCommand(b *CommandBase) *EnvironmentCommand {
...
+		}
+
+		environments[i] = environment
+	}
+
+	return e.printer.PrintEnvironments(environments...)
+}
+
+func (e *EnvironmentCommand) update(c *cli.Context) error {
+	args, err := extractArgs(c.Args(), "NAME", "COUNT")
```
jparsons04: If you make the changes above, `NAME` must become `ENVIRONMENT_NAME`

cli/command/environment.go
```diff
@@ -13,5 +18,264 @@ func NewEnvironmentCommand(b *CommandBase) *EnvironmentCommand {
...
+	if err != nil {
+		return err
+	}
+
+	minClusterCount, err := strconv.Atoi(args["COUNT"])
+	if err != nil {
+		return fmt.Errorf("'%s' is not a valid integer", args["COUNT"])
+	}
+
+	id, err := e.resolveSingleEntityIDHelper("environment", args["NAME"])
```
jparsons04: If you make the changes above, `NAME` must become `ENVIRONMENT_NAME`

cli/command/environment_test.go
```diff
@@ -0,0 +1,419 @@
...
+		"OS missing":       NewContext(t, Args{"env_name"}, nil),
+		"Count negative": NewContext(t, Args{"env_name"},
+			Flags{
+				"os":        "linux",
+				"min-count": "-1",
+			}),
+	}
+
+	for name, c := range contexts {
+		err := command.create(c)
```
zpatrick: `if err := ...`

cli/command/environment_test.go
```diff
@@ -0,0 +1,419 @@
...
+
+func TestEnvironmentCreate_userInputErrors(t *testing.T) {
+	base, ctrl := newTestCommand(t)
+	defer ctrl.Finish()
+
+	command := NewEnvironmentCommand(base.Command())
+
+	contexts := map[string]*cli.Context{
+		"Missing NAME arg": NewContext(t, nil, nil),
+		"OS missing":       NewContext(t, Args{"env_name"}, nil),
```
zpatrick: Isn't OS an optional flag with `linux` as default? That's what is was in develop

cli/command/environment_test.go
```diff
@@ -0,0 +1,419 @@
...
+		"Count negative": NewContext(t, Args{"env_name"},
+			Flags{
+				"os":        "linux",
```
zpatrick: if my comment above is true, then we don't need this for this test case

cli/command/environment_test.go
zpatrick: [nit] we've been switching to `t.Run` for table tests, e.g. 
```
    t.Run(name, func(t *testing.T){
        if err := command.create(c); err != nil { ... }
    })
}
```

cli/command/environment_test.go
```diff
@@ -0,0 +1,419 @@
...
+	defer ctrl.Finish()
+
+	command := NewEnvironmentCommand(base.Command())
+
+	contexts := map[string]*cli.Context{
+		"Missing NAME arg": NewContext(t, nil, nil),
+	}
+
+	for name, c := range contexts {
+		err := command.delete(c)
```
zpatrick: ditto `if err := ...` and the `t.Run` [nit] - there's multiple instances of this below but i'll stop commenting on each one

common/models/environment_link_request.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"fmt"
+
+	swagger "github.com/zpatrick/go-plugin-swagger"
+)
+
+type CreateEnvironmentLinkRequest EnvironmentLinkRequest
+
+type DeleteEnvironmentLinkRequest EnvironmentLinkRequest
```
zpatrick: [nit] we probably don't even need these extra types, seems reasonable to just use the `EnvironmentLinkRequest` in both create/delete scenarios

cli/command/environment_test.go
sesh-kebab: It is - but `EnvironmentCreateRequest` does validate that OS field is populated which is the reason that I've also added this as a test case.

cli/command/environment_test.go
sesh-kebab: see above.

cli/command/environment_test.go

cli/command/environment_test.go
zpatrick: hmm that seems to be a mistake w/ the api implementation. should we update that?

sesh-kebab: Do you mean not check the OS property in the EnvironmentCreateRequest.Validate()? I wouldn't say it is a mistake from the `EnvironmentCreateRequest` perspective. Seems reasonable to me that it can validate on things it deems as required.

zpatrick: yeah - if it's supposed to be optional with a default, we should not error there on validate (would also need to check if it's set in `environment.Create`)

cli/command/environment_test.go
```diff
@@ -0,0 +1,423 @@
...
+		"Count negative": NewContext(t, Args{"env_name"},
+			Flags{
+				"min-count": "-1",
+			}),
+	}
+
+	for name, c := range contexts {
+		t.Run(name, func(t *testing.T) {
+			if err := command.create(c); err == nil {
+				t.Fatalf("%s: error was nil!", name)
```
zpatrick: [nit] we don't need to print names anymore if we use `t.Run(name, ...`

PR #386, opened 2017-10-10
274: Service commands

tlake:
Adds environment command

**Todo**
- [x] Create
- [x] Delete
- [x] List
- [x] Logs
- [x] Read
- [x] Scale
- [x] Update

Manual testing, unit tests.

**Checklist**
- [x] Unit tests

**Issues**
Continues work towards #274


cli/command/service.go
```diff
@@ -10,6 +15,261 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+			{
+				Name:      "logs",
+				Usage:     "get the logs for a service",
+				Action:    s.logs,
+				ArgsUsage: "NAME",
+				Flags: []cli.Flag{
+					cli.IntFlag{
+						Name:  "tail",
+						Usage: "number of lines from the end to return",
+						Value: 0,
```
zpatrick: I don't think you need to set this if it's already 0

cli/command/service.go
```diff
@@ -10,6 +15,261 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+				ArgsUsage: "NAME",
+			},
+			{
+				Name:      "scale",
+				Usage:     "scale a service",
+				Action:    s.scale,
+				ArgsUsage: "NAME COUNT",
+				Flags: []cli.Flag{
+					cli.BoolFlag{
+						Name:  "nowait",
```
zpatrick: ditto

cli/command/service.go
```diff
@@ -10,6 +15,261 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+				},
+			},
+			{
+				Name:      "update",
+				Usage:     "run a new dploy on a service",
+				Action:    s.update,
+				ArgsUsage: "NAME DEPLOY",
+				Flags: []cli.Flag{
+					cli.BoolFlag{
+						Name:  "nowait",
```
zpatrick: ditto

cli/command/service.go
```diff
@@ -10,6 +15,261 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+		}
+
+		return s.printer.PrintServices(service)
+	}
+
+	return s.waitOnJobHelper(c, jobID, "creating", onCompleteFn)
+}
+
+func (s *ServiceCommand) delete(c *cli.Context) error {
+	deleteFn := func(serviceID string) (string, error) {
```
zpatrick: [nit] cap abbreviations `deleteFN`

cli/command/service.go
```diff
@@ -10,6 +15,261 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+	}
+
+	serviceID, err := s.resolveSingleEntityIDHelper("service", args["NAME"])
+	if err != nil {
+		return err
+	}
+
+	scale, err := strconv.Atoi(args["COUNT"])
+	if err != nil {
+		return err
```
zpatrick: we should probably wrap this error with something like `fmt.Errorf("Failed to parse COUNT argument: %v", err)`

cli/command/service_test.go
```diff
@@ -0,0 +1,322 @@
...
+		Resolve("load_balancer", "lb_name").
+		Return([]string{"lb_id"}, nil)
+
+	base.Client.EXPECT().
+		CreateService(req).
+		Return("job_id", nil)
+
+	base.Client.EXPECT().
+		ReadJob("job_id").
+		Return(&models.Job{
```
jparsons04: [nit] In [my PR](https://github.com/quintilesims/layer0/pull/394/files#diff-5de2225c2e27305b9f2011b5ab955858R35), zpatrick had me condense multi-line Returns, in this case, assign the `&models.Job` to a variable and then pass the variable here.

cli/command/service_test.go
```diff
@@ -0,0 +1,322 @@
...
+			Status: "Completed",
+			Result: "svc_id",
+		}, nil)
+
+	base.Client.EXPECT().
+		ReadService("svc_id").
+		Return(&models.Service{}, nil)
+
+	flags := map[string]interface{}{
```
jparsons04: If you don't have this yet, in test_helpers.go, we have two variables that we've created:
```
type args []string
type flags map[string]interface{}
```
So we can do some substitutions, in this case `flags := flags{"loadbalancer": "lb_name"}`

cli/command/service_test.go
```diff
@@ -0,0 +1,322 @@
...
+	c := getCLIContext(t, []string{"env_name", "svc_name", "dpl_name"}, flags)
+	if err := serviceCommand.create(c); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestCreateService_userInputErrors(t *testing.T) {
+	base, ctrl := newTestCommand(t)
+	defer ctrl.Finish()
+	serviceCommand := NewServiceCommand(base.Command())
```
jparsons04: Same as above for assginment of command just before the CLI Context assignment.

cli/command/service_test.go
```diff
@@ -0,0 +1,322 @@
...
+}
+
+func TestDeleteService(t *testing.T) {
+	base, ctrl := newTestCommand(t)
+	defer ctrl.Finish()
+	serviceCommand := NewServiceCommand(base.Command())
+
+	base.Resolver.EXPECT().
+		Resolve("service", "svc_name").
+		Return([]string{"id"}, nil)
```
jparsons04: [nit] Should this be `svc_id` to be consistent with the Job's return on line 95?

cli/command/service_test.go
```diff
@@ -0,0 +1,322 @@
...
+
+	base.Client.EXPECT().
+		DeleteService("id").
+		Return("job_id", nil)
+
+	base.Client.EXPECT().
+		ReadJob("job_id").
+		Return(&models.Job{
```
jparsons04: Same as above re: multi-line return to a variable.

cli/command/service_test.go
```diff
@@ -0,0 +1,322 @@
...
+		if err := serviceCommand.create(c); err == nil {
+			t.Fatal("%s: error was nil!", name)
+		}
+	}
+}
+
+func TestListServices(t *testing.T) {
+	base, ctrl := newTestCommand(t)
+	defer ctrl.Finish()
+	serviceCommand := NewServiceCommand(base.Command())
```
jparsons04: Same as above on assignment of command.

cli/command/service_test.go
```diff
@@ -0,0 +1,322 @@
...
+	c := getCLIContext(t, nil, nil)
+	if err := serviceCommand.list(c); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestServiceLogs(t *testing.T) {
+	base, ctrl := newTestCommand(t)
+	defer ctrl.Finish()
+	serviceCommand := NewServiceCommand(base.Command())
```
jparsons04: Same as above on assignment of command.

cli/command/service_test.go
```diff
@@ -0,0 +1,322 @@
...
+		Return([]string{"svc_id"}, nil)
+
+	base.Client.EXPECT().
+		ReadServiceLogs("svc_id", url.Values{
+			"tail":  []string{"100"},
+			"start": []string{"start"},
+			"end":   []string{"end"},
+		})
+
+	flags := map[string]interface{}{
```
jparsons04: [nit] You can use `flags{ ... }` here

cli/command/service_test.go
```diff
@@ -0,0 +1,322 @@
...
+	contexts := map[string]*cli.Context{
+		"Missing NAME arg":  getCLIContext(t, nil, nil),
+		"Missing COUNT arg": getCLIContext(t, []string{"svc_name"}, nil),
+	}
+
+	for name, c := range contexts {
+		if err := serviceCommand.scale(c); err == nil {
+			t.Fatal("%s: error was nil!", name)
+		}
+	}
```
jparsons04: I was made aware of `t.Run()` in my PR for these types of tests that unlocks some benefits when running the tests. The structure would look like:
```
		t.Run(name, func(t *testing.T) {
			if err := serviceCommand.scale(c); err == nil {
				t.Fatal("%s: error was nil!", name)
			}
		})
	}
```
You will get breakdowns of each iteration of `contexts` that will look like:
```
=== RUN   TestParsePortErrors
=== RUN   TestParsePortErrors/Non-integer_CONTAINER_PORT
=== RUN   TestParsePortErrors/Missing_HOST_PORT
=== RUN   TestParsePortErrors/Missing_CONTAINER_PORT
=== RUN   TestParsePortErrors/Missing_PROTOCOL
=== RUN   TestParsePortErrors/Non-integer_HOST_PORT
--- PASS: TestParsePortErrors (0.00s)
    --- PASS: TestParsePortErrors/Non-integer_CONTAINER_PORT (0.00s)
    --- PASS: TestParsePortErrors/Missing_HOST_PORT (0.00s)
    --- PASS: TestParsePortErrors/Missing_CONTAINER_PORT (0.00s)
    --- PASS: TestParsePortErrors/Missing_PROTOCOL (0.00s)
    --- PASS: TestParsePortErrors/Non-integer_HOST_PORT (0.00s)
```

cli/command/service_test.go
```diff
@@ -0,0 +1,322 @@
...
+
+	base.Client.EXPECT().
+		ReadService("svc_id").
+		Return(&models.Service{}, nil)
+
+	flags := map[string]interface{}{
+		"loadbalancer": "lb_name",
+	}
+
+	c := getCLIContext(t, []string{"env_name", "svc_name", "dpl_name"}, flags)
```
jparsons04: With the above comment on the variables in mind, this would now read:

`c := getCLIContext(t, args{"env_name", "svc_name", "dpl_name"}, flags)`

Or, you could just inline flags as you have only one pair:
`c := getCLIContext(t, args{"env_name", "svc_name", "dpl_name"}, flags{"loadbalancer": "lb_name"})`

Similar substitutions can be made on other `getCLIContext()` calls

cli/command/service.go
```diff
@@ -10,6 +16,260 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+		ServiceID: serviceID,
+		Scale:     &scale,
+	}
+
+	jobID, err := s.client.UpdateService(req)
+	if err != nil {
+		return err
+	}
+
+	onCompleteFN := func(serviceID string) error {
```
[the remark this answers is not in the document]
tlake: If there was, I missed it; one of the outdated comments above from @zpatrick mentioned using all-caps for abbreviations (preferring FN over what I had previously entered as Fn).

cli/command/service_test.go
tlake: @jparsons04 I may not have args and flags objects just yet, but assuming that I do: have you (& others) been preferring to declare both args and flags before the call to getCLIContext, or inline?

In other words:
```
args := args{"env_name", "svc_name", "dpl_name"}
flags := flags{"loadbalancer": "lb_name"}
c := getCLIContext(t, args, flags)
```
or
```
c := getCLIContext(t, args{"env_name", "svc_name", "dpl_name"}, flags{"loadbalancer": "lb_name"})
```

cli/command/service_test.go
tlake: Mm, yes, good catch. ✅

cli/command/service.go
```diff
@@ -10,6 +16,260 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+
+	return s.waitOnJobHelper(c, jobID, "creating", onCompleteFN)
+}
+
+func (s *ServiceCommand) delete(c *cli.Context) error {
+	deleteFN := func(serviceID string) (string, error) {
+		return s.client.DeleteService(serviceID)
+	}
+
+	return s.deleteHelper(c, "service", deleteFN)
```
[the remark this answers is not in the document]
tlake: Yeah, for sure - I had just pulled it out to get an opinion on condensation vs readability. This way seems more readable to me, and it seemed like @diemonster had a reasonably warm reception to it, but I'm happy to stuff it back in. @jparsons04 it seems like you're pro-condensation; is that accurate, or is this note more of a style/standardization-across-authors comment?

cli/command/service_test.go
```diff
@@ -0,0 +1,322 @@
...
+	"testing"
+
+	"github.com/quintilesims/layer0/common/models"
+	"github.com/urfave/cli"
+)
+
+func TestCreateService(t *testing.T) {
+	base, ctrl := newTestCommand(t)
+	defer ctrl.Finish()
+	serviceCommand := NewServiceCommand(base.Command())
```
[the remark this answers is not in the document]
tlake: That makes sense to me; I think I had gotten into this particular habit of setting up commandy things from the top from the previous iteration of these tests. Do you know if we have a team 
consensus/preference?

cli/command/service_test.go
tlake: Oh yeah, great note! I completely forgot about `t.Run()`, but now I recall that you and I had used it way back when we were poking around with benchmarking. 👍

cli/command/service_test.go
jparsons04: It looks like this is actually capitalized in that file:
```
type Args []string
type Flags map[string]interface{}
```

cli/command/service_test.go
tlake: I assume for keeping-like-with-like's sake, you would declare this variable above all of the base.yaddayadda?

cli/command/service_test.go
jparsons04: I might need more clarification on it, but I think I've been assigning things to variables if they are multi-line or would make the call to the function too long. I could conceivably make a case for assigning Args{} and Flags{} as variables before calling `getCLIContext`

cli/command/service_test.go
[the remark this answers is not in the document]
jparsons04: I'm not sure about whether a consensus exists yet. My movement of this was based on a comment in my PR.

cli/command/service.go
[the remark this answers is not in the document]
jparsons04: It is more of a standardization comment based on what was reviewed in my PR. I don't actually feel that strongly about either option in this case. I wouldn't be opposed to moving to the more readable version.

cli/command/service_test.go
jparsons04: I've been declaring them immediately above the `base.*.EXPECT()`

cli/command/service.go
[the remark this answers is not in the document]
zpatrick: yep, I'm team `FN` because go best practices uses all upper-case for abbreviations

cli/command/service.go
```diff
@@ -10,6 +16,260 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+		ServiceID: serviceID,
+		DeployID:  &deployID,
+	}
+
+	jobID, err := s.client.UpdateService(req)
+	if err != nil {
+		return err
+	}
+
+	onCompleteFN := func(serviceID string) error {
```
zpatrick: We need to add back in the `WaitForDeployment` helper function from develop and do that before returning

cli/command/service.go
```diff
@@ -10,6 +16,260 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+		ServiceName:    args["NAME"],
+	}
+
+	jobID, err := s.client.CreateService(req)
+	if err != nil {
+		return err
+	}
+
+	onCompleteFN := func(serviceID string) error {
+		service, err := s.client.ReadService(serviceID)
```
zpatrick: We need to add back in the `WaitForDeployment` helper function from develop and do that here. 
I think it should be another non-interface helper like [WaitForJob](https://github.com/quintilesims/layer0/blob/cli-refactor/client/helpers.go#L33)

cli/command/service.go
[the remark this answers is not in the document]
tlake: ✅

cli/command/service.go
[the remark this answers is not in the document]
tlake: ✅

tlake: ✅

cli/command/service_test.go
tlake: ✅

cli/command/service_test.go
tlake: Tests have been updated to define `Args{}` and `Flags{}` before passing them into the command functions. ✅

cli/command/service_test.go
tlake: ✅

cli/command/service_test.go
tlake: ✅

cli/command/service_test.go
tlake: ✅

cli/command/service_test.go
tlake: ✅

cli/command/service_test.go
tlake: Tests have been updated to define `Args{}` and `Flags{}` before passing them into the command functions. ✅

cli/command/service_test.go
tlake: Tests have been updated to define `Args{}` and `Flags{}` before passing them into the command functions. ✅

cli/command/service_test.go
tlake: ✅

cli/command/service_test.go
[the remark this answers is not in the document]
tlake: Moving this declaration to where it's relevant. ✅

cli/command/service.go
tlake: ✅

cli/command/service.go
tlake: ✅

cli/command/service.go
tlake: Removed. ✅

cli/command/service.go
```diff
@@ -10,6 +15,261 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+				Name:      "create",
+				Usage:     "create a new service",
+				Action:    s.create,
+				ArgsUsage: "ENVIRONMENT NAME DEPLOY",
+				Flags: []cli.Flag{
+					cli.StringFlag{
+						Name:  "loadbalancer",
+						Usage: "attach the service to the specified load balancer",
+					},
+					cli.BoolFlag{
```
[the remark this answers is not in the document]
tlake: ✅

cli/command/service.go
```diff
@@ -10,6 +15,261 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+				},
+			},
+			{
+				Name:      "delete",
+				Usage:     "delete a service",
+				Action:    s.delete,
+				ArgsUsage: "NAME",
+				Flags: []cli.Flag{
+					cli.BoolFlag{
+						Name:  "nowait",
```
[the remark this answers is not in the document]
tlake: ✅

cli/command/service.go
tlake: ✅

cli/command/service.go
tlake: ✅

cli/command/service.go
tlake: ✅

cli/command/service.go
```diff
@@ -10,6 +17,260 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+		ServiceName:    args["NAME"],
+	}
+
+	jobID, err := s.client.CreateService(req)
+	if err != nil {
+		return err
+	}
+
+	onCompleteFN := func(serviceID string) error {
+		service, err := WaitForDeployment(s.client, serviceID, c.GlobalDuration(config.FLAG_TIMEOUT))
```
zpatrick: :+1: perfect

cli/command/service_test.go
```diff
@@ -0,0 +1,512 @@
...
+
+	service := &models.Service{
+		Deployments:  deployments,
+		DesiredCount: 1,
+		RunningCount: 1,
+	}
+
+	base.Client.EXPECT().
+		ReadService("svc_id").
+		Times(4).
```
zpatrick: ditto on the `.Times()`

cli/command/service_test.go
```diff
@@ -0,0 +1,512 @@
...
+	}
+}
+
+func TestScaleService_userInputError(t *testing.T) {
+	base, ctrl := newTestCommand(t)
+	defer ctrl.Finish()
+
+	contexts := map[string]*cli.Context{
+		"Missing NAME arg":  NewContext(t, nil, nil),
+		"Missing COUNT arg": NewContext(t, Args{"svc_name"}, nil),
```
zpatrick: should have a case for non-integer COUNT arg

cli/command/service_test.go
```diff
@@ -0,0 +1,512 @@
...
+
+	service := &models.Service{
+		Deployments:  deployments,
+		DesiredCount: 1,
+		RunningCount: 1,
+	}
+
+	base.Client.EXPECT().
+		ReadService("svc_id").
+		Times(4).
```
zpatrick: ditto `Times()`

cli/command/utils.go
```diff
@@ -16,3 +26,62 @@ func extractArgs(received []string, names ...string) (map[string]string, error)
...
+	}
+
+	if end != "" {
+		query.Set(client.LogQueryParamEnd, end)
+	}
+
+	return query
+}
+
+func SetTimeMultiplier(v time.Duration) func() {
```
zpatrick: don't think we need a time multiplier in this package after moving the helper to `client`

cli/command/utils.go
```diff
@@ -16,3 +26,62 @@ func extractArgs(received []string, names ...string) (map[string]string, error)
...
+			}
+		}
+
+		successCount++
+		return successCount >= requiredSuccessCount, nil
+	}
+
+	delay := 5 * time.Second * timeMultiplier
+
+	for start := time.Now(); time.Since(start) < timeout; time.Sleep(delay) {
```
zpatrick: I like the sleeper pattern at https://github.com/quintilesims/layer0/blob/cli-refactor/client/helpers.go#L58. 
Perhaps we can incorporate that here? Also, I think the code can be improved a bit to avoid flapping since the current check only cares about 3 successes total, not 3 in a row:

```
var consecutiveSuccesses int
sleep := newLinearBackoffSleeper(time.Second)
for start := time.Now(); time.Since(start) < timeout; sleep() {
    service, err := client.ReadService(serviceID)
    ...

    for _, deployment := range service.Deployments {
        if deployment.DesiredCount != deployment.RunningCount {
            consecutiveSuccesses = 0
            continue
        }
    }

    consecutiveSuccesses++
    if consecutiveSuccesses >= 3 {
        return service, nil
    }
}

return nil, fmt.Errorf(...)
```

cli/command/service_test.go
```diff
@@ -0,0 +1,512 @@
...
```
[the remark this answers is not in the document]
zpatrick: Also, it may be better to use `.AnyTimes()` instead of `.Times(4)`, since we may switch-up the flapping check in the future.

cli/command/utils_test.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestWaitForDeployment(t *testing.T) {
+	defer SetTimeMultiplier(0)()
+
+	base, ctrl := newTestCommand(t)
+	defer ctrl.Finish()
+
+	deployments := []models.Deployment{
```
zpatrick: I think we can use `gomock.InOrder(...)` to make this test simulate a flapping service (see [docs](https://godoc.org/github.com/golang/mock/gomock)):
```
gomock.InOrder(
    mockObj.EXPECT().SomeMethod(1, "first"),
    mockObj.EXPECT().SomeMethod(2, "second"),
    mockObj.EXPECT().SomeMethod(3, "third"),
)
```

zpatrick: maybe like:
```
gomock.InOrder(
    // simulate deployments usually taking a while to run
    mockObj.EXPECT().Read(Desired: 1, Running: 0).Times(3),
    // simulate a flapping service
    mockObj.EXPECT().Read(Desired: 1, Running: 1),
    mockObj.EXPECT().Read(Desired: 1, Running: 0).Times(2),
    // simulate the service up and running
    mockObj.EXPECT().Read(Desired: 1, Running: 1).AnyTimes()
)
```

cli/command/utils_test.go
```diff
@@ -0,0 +1,86 @@
...
+			DesiredCount: 1,
+			RunningCount: 0,
+		},
+	}
+
+	expected := &models.Service{
+		Deployments: deployments,
+	}
+
+	base.Client.EXPECT().
```
zpatrick: shouldn't we expect this not to execute?

cli/printer/text.go
```diff
@@ -338,6 +340,7 @@ func (t *TextPrinter) PrintServiceSummaries(services ...*models.ServiceSummary)
 		rows = append(rows, row)
 	}
 
+	t.StopSpinner()
```
zpatrick: perhaps a better way to do this would be to use a single function for `TextPrinter` to actually print to stdout. It basically would just wrap `fmt.Println` but call StopSpinner() first

cli/command/service_test.go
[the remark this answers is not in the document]
tlake: ✅

cli/command/service_test.go
```diff
@@ -0,0 +1,512 @@
...
+
+func TestServiceLogs(t *testing.T) {
+	base, ctrl := newTestCommand(t)
+	defer ctrl.Finish()
+
+	base.Resolver.EXPECT().
+		Resolve("service", "svc_name").
+		Return([]string{"svc_id"}, nil)
+
+	values := url.Values{
```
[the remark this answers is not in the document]
tlake: ✅

cli/command/service_test.go
```diff
@@ -0,0 +1,512 @@
...
+		"tail":  []string{"100"},
+		"start": []string{"start"},
+		"end":   []string{"end"},
+	}
+
+	base.Client.EXPECT().
+		ReadServiceLogs("svc_id", values)
```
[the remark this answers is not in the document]
tlake: ✅

cli/command/service_test.go
tlake: ✅

cli/command/service_test.go
tlake: ✅

cli/command/service_test.go
tlake: ✅

cli/command/service_test.go
```diff
@@ -0,0 +1,512 @@
...
+		t.Run(name, func(t *testing.T) {
+			if err := serviceCommand.read(c); err == nil {
+				t.Fatal("%s: error was nil!", name)
+			}
+		})
+	}
+}
+
+func TestScaleService(t *testing.T) {
+	defer SetTimeMultiplier(0)()
```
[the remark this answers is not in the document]
tlake: ✅

cli/command/utils.go
tlake: ✅

cli/command/utils.go
```diff
@@ -16,3 +26,62 @@ func extractArgs(received []string, names ...string) (map[string]string, error)
...
+
+	return query
+}
+
+func SetTimeMultiplier(v time.Duration) func() {
+	timeMultiplier = v
+	return func() { timeMultiplier = 1 }
+}
+
+func WaitForDeployment(client client.Client, serviceID string, timeout time.Duration) (*models.Service, error) {
```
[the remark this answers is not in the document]
tlake: ✅

cli/command/service_test.go
```diff
@@ -0,0 +1,512 @@
...
+
+import (
+	"net/url"
+	"testing"
+
+	"github.com/quintilesims/layer0/common/models"
+	"github.com/urfave/cli"
+)
+
+func TestCreateService(t *testing.T) {
```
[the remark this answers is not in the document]
tlake: ✅

cli/command/service.go
```diff
@@ -10,6 +18,260 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+func (s *ServiceCommand) Command() cli.Command {
+	return cli.Command{
+		Name:  "service",
+		Usage: "manage Layer0 services",
+		Subcommands: []cli.Command{
+			{
+				Name:      "create",
+				Usage:     "create a new service",
+				Action:    s.create,
+				ArgsUsage: "ENVIRONMENT NAME DEPLOY",
```
jparsons04: [nit] We were having a discussion about the ArgsUsage `NAME` here becoming `SERVICE_NAME`. I think this is the currently agreed consensus on the discussions I saw about this.

cli/command/service.go
```diff
@@ -10,6 +18,260 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+						Name:  "loadbalancer",
+						Usage: "attach the service to the specified load balancer",
+					},
+				},
+			},
+			{
+				Name:      "delete",
+				Usage:     "delete a service",
+				Action:    s.delete,
+				ArgsUsage: "NAME",
```
jparsons04: [nit] Same as above, `SERVICE_NAME`

cli/command/service.go
```diff
@@ -10,6 +18,260 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+			},
+			{
+				Name:      "get",
+				Usage:     "describe a service",
+				Action:    s.read,
+				ArgsUsage: "NAME",
```
jparsons04: [nit] Same as above, `SERVICE_NAME`

cli/command/service.go
```diff
@@ -10,6 +18,260 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+			{
+				Name:   "list",
+				Usage:  "list all services",
+				Action: s.list,
+			},
+			{
+				Name:      "logs",
+				Usage:     "get the logs for a service",
+				Action:    s.logs,
+				ArgsUsage: "NAME",
```
jparsons04: [nit] Same as above, `SERVICE_NAME`

cli/command/service.go
```diff
@@ -10,6 +18,260 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+						Name:  "end",
+						Usage: "the end of the time range to fetch logs (format: YYYY-MM-DD HH:MM)",
+					},
+				},
+			},
+			{
+				Name:      "scale",
+				Usage:     "scale a service",
+				Action:    s.scale,
+				ArgsUsage: "NAME COUNT",
```
jparsons04: [nit] Same as above, `SERVICE_NAME`

cli/command/service.go
```diff
@@ -10,6 +18,260 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+			},
+			{
+				Name:      "update",
+				Usage:     "run a new dploy on a service",
+				Action:    s.update,
+				ArgsUsage: "NAME DEPLOY",
```
jparsons04: [nit] Same as above, `SERVICE_NAME`

cli/command/service.go
```diff
@@ -10,6 +18,260 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+			},
+		},
+	}
+}
+
+func (s *ServiceCommand) create(c *cli.Context) error {
+	args, err := extractArgs(c.Args(), "ENVIRONMENT", "NAME", "DEPLOY")
```
jparsons04: If you make the changes above, `NAME` must become `SERVICE_NAME`

cli/command/service.go
```diff
@@ -10,6 +18,260 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+		}
+
+		loadBalancerID = id
+	}
+
+	req := models.CreateServiceRequest{
+		DeployID:       deployID,
+		EnvironmentID:  environmentID,
+		LoadBalancerID: loadBalancerID,
+		ServiceName:    args["NAME"],
```
jparsons04: If you make the changes above, `NAME` must become `SERVICE_NAME`

cli/command/service.go
```diff
@@ -10,6 +18,260 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+	serviceSummaries, err := s.client.ListServices()
+	if err != nil {
+		return err
+	}
+
+	return s.printer.PrintServiceSummaries(serviceSummaries...)
+}
+
+func (s *ServiceCommand) logs(c *cli.Context) error {
+	args, err := extractArgs(c.Args(), "NAME")
```
jparsons04: If you make the changes above, `NAME` must become `SERVICE_NAME`

cli/command/service.go
```diff
@@ -10,6 +18,260 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+	if err != nil {
+		return err
+	}
+
+	serviceID, err := s.resolveSingleEntityIDHelper("service", args["NAME"])
```
jparsons04: If you make the changes above, `NAME` must become `SERVICE_NAME`

cli/command/service.go
```diff
@@ -10,6 +18,260 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+	logs, err := s.client.ReadServiceLogs(serviceID, query)
+	if err != nil {
+		return err
+	}
+
+	return s.printer.PrintLogs(logs...)
+}
+
+func (s *ServiceCommand) read(c *cli.Context) error {
+	args, err := extractArgs(c.Args(), "NAME")
```
jparsons04: If you make the changes above, `NAME` must become `SERVICE_NAME`

cli/command/service.go
```diff
@@ -10,6 +18,260 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+	if err != nil {
+		return err
+	}
+
+	serviceIDs, err := s.resolver.Resolve("service", args["NAME"])
```
jparsons04: If you make the changes above, `NAME` must become `SERVICE_NAME`

cli/command/service.go
```diff
@@ -10,6 +18,260 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+		}
+
+		services[i] = service
+	}
+
+	return s.printer.PrintServices(services...)
+}
+
+func (s *ServiceCommand) scale(c *cli.Context) error {
+	args, err := extractArgs(c.Args(), "NAME", "COUNT")
```
jparsons04: If you make the changes above, `NAME` must become `SERVICE_NAME`

cli/command/service.go
```diff
@@ -10,6 +18,260 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+	if err != nil {
+		return err
+	}
+
+	serviceID, err := s.resolveSingleEntityIDHelper("service", args["NAME"])
```
jparsons04: If you make the changes above, `NAME` must become `SERVICE_NAME`

cli/command/service.go
```diff
@@ -10,6 +18,260 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+		}
+
+		return s.printer.PrintServices(service)
+	}
+
+	return s.waitOnJobHelper(c, jobID, "scaling", onCompleteFN)
+}
+
+func (s *ServiceCommand) update(c *cli.Context) error {
+	args, err := extractArgs(c.Args(), "NAME", "DEPLOY")
```
jparsons04: If you make the changes above, `NAME` must become `SERVICE_NAME`

cli/command/service.go
```diff
@@ -10,6 +18,260 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+	if err != nil {
+		return err
+	}
+
+	serviceID, err := s.resolveSingleEntityIDHelper("service", args["NAME"])
```
jparsons04: If you make the changes above, `NAME` must become `SERVICE_NAME`

cli/printer/text.go
tlake: ✅

cli/command/service.go
tlake: Oh cool, I support this. ✅

cli/command/service.go
tlake: ✅

cli/command/service.go
tlake: ✅

cli/command/service.go
tlake: ✅

cli/command/service.go
tlake: ✅

cli/command/service.go
tlake: ✅

cli/command/service.go
tlake: ✅

cli/command/service.go
tlake: ✅

cli/command/service.go
tlake: ✅

cli/command/service.go
tlake: ✅

cli/command/service.go
tlake: ✅

cli/command/service.go
tlake: ✅

cli/command/service.go
tlake: ✅

cli/command/service.go
tlake: ✅

cli/command/service.go
tlake: ✅

cli/command/service.go
tlake: ✅

cli/command/utils_test.go
tlake: @zpatrick This got a little lost as I was moving things about; I assume you're talking about `TestWaitForDeployment_timeout()`, which now lives in `layer0/client/helpers_test.go`.

```
func TestWaitForDeployment_timeout(t *testing.T) {
	defer SetTimeMultiplier(0)()

	ctrl := gomock.NewController(t)
	defer ctrl.Finish()
	client := mock_client.NewMockClient(ctrl)

	deployments := []models.Deployment{
		{
			DesiredCount: 1,
			RunningCount: 0,
		},
	}

	expected := &models.Service{
		Deployments: deployments,
	}

	client.EXPECT().
		ReadService("svc_id").
		Return(expected, nil).
		AnyTimes()

	if _, err := WaitForDeployment(client, "svc_id", time.Second); err == nil {
		t.Fatal("Error was nil!")
	}
}
```

I expect this to execute many times; each time a `Deployment` is to be checked, `ReadService` must be called in order to get the most recent `Deployment` status, since `Deployments` are objects within `Service`s.

cli/command/utils.go
tlake: ✅

tlake: @zpatrick I've taken those things into consideration, and refactored accordingly (see `layer0/client/helpers.go:WaitForDeployment()`). However, I kept my implementation using that `check()` function instead of this suggestion with `continue`, since that `continue` repeats the inner `for` loop over `service.Deployments` instead of the outer `time` loop. This has two undesired effects:
- `Deployment`s continue to be checked for validation after a failure has been detected.
- The `consecutiveSuccess++` operation happens for every iteration of the `time` loop, not just the iterations where the `*Count`s match.

cli/command/utils_test.go
tlake: I like this a lot. ✅

cli/command/service_test.go
```diff
@@ -0,0 +1,420 @@
...
+				Result: "svc_id",
+			}
+
+			base.Client.EXPECT().
+				ReadJob("job_id").
+				Return(job, nil)
+
+			deployments := []models.Deployment{
+				{
+					DesiredCount: 1,
```
zpatrick: [nit] desired count above was 2 -- I know all that matters is that desired == running, but having 1 here instead of 2 may throw people off

client/helpers.go
```diff
@@ -30,6 +30,41 @@ const (
...
 )
 
+func SetTimeMultiplier(v time.Duration) func() {
+	timeMultiplier = v
+	return func() { timeMultiplier = 1 }
+}
+
+func WaitForDeployment(client Client, serviceID string, timeout time.Duration) (*models.Service, error) {
+	var consecutiveSuccesses int
+	check := func(service *models.Service) bool {
```
zpatrick: This fails to reset `consecutiveSuccesses` after a failure. 

As more of a [nit],  the `check` function seems unnecessary here? I feel it would read better to just have this logic in the `for` loop's body.

client/helpers.go
```diff
@@ -30,6 +30,41 @@ const (
...
+	sleep := newLinearBackoffSleeper(time.Second)
+	for start := time.Now(); time.Since(start) < timeout; sleep() {
+		service, err := client.ReadService(serviceID)
+		if err != nil {
+			return nil, err
+		}
+
+		finished := check(service)
+
+		if finished {
```
zpatrick: [nit]
```
if check(service) {
    return service, nil
}
```

client/helpers.go
```diff
@@ -30,6 +30,41 @@ const (
...
+			return service, nil
+		}
+	}
+
+	return nil, fmt.Errorf("Deployment of service '%s' has not completed within the timeout '%v'", serviceID, timeout)
```
zpatrick: [nit] wording is a little awkward - maybe `...has not completed after %v", serviceID, timeout)`.

client/helpers_test.go
```diff
@@ -11,13 +12,109 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func setTimeMultiplier(v time.Duration) func() {
-	timeMultiplier = v
-	return func() { timeMultiplier = 1 }
+func TestWaitForDeployment_standard(t *testing.T) {
```
zpatrick: :+1: good test

client/helpers_test.go
```diff
@@ -11,13 +12,109 @@ import (
...
+}
+
+func TestWaitForDeployment_error(t *testing.T) {
+	defer SetTimeMultiplier(0)()
+
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+	client := mock_client.NewMockClient(ctrl)
+
+	expected := fmt.Errorf("Error reading service")
```
zpatrick: [nit] we typically throw errors inline with `.Return(nil, fmt.Errorf("..."))` to save some space

cli/command/service_test.go
tlake: Nice catch ✅

client/helpers.go
tlake: Great point about the `consecutiveSuccesses` reset that I forgot to put in; that's fixed.

As for the `check()` function, I opted to keep it because it seemed to me to be the best way to handle looping over deployments and avoiding `consecutiveSuccesses++`. I don't know if you caught my previous comment about your suggestion of `continue`, but I still feel like it applies here (https://github.com/quintilesims/layer0/pull/386#discussion_r146074990).

client/helpers.go
tlake: ✅

client/helpers.go
tlake: ✅

client/helpers_test.go
tlake: ✅

cli/command/service.go
```diff
@@ -10,6 +18,260 @@ func NewServiceCommand(b *CommandBase) *ServiceCommand {
...
+	}
+
+	query := buildLogQueryHelper(serviceID, c.String("start"), c.String("end"), c.Int("tail"))
+
+	logs, err := s.client.ReadServiceLogs(serviceID, query)
+	if err != nil {
+		return err
+	}
+
+	return s.printer.PrintLogs(logs...)
```
diemonster: @tlake in manual testing, how long have you observed it takes for this command to return?

tlake: @diemonster I don't recall it taking a particularly long time, but I'll whack at it again and get back to you.

tlake:
```
$ time go run main.go se
rvice logs guestbod0dfc


2017/10/12 19:20:48 Using memory backend
2017/10/12 19:20:48 Listening on :80


real    0m2.781s
user    0m1.312s
sys     0m0.258s
```

So yeah, nothing strange on my test.

tlake: @diemonster
```
# from layer0/cli/
$ eval $(l0-setup endpoint -d -i tlakedev | sed -n '1!p')
$ export LAYER0_AWS_LOG_GROUP_NAME="l0-tlakedev"
$ export LAYER0_API_ENDPOINT=http://localhost:9090
$ go install .
$ time for i in {1..10}; do go run main.go service logs guestbod0dfc; done
...
real    0m43.647s
user    0m4.905s
sys     0m1.280s
```

So about 4.4 seconds, on average.

diemonster: I've raised this issue on a couple other PRs, it's mostly a UX thing, but if we have commands that tend to take a while to return, we should consider using the spinner functionality to let the user know that something is happening. Unfortunately, we don't have a good pattern for doing this (at the moment, at least) without the corresponding `client` call returning a `job`.

client/helpers.go
zpatrick: oh i didn't see that - makes sense now!

cli/command/service_test.go
```diff
@@ -0,0 +1,418 @@
...
+
+	contexts := map[string]*cli.Context{
+		"Missing NAME arg": NewContext(t, nil, nil),
+	}
+
+	serviceCommand := NewServiceCommand(base.Command())
+	for name, c := range contexts {
+		t.Run(name, func(t *testing.T) {
+			if err := serviceCommand.create(c); err == nil {
+				t.Fatal("%s: error was nil!", name)
```
zpatrick: [nit] we don't need to show the case name anymore in the error since `t.Run(name` does that for us

cli/command/test_helpers.go
```diff
@@ -17,6 +17,18 @@ type TestCommandBase struct {
 	Resolver *mock_resolver.MockResolver
 }
 
+func NewTestCommand(t *testing.T) (*TestCommandBase, *gomock.Controller) {
```
zpatrick: doesn't this exist somewhere already in the `cli-refactor` branch?

cli/command/utils.go
```diff
@@ -16,3 +20,21 @@ func extractArgs(received []string, names ...string) (map[string]string, error)
 
 	return args, nil
 }
+
+func buildLogQueryHelper(id, start, end string, tail int) url.Values {
```
zpatrick: this does't need/use the `id` param

cli/command/test_helpers.go
tlake: Oop, yeah, actually `newTestCommand()` exists right below this. I had created this public function when I thought that the `client` package would need to use things from the `command` package, and had forgotten to take this out. Nice catch. ✅

cli/command/utils.go
tlake: ✅

client/helpers.go
tlake: 😃 ✅

cli/command/service_test.go
tlake: Mm, yep yep. ✅

cli/command/service.go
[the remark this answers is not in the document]
jparsons04: fwiw, I changed all of my references to `FN` in my PR.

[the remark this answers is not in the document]
jparsons04: Actually, looking back on this from my PR, it looks like I removed all of the function variables and instead passed them inline in each case.

cli/command/service.go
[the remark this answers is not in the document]
tlake: ✅

PR #387, opened 2017-10-11
Add cli command task

nehayward:
This adds cli commands for CRD operations on task.

Manually running `cd /cli && go run main.go` with various list, create, read, delete functions.

**Checklist**
- [x] Unit tests


cli/command/task.go
```diff
@@ -10,6 +13,162 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+					cli.IntFlag{
+						Name:  "copies",
+						Value: 1,
+						Usage: "number of copies of deploy to run (default: 1)",
+					},
+					cli.StringSliceFlag{
+						Name:  "env",
+						Usage: "environment variable override in format 'CONTAINER:VAR=VAL' (can be specified multiple times)",
+					},
+					cli.BoolFlag{
```
zpatrick: the `--no-wait` flag is a global bool flag defined in main

cli/command/task.go
```diff
@@ -10,6 +13,162 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+}
+
+func (t *TaskCommand) delete(c *cli.Context) error {
+	return t.deleteHelper(c, "Task", func(TaskID string) (string, error) {
+		return t.client.DeleteTask(TaskID)
+	})
+}
+
+func (t *TaskCommand) list(c *cli.Context) error {
+	TaskSummaries, err := t.client.ListTasks()
```
zpatrick: lowercase var names `taskSummaries`

cli/command/task.go
```diff
@@ -10,6 +13,162 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+
+	return t.printer.PrintTaskSummaries(TaskSummaries...)
+}
+
+func (t *TaskCommand) read(c *cli.Context) error {
+	args, err := extractArgs(c.Args(), "NAME")
+	if err != nil {
+		return err
+	}
+
```
zpatrick: The `NAME` arg needs to be resolved first. For read, we don't resolve just a single id, so you'll need to do a `client.Read` on everything the resolver returns.

cli/command/utils.go
```diff
@@ -16,3 +18,21 @@ func extractArgs(received []string, names ...string) (map[string]string, error)
 
 	return args, nil
 }
+
+func buildQueryHelper(id, start, end string, tail int) url.Values {
+	query := url.Values{}
+
+	if tail > 0 {
+		query.Set("tail", strconv.Itoa(tail))
```
zpatrick: use the `const`s in the `client` package for these keys:
```
query.Set(client.LogQueryParamTail, strconv.Itoa(tail))
```
see: https://github.com/quintilesims/layer0/blob/cli-refactor/client/helpers.go#L19

cli/command/task.go
nehayward: So it's not needed here then.

zpatrick: correct

cli/command/base.go
```diff
@@ -65,18 +68,41 @@ func (b *CommandBase) deleteHelper(c *cli.Context, entityType string, deleteFN f
...
+func (b *CommandBase) waitOnJobHelper(c *cli.Context, jobID string, onCompleteFN func(entityID string) error) error {
+	waitFlag := c.GlobalBool(config.FLAG_NO_WAIT)
+	waitTimeout := c.GlobalDuration(config.FLAG_TIMEOUT)
+
+	if waitFlag {
+		b.printJobHelper(jobID)
+		return nil
+	}
+
+	b.printer.StartSpinner("creating")
```
zpatrick: I realized I made a mistake - this is probably going to be used for `updates` as well, which means the spinner text will change.

cli/command/task.go
```diff
@@ -10,6 +14,157 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+			{
+				Name:      "create",
+				Usage:     "create a new task",
+				Action:    t.create,
+				ArgsUsage: "ENVIRONMENT NAME DEPLOY",
+				Flags: []cli.Flag{
+					cli.IntFlag{
+						Name:  "copies",
+						Value: 1,
+						Usage: "number of copies of deploy to run (default: 1)",
```
zpatrick: I don't think we need `(default:1)` here since the cli package we use prints out the default for you. I could be wrong though

cli/command/task.go
```diff
@@ -10,6 +14,157 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+	if err != nil {
+		return err
+	}
+
+	deployID, err := t.resolveSingleEntityIDHelper("deploy", args["DEPLOY"])
+	if err != nil {
+		return err
+	}
+
+	req := models.CreateTaskRequest{
```
zpatrick: We need to add back in the container overrides to this model/function

cli/command/task.go
```diff
@@ -10,6 +14,157 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+	return t.printer.PrintTaskSummaries(taskSummaries...)
+}
+
+func (t *TaskCommand) read(c *cli.Context) error {
+	args, err := extractArgs(c.Args(), "NAME")
+	if err != nil {
+		return err
+	}
+
+	taskID, err := t.resolveSingleEntityIDHelper("task", args["NAME"])
```
zpatrick: see comment above

cli/command/task.go
```diff
@@ -10,6 +14,157 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+	if err != nil {
+		return err
+	}
+
+	id, err := t.resolveSingleEntityIDHelper("task", args["NAME"])
+	if err != nil {
+		return err
+	}
+
+	query := buildLogQueryHelper(id, c.String(client.LogQueryParamStart), c.String(client.LogQueryParamEnd), c.Int(client.LogQueryParamTail))
```
zpatrick: you're mixing up the flag names used for `c.String()` and the query paramater names. In this scope, you should be using flag names. Within the body of `buildLogQueryHelper`, you should be using the `client` constants.

cli/command/task.go
nehayward: I noticed https://github.com/quintilesims/layer0/blob/03b20129285df3e1b92da3393ec2844c405768dd/common/models/container_override.go

this isn't in the cli-refactor is this still needed to do the override.

cli/command/task.go
nehayward: Got it. The c.String looks at what's passed by the cli, right?

cli/command/task.go
nehayward: Should I add the parse override to the util.go?

zpatrick: I think it makes sense to keep it in this file, since this file will be the only one using it.

cli/command/task.go
zpatrick: Right - it needs to match the flag name you specified in like `cli.StringFlag{ Name: "..." }`

cli/command/base.go
```diff
@@ -59,15 +62,15 @@ func (b *CommandBase) deleteHelper(c *cli.Context, entityType string, deleteFN f
...
+	if waitFlag {
 		b.printJobResponse(jobID)
 		return nil
 	}
 
 	b.printer.StartSpinner("deleting")
 	defer b.printer.StopSpinner()
 
-	if _, err := client.WaitForJob(b.client, jobID, c.GlobalDuration(config.FLAG_TIMEOUT)); err != nil {
+	if _, err := client.WaitForJob(b.client, jobID, waitTimeout); err != nil {
```
zpatrick: [nit] i think it was better the way it was before

cli/command/base.go
```diff
@@ -76,15 +79,18 @@ func (b *CommandBase) deleteHelper(c *cli.Context, entityType string, deleteFN f
 }
 
 func (b *CommandBase) waitOnJobHelper(c *cli.Context, jobID, spinnerText string, onCompleteFN func(entityID string) error) error {
-	if c.GlobalBool(config.FLAG_NO_WAIT) {
+	waitFlag := c.GlobalBool(config.FLAG_NO_WAIT)
```
zpatrick: [nit] ditto

cli/command/task.go
```diff
@@ -10,6 +16,221 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+	}
+
+	taskIDs, err := t.resolver.Resolve("task", args["NAME"])
+	if err != nil {
+		return err
+	}
+
+	tasks := make([]*models.Task, len(taskIDs))
+	for i, taskID := range taskIDs {
+		if taskExists(taskID) {
```
zpatrick: we should throw in a debug statement here: 
```
if !taskExists[taskID] {
    log.Printf("[DEBUG] Resolver returned an expired task '%s'", taskID)
    continue
}

task, err := t.client.ReadTask(taskID)
...
```

(note that we try to avoid if/else blocks by using `continue` statements)

cli/command/task.go
```diff
@@ -10,6 +16,221 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+}
+
+func (t *TaskCommand) read(c *cli.Context) error {
+
+	taskSummaries, err := t.client.ListTasks()
+	if err != nil {
+		return err
+	}
+
+	taskExists := func(id string) bool {
```
zpatrick: it would be faster to have `taskExists` be a `map[string]bool`, e.g.:
```
taskExists := map[string]bool{}
for _, taskSummary := range taskSummaries {
    taskExists[taskSummary.TaskID] = true
}

...
for _, taskID := range taskIDs {
    if taskExists[taskID] {
...
```

cli/command/task_test.go
```diff
@@ -0,0 +1,295 @@
...
+func TestCreateTask(t *testing.T) {
+	testWaitHelper(t, func(t *testing.T, wait bool) {
+		base, ctrl := newTestCommand(t)
+		defer ctrl.Finish()
+		taskCommand := NewTaskCommand(base.Command())
+		args := Args{"env_name", "dpl_name", "task_name"}
+		flags := Flags{
+			"copies": 1,
+			"env":    []string{"container:key=val"},
+		}
```
zpatrick: newline after closing bracket

cli/command/task_test.go
```diff
@@ -0,0 +1,295 @@
...
+	"github.com/quintilesims/layer0/common/models"
+	"github.com/stretchr/testify/assert"
+	"github.com/urfave/cli"
+)
+
+func TestCreateTask(t *testing.T) {
+	testWaitHelper(t, func(t *testing.T, wait bool) {
+		base, ctrl := newTestCommand(t)
+		defer ctrl.Finish()
+		taskCommand := NewTaskCommand(base.Command())
```
zpatrick: this block could use some better spacing

cli/command/task_test.go
```diff
@@ -0,0 +1,295 @@
...
+			Return(Args{"env_id"}, nil)
+
+		base.Client.EXPECT().
+			CreateTask(req).
+			Return("job_id", nil)
+
+		if wait {
+			base.Client.EXPECT().
+				ReadJob("job_id").
+				Return(&models.Job{
```
zpatrick: we typically don't add multi-line objects in our mock-chains since it makes things so hard to read. 
Also, don't use raw strings for the `job.Status` field (or `job.JobType` if you ever encounter that) - we have special types in the `api/job` package for that. 

For example, this would be fine: 
```
base.Client.EXPECT().
    ReadJob("job_id").
    Return(&models.Job{JobD: "job_id"}, nil)
```

but this should be:
```
job := &models.Job{
    Result: "task_id",
}

base.Client.EXPECT().
    ReadJob("job_id").
    Return(job, nil)
```

cli/command/task_test.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	for name, c := range contexts {
+		if err := taskCommand.create(c); err == nil {
+			t.Fatalf("%s: error was nil!", name)
+		}
+	}
+}
+
+func TestDeleteTask(t *testing.T) {
```
zpatrick: need to test wait vs. non-wait

cli/command/task_test.go
```diff
@@ -0,0 +1,295 @@
...
+		if err := taskCommand.delete(c); err == nil {
+			t.Fatalf("%s: error was nil!", name)
+		}
+	}
+}
+
+func TestReadTask(t *testing.T) {
+	base, ctrl := newTestCommand(t)
+	defer ctrl.Finish()
+	taskCommand := NewTaskCommand(base.Command())
```
zpatrick: [nit] spacing

cli/command/task_test.go
```diff
@@ -0,0 +1,295 @@
...
+		t.Fatal(err)
+	}
+}
+
+func TestReadTask_userInputErrors(t *testing.T) {
+	base, ctrl := newTestCommand(t)
+	defer ctrl.Finish()
+	taskCommand := NewTaskCommand(base.Command())
+
+	base.Client.EXPECT().
```
zpatrick: We need to change the implementation to do resolving / arg checking first. 
Part of the `userInputErrors` tests is checking that we don't 'waste' api calls when user input is wrong.

cli/command/task_test.go
```diff
@@ -0,0 +1,295 @@
...
+		if err := taskCommand.read(c); err == nil {
+			t.Fatalf("%s: error was nil!", name)
+		}
+	}
+}
+
+func TestReadTask_expiredTasks(t *testing.T) {
+	base, ctrl := newTestCommand(t)
+	defer ctrl.Finish()
+	taskCommand := NewTaskCommand(base.Command())
```
zpatrick: [nit] spacing

cli/command/task_test.go
```diff
@@ -0,0 +1,295 @@
...
+	result := []*models.TaskSummary{
+		{TaskID: "task_id"},
+	}
+
+	base.Resolver.EXPECT().
+		Resolve("task", "task_name").
+		Return(Args{"task_id", "task_id2", "task_id3"}, nil)
```
zpatrick: [nit] I believe the shorthand we agreed upon in slack was `tsk_name` and `tsk_id1`, `tsk_id2`, etc..
Also, it could probably be a little clearer if we did `Return([]string{"tsk_id1", "expired1", "expired2"}...`

zpatrick: using the `Args` type doesn't make a lot of sense here. Should stick with `[]string`

cli/command/task_test.go
```diff
@@ -0,0 +1,295 @@
...
+	taskCommand := NewTaskCommand(base.Command())
+
+	base.Resolver.EXPECT().
+		Resolve("task", "name").
+		Return(Args{"id"}, nil)
+
+	query := buildLogQueryHelper("id", "start", "end", 100)
+
+	base.Client.EXPECT().
+		ReadTaskLogs("id", query)
```
zpatrick: [nit] should probably `.Return([]*models.LogFile{}, nil)`

cli/command/task_test.go
```diff
@@ -0,0 +1,295 @@
...
+
+	for name, c := range contexts {
+		if err := taskCommand.logs(c); err == nil {
+			t.Fatalf("%s: error was nil!", name)
+		}
+	}
+}
+
+func TestParseOverrides(t *testing.T) {
+	input := Args{
```
zpatrick: using the `Args` type doesn't make sense here. (I'm guessing this was a `sed` command gone wrong?)

cli/command/task_test.go
```diff
@@ -0,0 +1,295 @@
...
+}
+func TestParseOverridesErrors(t *testing.T) {
+	cases := map[string]string{
+		"Missing CONTAINER": ":key=val",
+		"Missing KEY":       "container:=val",
+		"Missing VAL":       "container:key=",
+	}
+
+	for name, input := range cases {
+		if _, err := parseOverrides(Args{input}); err == nil {
```
zpatrick: ditto `Args`

common/models/create_task_request.go
```diff
@@ -7,9 +7,10 @@ import (
 )
 
 type CreateTaskRequest struct {
-	TaskName      string `json:"task_name"`
-	EnvironmentID string `json:"environment_id"`
-	DeployID      string `json:"deploy_id"`
+	ContainerOverrides []ContainerOverride `json:"container_overrides"`
```
zpatrick: We need to add the `container_overrides` field into the swagger definition below

common/models/container_override.go
```diff
@@ -0,0 +1,6 @@
+package models
+
+type ContainerOverride struct {
```
zpatrick: this needs a swagger definition and to be defined in `controllers/swagger.go`.

I don't think there's a type for `maps` in swagger yet, so probably just say it's a `string` type for now and throw a `todo` statement.

cli/command/task_test.go
```diff
@@ -0,0 +1,295 @@
...
+	output, err := parseOverrides(input)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	assert.Equal(t, len(output), len(expected))
+	assert.Equal(t, expected[0], output[0])
+	assert.Equal(t, expected[1], output[1])
+}
+func TestParseOverridesErrors(t *testing.T) {
```
zpatrick: [nit] newline between tests

cli/command/task.go
```diff
@@ -10,6 +16,221 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+	return t.printer.PrintTasks(tasks...)
+}
+
+func (t *TaskCommand) logs(c *cli.Context) error {
+	args, err := extractArgs(c.Args(), "NAME")
+	if err != nil {
+		return err
+	}
+
+	id, err := t.resolveSingleEntityIDHelper("task", args["NAME"])
```
zpatrick: [nit] `taskID`

cli/command/task_test.go
nehayward: Yup. My bad.

cli/command/task.go
nehayward: I was wondering about this but wasn't sure if the conversion to map would be okay. Thanks though I agree.

cli/command/task.go
nehayward: Does it need the continue here? Won't it continue without it?

cli/command/task.go
```diff
@@ -10,6 +17,220 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+func (t *TaskCommand) Command() cli.Command {
+	return cli.Command{
+		Name:  "task",
+		Usage: "manage layer0 tasks",
+		Subcommands: []cli.Command{
+			{
+				Name:      "create",
+				Usage:     "create a new task",
+				Action:    t.create,
+				ArgsUsage: "ENVIRONMENT NAME DEPLOY",
```
jparsons04: [nit] We were having a discussion about the `ArgsUsage` `NAME` becoming `TASK_NAME`. I think this is the currently agreed consensus on the discussions I saw about this.

cli/command/task.go
```diff
@@ -10,6 +17,220 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+					cli.StringSliceFlag{
+						Name:  "env",
+						Usage: "environment variable override in format 'CONTAINER:VAR=VAL' (can be specified multiple times)",
+					},
+				},
+			},
+			{
+				Name:      "delete",
+				Usage:     "delete a Task",
+				ArgsUsage: "NAME",
```
jparsons04: [nit] Same as above, `TASK_NAME`

cli/command/task.go
```diff
@@ -10,6 +17,220 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+				Name:      "list",
+				Usage:     "list all Tasks",
+				Action:    t.list,
+				ArgsUsage: " ",
+			},
+			{
+				Name:      "read",
+				Usage:     "describe a Task",
+				Action:    t.read,
+				ArgsUsage: "NAME",
```
jparsons04: [nit] `TASK_NAME`

cli/command/task.go
```diff
@@ -10,6 +17,220 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+			},
+			{
+				Name:      "logs",
+				Usage:     "get the logs for a task",
+				Action:    t.logs,
+				ArgsUsage: "NAME",
```
jparsons04: [nit] `TASK_NAME`

cli/command/task.go
```diff
@@ -10,6 +17,220 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+						Usage: "the end of the time range to fetch logs (format: YYYY-MM-DD HH:MM)",
+					},
+				},
+			},
+		},
+	}
+}
+
+func (t *TaskCommand) create(c *cli.Context) error {
+	args, err := extractArgs(c.Args(), "ENVIRONMENT", "NAME", "DEPLOY")
```
jparsons04: If you make the changes above, `NAME` must become `TASK_NAME`

cli/command/task.go
```diff
@@ -10,6 +17,220 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+		return err
+	}
+
+	deployID, err := t.resolveSingleEntityIDHelper("deploy", args["DEPLOY"])
+	if err != nil {
+		return err
+	}
+
+	req := models.CreateTaskRequest{
+		TaskName:           args["NAME"],
```
jparsons04: If you make the changes above, `NAME` must become `TASK_NAME`

cli/command/task.go
```diff
@@ -10,6 +17,220 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+	if err != nil {
+		return err
+	}
+
+	taskExists := map[string]bool{}
+	for _, taskSummary := range taskSummaries {
+		taskExists[taskSummary.TaskID] = true
+	}
+
+	args, err := extractArgs(c.Args(), "NAME")
```
jparsons04: If you make the changes above, `NAME` must become `TASK_NAME`

cli/command/task.go
```diff
@@ -10,6 +17,220 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+	if err != nil {
+		return err
+	}
+
+	taskIDs, err := t.resolver.Resolve("task", args["NAME"])
```
jparsons04: If you make the changes above, `NAME` must become `TASK_NAME`

cli/command/task.go
jparsons04: If you make the changes above, `NAME` must become `TASK_NAME`

cli/command/task.go
```diff
@@ -10,6 +17,220 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+
+			tasks[i] = task
+		}
+	}
+
+	return t.printer.PrintTasks(tasks...)
+}
+
+func (t *TaskCommand) logs(c *cli.Context) error {
+	args, err := extractArgs(c.Args(), "NAME")
```
jparsons04: If you make the changes above, `NAME` must become `TASK_NAME`

cli/command/task.go
diemonster: [nit] remove `(default: 1)`

cli/command/task.go
```diff
@@ -10,6 +17,220 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+	})
+}
+
+func (t *TaskCommand) list(c *cli.Context) error {
+	taskSummaries, err := t.client.ListTasks()
+	if err != nil {
+		return err
+	}
+
+	return t.printer.PrintTaskSummaries(taskSummaries...)
```
diemonster: [nit] we may want to spin this off into a separate issue, but i've found during testing (and experience) that this command can take quite a while to return, especially when there are many tasks. I think we may want to consider adding the spinner behavior for this as a result, and possibly other `list` operations as well.

nehayward: Yeah I think it should be a separate issue, but if there is a long wait we should change this.

cli/command/task.go
```diff
@@ -10,6 +17,219 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+		return err
+	}
+
+	query := buildLogQueryHelper(taskID, c.String("start"), c.String("end"), c.Int("tail"))
+	logs, err := t.client.ReadTaskLogs(taskID, query)
+	if err != nil {
+		return err
+	}
+
+	return t.printer.PrintLogs(logs...)
```
diemonster: [nit] This can take a while to return as well, consider adding the job and/or spinner behavior

common/models/create_task_request.go
```diff
@@ -35,6 +36,8 @@ func (c CreateTaskRequest) Definition() swagger.Definition {
 			"task_name":      swagger.NewStringProperty(),
 			"environment_id": swagger.NewStringProperty(),
 			"deploy_id":      swagger.NewStringProperty(),
+			// TODO: Change to Map defintion
```
diemonster: Is this still an intended TODO?

nehayward: Yes it is. @zpatrick suggested it until we add maps to swagger

cli/command/task.go
```diff
@@ -10,6 +17,219 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+		Subcommands: []cli.Command{
+			{
+				Name:      "create",
+				Usage:     "create a new task",
+				Action:    t.create,
+				ArgsUsage: "ENVIRONMENT TASK_NAME DEPLOY",
+				Flags: []cli.Flag{
+					cli.IntFlag{
+						Name:  "copies",
+						Usage: "number of copies of deploy to run (default: 1)",
```
zpatrick: The `copies` param is no longer supported

cli/command/task.go
zpatrick: I'm hoping with the updated api changes this command will actually be pretty fast. I haven't tested with a ton of tasks yet, but hoping this will be a non-issue

cli/command/task.go
```diff
@@ -10,6 +17,219 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+	taskSummaries, err := t.client.ListTasks()
+	if err != nil {
+		return err
+	}
+
+	return t.printer.PrintTaskSummaries(taskSummaries...)
+}
+
+func (t *TaskCommand) read(c *cli.Context) error {
+	taskSummaries, err := t.client.ListTasks()
```
zpatrick: You need to do this after parsing args and resolving. We don't want to waste this api call if we are just going to error out.

cli/command/task.go
```diff
@@ -10,6 +17,219 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+	if err != nil {
+		return err
+	}
+
+	taskIDs, err := t.resolver.Resolve("task", args["TASK_NAME"])
+	if err != nil {
+		return err
+	}
+
+	tasks := make([]*models.Task, len(taskIDs))
```
zpatrick: Can't use `make` in this way here since we aren't sure every task will be read. 

However, you can use `make([]*models.Task, 0, len(taskIDs))`. This means "create a new slice of length 0, but prepare to allocate up to `len(taskIDs)` worth of memory". With that method, you will need to use `append` to the slice instead of `tasks[i] = ...`

cli/command/task_test.go
```diff
@@ -0,0 +1,317 @@
...
+	base, ctrl := newTestCommand(t)
+	defer ctrl.Finish()
+
+	taskCommand := NewTaskCommand(base.Command())
+
+	contexts := map[string]*cli.Context{
+		"Missing TASK_NAME arg": NewContext(t, nil, nil),
+	}
+
+	base.Client.EXPECT().
```
zpatrick: we shouldn't expect to make api calls when a user input error occurs.

cli/command/task.go
```diff
@@ -10,6 +17,215 @@ func NewTaskCommand(b *CommandBase) *TaskCommand {
...
+	}
+
+	tasks := make([]*models.Task, 0, len(taskIDs))
+	for _, taskID := range taskIDs {
+		if !taskExists[taskID] {
+			log.Printf("[DEBUG] Resolver returned an expired task '%s'", taskID)
+			continue
+		}
+
+		if taskExists[taskID] {
```
zpatrick: We don't need this wrapped in a `if taskExists[taskID]` block since we have `if !taskExists[taskID] { ... continue }` above

cli/command/task_test.go
```diff
@@ -0,0 +1,313 @@
...
+	taskCommand := NewTaskCommand(base.Command())
+
+	contexts := map[string]*cli.Context{
+		"Missing ENVIRONMENT arg": NewContext(t, nil, nil),
+		"Missing TASK_NAME arg":   NewContext(t, Args{"environment"}, nil),
+		"Missing DEPLOY arg":      NewContext(t, Args{"environment", "name"}, nil),
+	}
+
+	for name, c := range contexts {
+		if err := taskCommand.create(c); err == nil {
```
zpatrick: [nit] We've been switching over to use  `t.Run` when doing table tests, e.g.:
```
    t.Run(name, func(t *testing.T){
        if err := taskCommand.Create(c); err == nil {
            t.Fatalf("Error was nil!")
        }
    })
}
```

cli/command/task_test.go
```diff
@@ -0,0 +1,313 @@
...
+	defer ctrl.Finish()
+
+	taskCommand := NewTaskCommand(base.Command())
+
+	contexts := map[string]*cli.Context{
+		"Missing TASK_NAME arg": NewContext(t, nil, nil),
+	}
+
+	for name, c := range contexts {
+		if err := taskCommand.delete(c); err == nil {
```
zpatrick: [nit] ditto `t.Run`

cli/command/task_test.go
```diff
@@ -0,0 +1,313 @@
...
+	defer ctrl.Finish()
+
+	taskCommand := NewTaskCommand(base.Command())
+
+	contexts := map[string]*cli.Context{
+		"Missing TASK_NAME arg": NewContext(t, nil, nil),
+	}
+
+	for name, c := range contexts {
+		if err := taskCommand.read(c); err == nil {
```
zpatrick: [nit] ditto

cli/command/utils.go
```diff
@@ -16,3 +20,21 @@ func extractArgs(received []string, names ...string) (map[string]string, error)
 
 	return args, nil
 }
+
+func buildLogQueryHelper(id, start, end string, tail int) url.Values {
```
zpatrick: this doesn't need/use the `id` param

cli/command/task_test.go
```diff
@@ -0,0 +1,313 @@
...
+	base, ctrl := newTestCommand(t)
+	defer ctrl.Finish()
+	taskCommand := NewTaskCommand(base.Command())
+
+	contexts := map[string]*cli.Context{
+		"Missing TASK_NAME arg": NewContext(t, nil, nil),
+	}
+
+	for name, c := range contexts {
+		if err := taskCommand.logs(c); err == nil {
```
zpatrick: [nit] ditto

cli/command/task_test.go
```diff
@@ -0,0 +1,313 @@
...
+			EnvironmentOverrides: map[string]string{"k1": "v1"},
+		},
+	}
+
+	output, err := parseOverrides(input)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	assert.Equal(t, len(output), len(expected))
```
zpatrick: Can't you just use `assert.Equal(t, expected, output)`?

cli/command/task_test.go
```diff
@@ -0,0 +1,313 @@
...
+
+func TestParseOverridesErrors(t *testing.T) {
+	cases := map[string]string{
+		"Missing CONTAINER": ":key=val",
+		"Missing KEY":       "container:=val",
+		"Missing VAL":       "container:key=",
+	}
+
+	for name, input := range cases {
+		if _, err := parseOverrides(Args{input}); err == nil {
```
zpatrick: Still need to convert `Args` to `[]string` here

cli/command/task_test.go
nehayward: Yes you can.

cli/command/task_test.go
```diff
@@ -293,7 +301,7 @@ func TestParseOverrides(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	assert.Equal(t, len(output), len(expected))
+	assert.Equal(t, output, expected)
```
zpatrick: params need to be reversed here, `assert.Equal(t, expected, output)`

nehayward: got it

PR #390, opened 2017-10-11
Resolver and Tag controller refactor

zpatrick:
Adds tag resolver and the tag controller. 

unit tests

**Checklist**
- [x] Unit tests


cli/resolver/tag.go
```diff
@@ -1,11 +1,104 @@
...
+	"fmt"
+	"net/url"
+	"strings"
+
 	"github.com/quintilesims/layer0/client"
+	"github.com/quintilesims/layer0/common/errors"
 )
 
-func NewTagResolver(client client.Client) ResolverFunc {
+// todo: support glob matching '*' in api
```
zpatrick: need to remove this

api/controllers/tag_test.go
```diff
@@ -0,0 +1,173 @@
...
+				client.TagQueryParamEnvironmentID: []string{"eid1"},
+			},
+			CheckResult: func(t *testing.T, result models.Tags) {
+				assert.False(t, result.Any(func(tag models.Tag) bool {
+					return tag.Key == "environment_id" && tag.Value != "eid1"
+				}))
+			},
+		},
+		 {
+                        Name: "fuzz+type",
```
zpatrick: I should probably add a few more cases for blob matching

api/controllers/swagger.go
```diff
@@ -423,6 +429,51 @@ func (s *SwaggerController) ServeSwaggerSpec(c *fireball.Context) (fireball.Resp
 					},
 				},
 			},
+			"/tag": map[string]swagger.Method{
+				"delete": {
```
jparsons04: [nit] Other entities in the swagger controller list List first, then Add, then Update (if applicable), then Delete

api/controllers/tag.go
```diff
@@ -0,0 +1,201 @@
...
+			return nil, err
+		}
+
+		return tags.WithID(entityID), nil
+	}
+
+	return t.TagStore.SelectAll()
+}
+
+func filterTagsByEnvironmentID(entityTags map[string]models.Tags, environmentID string) {
```
sesh-kebab: [nit] wouldn't it be better to return a new filtered map rather than modifying the maps argument passed in?

api/controllers/tag.go
```diff
@@ -0,0 +1,201 @@
...
+			return t.Key == "environment_id" && t.Value == environmentID
+		})
+
+		if !hasMatch {
+			delete(entityTags, entityID)
+		}
+	}
+}
+
+func filterTagsByName(entityTags map[string]models.Tags, name string) {
```
sesh-kebab: [nit] same as above - but I see there are a few methods which modify the argument passed in to filter. preferable in go as maps are actually pointers?

api/controllers/tag_test.go
```diff
@@ -0,0 +1,195 @@
...
+		{EntityID: "did2", EntityType: "deploy", Key: "name", Value: "dname2"},
+		{EntityID: "did2", EntityType: "deploy", Key: "version", Value: "2"},
+	}
+
+	store := tag.NewMemoryStore()
+	for _, tag := range tags {
+		store.Insert(tag)
+	}
+
+	testCases := []struct {
```
sesh-kebab: OMG this table test is so sick! 😲

cli/resolver/tag.go
```diff
@@ -1,11 +1,103 @@
...
+			case 2:
+				return resolveFullyQualifiedEntity(c, entityType, targets[0], targets[1])
+			default:
+				return nil, fmt.Errorf("Invalid target format (expected ENVIRONMENT:%s)", strings.ToUpper(entityType))
+			}
+		case "deploy":
+			targets := strings.Split(target, ":")
+			switch len(targets) {
+			case 1:
+				return resolveUnqualifiedEntity(c, entityType, target)
```
sesh-kebab: Do you not need an `resolveUnqualifiedDeploy` here as the `resolveUnqualifiedEntity` doesn't filter by deploy entity type?

cli/resolver/tag.go
```diff
@@ -1,11 +1,103 @@
...
+	}
+
+	// check if we have an exact ID match
+	if tag, ok := tags.WithID(target).First(); ok {
+		return []string{tag.EntityID}, nil
+	}
+
+	uniqueEntityIDs := map[string]bool{}
+	for _, tag := range tags {
+		uniqueEntityIDs[tag.EntityID] = true
```
sesh-kebab: This is neat!

common/models/tags.go
```diff
@@ -59,3 +61,42 @@ func (t Tags) Any(f filter) bool {
...
+// sorting functions
+func (t Tags) Len() int {
+	return len(t)
+}
+
+func (t Tags) Swap(i, j int) {
+	t[i], t[j] = t[j], t[i]
+}
+
+func (t Tags) Less(i, j int) bool {
```
sesh-kebab: Is this needed? I can't see where is is being used?

api/controllers/tag.go
zpatrick: I think it's kind of a toss-up given the context. I went with in-place filtering because it's less code and slightly better perf.

cli/resolver/tag.go
zpatrick: I'm not 100% sure I understand your question, but `resolveUnqualifiedEntity` doesn't do any filtering period. It just fuzzy matches based of the "raw" input and whatever `entityType` you pass in (in this case, we know it's `"deploy"`). 

Or maybe you are saying this should be: `return resolveUnqualifiedEntity(c, "deploy", target)`?

cli/resolver/tag.go
zpatrick: oh, just realized I can use `tags.GroupByEnity`

cli/resolver/tag_test.go
```diff
@@ -0,0 +1,207 @@
...
+			Target:     "e*:s*",
+			ExpectedQueries: []url.Values{
+				{
+					client.TagQueryParamType: []string{"environment"},
+					client.TagQueryParamFuzz: []string{"e*"},
+				},
+				{
+					client.TagQueryParamType:          []string{"service"},
+					client.TagQueryParamFuzz:          []string{"s*"},
+					client.TagQueryParamEnvironmentID: []string{"eid"},
```
zpatrick: @seshi: is this too much "magic"? Is it clear where `"eid"` comes from?

common/models/tags.go
zpatrick: It was in my old tests, but now it's not.

diemonster:
I'd like to understand from a higher level how this differs from the existing `tag` and `resolver` implementations. Not feeling like I have enough context to ask the right (annoying) questions. Could you list pros and cons and/or help me understand where the improvements are?

zpatrick:
@bivers: There's actually very little difference. Most changes are just cleaning up the existing code to make it more readable and updating the unit tests to be more robust. 

The one "big" change is that fuzzy matching now supports glob matching instead of just suffix wildcards.

cli/resolver/tag_test.go
sesh-kebab: Assuming it comes from your mock because e* will always resolve to eid? If not then yes this is too much magic.

cli/resolver/tag.go
sesh-kebab: Oh I see - upon reading this again, I misunderstood what the code was doing.

cli/resolver/tag.go
sesh-kebab: exactly... 😅

api/controllers/tag.go
```diff
@@ -0,0 +1,201 @@
...
+func (t *TagController) selectTagsFromQuery(query url.Values) (models.Tags, error) {
+	entityID := query.Get(client.TagQueryParamID)
+	entityType := query.Get(client.TagQueryParamType)
+
+	if entityType != "" && entityID != "" {
+		return t.TagStore.SelectByTypeAndID(entityType, entityID)
+	}
+
+	if entityType != "" {
+		return t.TagStore.SelectByType(entityType)
```
diemonster: My ape-like brain has some difficulty following the flow here. We're calling `selectTagsFromQuery`() from within `ListTags()`, but our return is coming from a different package called from the `selectTagsFromQuery()` helper. This seems inconsistent with how we do things elsewhere within the `controller` package. I've found it a lot more straightforward to see something like:

```
func (j *JobController) DeleteJob(c *fireball.Context) (fireball.Response, error) {
	id := c.PathVariables["id"]
	if err := j.JobStore.Delete(id); err != nil {
		return nil, err
	}

	return fireball.NewJSONResponse(200, nil)
}
```

I know the above example is a bit unfair since it's doing a lot less, but it seems like using the `provider` abstraction here might give us some benefit in clarity. Then we could do all of the `client` interaction in there and keep this abstraction a little more focused. Thoughts?

api/controllers/tag.go
```diff
@@ -0,0 +1,201 @@
...
+			return t.Key == "name" && t.Value == name
+		})
+
+		if !hasMatch {
+			delete(entityTags, entityID)
+		}
+	}
+}
+
+func filterTagsByIDOrNameGlob(entityTags map[string]models.Tags, pattern string) {
```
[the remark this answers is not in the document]
zpatrick: abbreviations are upper-case in go

cli/resolver/tag_test.go
zpatrick: ok, fixed it to make it more clear

api/controllers/tag.go
```diff
@@ -0,0 +1,201 @@
...
+
+	filteredTags := models.Tags{}
+	for _, tags := range entityTags {
+		filteredTags = append(filteredTags, tags...)
+	}
+
+	return fireball.NewJSONResponse(200, filteredTags)
+}
+
+func (t *TagController) selectTagsFromQuery(query url.Values) (models.Tags, error) {
```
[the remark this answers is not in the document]
zpatrick: We use query for path query and req for the request body.

api/controllers/tag.go
zpatrick: The tag controller is a much different beast than the entity controllers. Even further, the ListTags function is a huge outlier to anything else in the controller package. Moving this logic elsewhere would just be another example of "kicking the complexity down the road" (e.g. the old `logic` package), which we know usually does more harm than good.

api/controllers/tag.go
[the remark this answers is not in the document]
diemonster: I know they're kind of hard-line about it, which is why I said `mega-nit`... but it's still funny looking. Makes it seem like `IDO` is a thing.

diemonster:
Would now be a bad time for `UpdateTag()` ? We've wanted this for a while.

api/controllers/tag.go
diemonster: I get what you're saying, and the more I think about it, the less it would make sense to move this logic to a different package given how this all currently works. 

Anyway, after sitting with this PR for a good chunk of the morning, I get the logic, but I think my comments have more to do with the flow of the functions --- it seems more in the mould of our pre-refactor efforts. Maybe that's acceptable. But if one of the rest of us wanted to add `UpdateTags()` ourselves, I'm guessing it would be pretty painful.

api/controllers/tag.go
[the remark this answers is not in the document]
diemonster: ✅

api/controllers/tag.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	if err := t.TagStore.Insert(tag); err != nil {
+		return nil, err
+	}
+
+	return fireball.NewJSONResponse(200, tag)
+}
+
+func (t *TagController) ListTags(c *fireball.Context) (fireball.Response, error) {
```
diemonster: Here's a (perhaps) more constructive idea -- let's add some docs to these methods that describe the flow of what's happening similar to what @jparsons04 did in the `provider` layer. I realize this wouldn't be consistent with the lack of docs elsewhere in `controller`, but as this is an outlier, it seems appropriate to call it out as such.

zpatrick: added

diemonster: cool, that does help

PR #394, opened 2017-10-11
274: CLI Load Balancer Command

jparsons04:
Adds the Load Balancer CLI command

- [x] addport
- [x] create
- [x] delete
- [x] dropport
- [x] healthcheck
- [x] list
- [x] get

Manual CLI testing
Unit tests

**Checklist**
- [x] Unit tests

Continued work on issue #274 

**Note:** There are changes I've made within the `cli.Command{}` section of the main command file that are departures from the cli usage display patterns in other commands. This is done in the spirit of wanting to raise points about the clarity of the CLI in its current form and how the help output can be made clearer.


cli/command/load_balancer_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+func TestCreateLoadBalancer(t *testing.T) {
+	base, ctrl := newTestCommand(t)
+	defer ctrl.Finish()
+
+	loadBalancerCommand := NewLoadBalancerCommand(base.Command())
+	c := &cli.Context{}
+	loadBalancerCommand.create(c)
+
+	ports := []models.Port{}
```
zpatrick: You typically shouldn't use 0-values when testing things of this nature. It's easier to catch bugs when things aren't 0-values Ports should have at least one object, healthcheck should be populated, isPublic should be true.

cli/command/load_balancer_test.go
```diff
@@ -0,0 +1,40 @@
...
+		HealthCheck:      healthCheck,
+	}
+
+	base.client.EXPECT().
+		CreateLoadBalancer(req).
+		Return(models.Job{JobID: "jid"}, nil)
+
+	base.client.EXPECT().
+		ReadJob("jid").
+		Return(models.Job{JobID: "lid"}, nil)
```
zpatrick: small thing: this should be `models.Job{Result: "lid"}`

cli/command/load_balancer_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/quintilesims/layer0/common/models"
+	"github.com/urfave/cli"
+)
+
+func TestCreateLoadBalancer(t *testing.T) {
+	base, ctrl := newTestCommand(t)
+	defer ctrl.Finish()
+
+	loadBalancerCommand := NewLoadBalancerCommand(base.Command())
+	c := &cli.Context{}
```
zpatrick: use the `getCLIContext` helper function (in `test_helpers.go`) to generate your context. It takes positional args and flags to generate the context. For example:
```
// Args is an alias for []string 
args := Args{"env", "name"}

// Flags is an alias for map[string]interface{}
flags := Flags{
    "public": true,
    "port": []string{"80:80/tcp", "443:443/https"},
    "healthcheck-target": "tcp:80", 
    ...
}

c := getCLIContext(t, args, flags)
command := NewLoadBalancerCommand(base.Command())
if err := command.Create(c); err != nil {
    t.Fatal(err)
}
```

cli/command/load_balancer_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+	req := models.CreateLoadBalancerRequest{
+		LoadBalancerName: "lname",
+		EnvironmentID:    "eid",
+		IsPublic:         false,
+		Ports:            ports,
+		HealthCheck:      healthCheck,
+	}
+
+	base.client.EXPECT().
```
zpatrick: looks like you forgot to setup your mock resolver:
```
base.Resolver.EXPECT().
    Resolve("environment", "env").
    Return([]string{"eid"}, nil)
```

cli/command/load_balancer_test.go
```diff
@@ -0,0 +1,40 @@
+package command
+
+import (
+	"testing"
+
+	"github.com/quintilesims/layer0/common/models"
+	"github.com/urfave/cli"
+)
+
+func TestCreateLoadBalancer(t *testing.T) {
```
zpatrick: We should add another `TestCreateLoadBalancerDefaults` that ensure all the default arguments get passed correctly.

cli/command/load_balancer.go
```diff
@@ -481,3 +477,15 @@ func parsePort(port, certificateName string) (*models.Port, error) {
 
 	return model, nil
 }
+
+func parseTarget(target string) (string, error) {
+	split := strings.FieldsFunc(target, func(r rune) bool {
+		return r == ':' || r == '/'
+	})
+
+	if len(split) != 3 && strings.ToLower(split[0]) == "https" {
```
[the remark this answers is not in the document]
jparsons04: The slash pathing is only required if you specify HTTPS as the protocol of a target. If this check isn't done, and this condition is hit, AWS itself returns this same error, and worse, a load balancer is created with a default target (TCP:80) and l0 is not able to see this target.

```$ ./l0 loadbalancer create --port 443:443/https --certificate imsio2016-2019 --private --healthcheck-target HTTPS:443 --healthcheck-interval 31 --healthcheck-timeout 6 --healthcheck-healthy-threshold 3 --healthcheck-unhealthy-threshold 3 lbcomma7fdf7 lbcommand-lb
Creating Load Balancer. . 
An error occurred during the job's execution: ValidationError: HealthCheck HTTPS Target must specify a port followed by a path that begins with a slash. e.g. HTTP:80/ping/this/path
	status code: 400, request id: e0520dd2-af8e-11e7-8eb0-e591f7c457bf
```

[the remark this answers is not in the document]
zpatrick: I think you misunderstood my original comment: doesn't this check apply to `http` targets **in addition to** `https` targets?

cli/command/load_balancer.go
```diff
@@ -205,6 +204,10 @@ func (l *LoadBalancerCommand) create(c *cli.Context) error {
 
 	// The default port configuration is filled here as opposed to the cli.Flag.Value
 	if len(ports) == 0 {
+		if c.String("certificate") != "" {
+			fmt.Printf("Warning: Certificate specified, but protocol of port is not HTTPS. Certificate will not be used.\n")
```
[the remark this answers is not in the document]
jparsons04: I'm going to opt to put it in a debug statement. The result of this combination is not intuitive if the user explicitly specifies the cert (accidentally or purposely) and doesn't specify an HTTPS port or, doesn't include the port at all (the default port is 80:80/TCP).

cli/command/load_balancer.go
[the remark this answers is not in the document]
jparsons04: The slash pathing is not required if you specify non-HTTPS protocols.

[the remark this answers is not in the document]
jparsons04: Actually, disregard the last statement, you are right, it does apply to HTTP.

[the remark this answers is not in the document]
zpatrick:
```
$ l0 loadbalancer create dev test --healthcheck-target http:80
ServerError (code=21) AWS Error: healthcheck http target must specify a port followed by a path that begins with a slash. e.g. http:80/ping/this/path (code 'ValidationError')

```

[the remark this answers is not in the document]
jparsons04: :white_check_mark:

cli/command/load_balancer_test.go
```diff
@@ -0,0 +1,40 @@
...
+	base, ctrl := newTestCommand(t)
+	defer ctrl.Finish()
+
```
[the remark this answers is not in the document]
jparsons04: :white_check_mark:

cli/command/load_balancer.go
```diff
@@ -320,49 +328,34 @@ func (l *LoadBalancerCommand) healthcheck(c *cli.Context) error {
 	updateIsRequired := false
```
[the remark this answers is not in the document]
jparsons04: There appears to be a bug with this method from urfave/cli where IsSet isn't returning the correct bool. For now, I've gone back to the previous implementation unless something comes up that allows me to do something like this.

cli/command/load_balancer.go
```diff
@@ -217,13 +218,14 @@ func (l *LoadBalancerCommand) create(c *cli.Context) error {
 		ports = append(ports, port)
 	}
 
-	target, err := parseTarget(c.String("healthcheck-target"))
-	if err != nil {
-		return err
+	if c.String("healthcheck-target") != "" {
```
zpatrick: what is `parseTarget` being used for? If it's just validation, we don't need to do that here since aws takes care of that for us. If we decide to keep it, `validateTarget` is probably a more appropriate name (and it shouldn't return the target since it doesn't manipulate it in any way)

cli/command/load_balancer_test.go
```diff
@@ -6,41 +6,323 @@ import (
...
+			ContainerPort: 80,
+			HostPort:      80,
+			Protocol:      "tcp",
+		},
+	}
 	healthCheck := models.HealthCheck{}
 
+	base.Client.EXPECT().
+		ReadLoadBalancer("lid").
+		Return(&models.LoadBalancer{
```
zpatrick: [nit] this is difficult to read; we should probably put the load balancer model into its own variable.

cli/command/load_balancer_test.go
```diff
@@ -6,41 +6,323 @@ import (
...
+		HealthCheck:    &healthCheck,
+	}
+
+	base.Client.EXPECT().
+		UpdateLoadBalancer(req).
+		Return("jid", nil)
+
+	base.Client.EXPECT().
+		ReadJob("jid").
+		Return(&models.Job{
```
zpatrick: [nit] should probably put the job in a variable - this multi-line `Return()` statement is pretty hard to read

cli/command/load_balancer_test.go
```diff
@@ -6,41 +6,323 @@ import (
...
+		Return([]string{"eid"}, nil)
+
+	ports := []models.Port{
+		models.Port{
+			ContainerPort: 80,
+			HostPort:      80,
+			Protocol:      "tcp",
+		},
+	}
+	healthCheck := models.HealthCheck{}
```
zpatrick: we should test a non zero-value `healthcheck` during `create`

cli/command/load_balancer_test.go
```diff
@@ -6,41 +6,323 @@ import (
...
+			LoadBalancerID:   "lid",
+			LoadBalancerName: "lname",
+			Ports:            ports,
+		}, nil)
+
+	ports = ports[:1]
+	req := models.UpdateLoadBalancerRequest{
+		LoadBalancerID: "lid",
+		Ports:          &ports,
+		HealthCheck:    &healthCheck,
```
zpatrick: this should just be `nil`

cli/command/load_balancer_test.go
```diff
@@ -6,41 +6,323 @@ import (
...
+	ports = append(ports, models.Port{
+		ContainerPort: 81,
+		HostPort:      81,
+		Protocol:      "tcp",
+	})
+
+	req := models.UpdateLoadBalancerRequest{
+		LoadBalancerID: "lid",
+		Ports:          &ports,
+		HealthCheck:    &healthCheck,
```
zpatrick: this should just be `nil`

cli/command/load_balancer_test.go
jparsons04: :white_check_mark:

cli/command/load_balancer.go
```diff
@@ -10,6 +18,497 @@ func NewLoadBalancerCommand(b *CommandBase) *LoadBalancerCommand {
...
+	}
+
+	loadBalancer, err := l.client.ReadLoadBalancer(loadBalancerID)
+	if err != nil {
+		return err
+	}
+
+	if requiresUpdate {
+		if healthCheck.Target != "" {
+			loadBalancer.HealthCheck.Target = healthCheck.Target
```
zpatrick: [nit] ditto as above - not assigning a new variable for the update healthcheck is a little odd. Seems clearer to assign/update a new one, e.g.
```go
loadBalancer, err := client.ReadLoadBalancer(...)
...
healthCheck := loadBalancer.HealthCheck
...
if timeout := c.Int("healthcheck-timeout"); timeout != 0 {
    healthCheck.Timeout = timeout
    requiredUpdate = true
}
...
req := models.UpdateLoadBalancerRequest{
    LoadBalancerID: loadBalancerID,
    HealthCheck: &healthCheck
}

...
```

cli/command/load_balancer.go
```diff
@@ -10,6 +18,497 @@ func NewLoadBalancerCommand(b *CommandBase) *LoadBalancerCommand {
...
+			return err
+		}
+
+		onCompleteFn := func(loadBalancerID string) error {
+			loadBalancer, err := l.client.ReadLoadBalancer(loadBalancerID)
+			if err != nil {
+				return err
+			}
+
+			l.printer.StopSpinner()
```
zpatrick: doesn't the helper function handle calling `l.printer.StopSpinner()`? If not, it probably should w/ a `defer` statement (otherwise the spinner will keep going if an error occurs)

cli/command/load_balancer.go
```diff
@@ -10,6 +18,497 @@ func NewLoadBalancerCommand(b *CommandBase) *LoadBalancerCommand {
...
+			return l.printer.PrintLoadBalancerHealthCheck(loadBalancer)
+		}
+
+		return l.waitOnJobHelper(c, jobID, "Updating health check on load balancer", onCompleteFn)
+
+	} else {
```
zpatrick: try to avoid `if else` blocks if possible. It's usually better to return early than to have nested blocks. In this case that would be:
```
if !requiredUpdate {
    return l.printer.PrintLoadBalancerHealthCheck(loadBalancer)
}

// update logic here
```

cli/command/load_balancer.go
```diff
@@ -10,6 +18,497 @@ func NewLoadBalancerCommand(b *CommandBase) *LoadBalancerCommand {
...
+		l.printer.Printf("No health check update required. Displaying load balancer health check.\n")
```
zpatrick: Don't print this here since that would mess up someones script. 
This function is kind of odd in that it is both a read and an update - maybe we should rethink that ux?

cli/command/load_balancer.go
```diff
@@ -10,6 +18,497 @@ func NewLoadBalancerCommand(b *CommandBase) *LoadBalancerCommand {
...
+	return l.printer.PrintLoadBalancerSummaries(loadBalancerSummaries...)
+}
+
+func (l *LoadBalancerCommand) read(c *cli.Context) error {
+	args, err := extractArgs(c.Args(), "LOADBALANCER_NAME")
+	if err != nil {
+		return err
+	}
+
+	loadBalancer, err := l.client.ReadLoadBalancer(args["LOADBALANCER_NAME"])
```
zpatrick: This needs to follow the proper pattern for `read`. See [tlake's pr](https://github.com/quintilesims/layer0/pull/386/files#diff-9e6f000570872e0a4ac06e2e2e4ba195R177)

cli/command/load_balancer.go
```diff
@@ -10,6 +18,497 @@ func NewLoadBalancerCommand(b *CommandBase) *LoadBalancerCommand {
...
+
+	containerPort, err := strconv.ParseInt(split[1], 10, 64)
+	if err != nil {
+		return nil, fmt.Errorf("'%s' is not a valid integer", split[1])
+	}
+
+	protocol := split[2]
+	if strings.ToLower(protocol) != "https" {
+		if certificateName != "" {
+			fmt.Printf("Warning: Certificate specified, but protocol of port is not HTTPS. Certificate will not be used.\n")
```
zpatrick: I think this should actually be the inverse logic: we should check if the user specified `https` but the certificate is `""` and throw an error in that scenario. 

The reason I don't think we need this block is because you can only specify a certificate once, but can have multiple ports. You would always get warning if you did:
```
l0 loadbalancer create --certificate foo --port 443:443/https --port 80:80/http bar baz
``` 
even though that's a perfectly valid command.  

If you feel strongly we should keep this check, then it should be in a debug statement `log.Printf("[DEBUG] ...")` and specify which port is the culprit.

cli/command/load_balancer_test.go
zpatrick: didn't realize this was in `TestAddPort`. In that case, I think having any sort of `healthcheck` in this test is a sort of red herring - should be removed IMO

cli/command/load_balancer_test.go
```diff
@@ -0,0 +1,491 @@
...
+		Return(loadBalancer, nil)
+
+	ports = append(ports, models.Port{
+		ContainerPort: 81,
+		HostPort:      81,
+		Protocol:      "tcp",
+	})
+
+	req := models.UpdateLoadBalancerRequest{
+		HealthCheck:    nil,
```
zpatrick: [nit] no need to specify a field if it is just going to be its zero value

cli/command/load_balancer_test.go
```diff
@@ -0,0 +1,491 @@
...
+		Return([]string{"eid"}, nil)
+
+	ports := []models.Port{
+		models.Port{
+			ContainerPort: 80,
+			HostPort:      80,
+			Protocol:      "tcp",
+		},
+	}
+	healthCheck := models.HealthCheck{
```
zpatrick: [nit] newline after closing bracket

cli/command/load_balancer_test.go
```diff
@@ -0,0 +1,491 @@
...
+		Return(job, nil)
+
+	loadBalancer := &models.LoadBalancer{
+		HealthCheck: healthCheck,
+		Ports:       ports,
+	}
+
+	base.Client.EXPECT().
+		ReadLoadBalancer("lid").
+		Return(loadBalancer, nil)
```
zpatrick: [nit] this could just be `Return(&models.LoadBalancer{}, nil)`

...perhaps I'm thinking too much into this, but as someone reading this test it would be very clear to me that `Return(&models.LoadBalancer{}, nil)` means "it doesn't really matter what gets returned in this call" vs. right now it looks like having the `HealthCheck` and `Ports` field set is important.

cli/command/load_balancer_test.go
```diff
@@ -0,0 +1,491 @@
...
+			HostPort:      80,
+			Protocol:      "tcp",
+		},
+		models.Port{
+			ContainerPort: 81,
+			HostPort:      81,
+			Protocol:      "tcp",
+		},
+	}
+	healthCheck := models.HealthCheck{}
```
zpatrick: ditto as AddPort

cli/command/load_balancer_test.go
```diff
@@ -0,0 +1,491 @@
...
+
+func TestUpdateHealthCheck(t *testing.T) {
+	base, ctrl := newTestCommand(t)
+	defer ctrl.Finish()
+
+	base.Resolver.EXPECT().
+		Resolve("load_balancer", "lname").
+		Return([]string{"lid"}, nil)
+
+	ports := []models.Port{}
```
zpatrick: ditto: we don't care about ports in this test. The field should be `nil` and ignored for this test

cli/command/load_balancer_test.go
```diff
@@ -0,0 +1,491 @@
...
+		"Non-int '--healthcheck-interval' flag": getCLIContext(t, args{"lname"}, flags{"healthcheck-interval": "two"}),
+		"Non-int '--healthcheck-timeout' flag":  getCLIContext(t, args{"lname"}, flags{"healthcheck-timeout": "two"}),
+		"Non-int '--healthy-threshold' flag":    getCLIContext(t, args{"lname"}, flags{"healthy-threshold": "two"}),
+		"Non-int '--unhealthy-threshold' flag":  getCLIContext(t, args{"lname"}, flags{"unhealthy-threshold": "two"}),
+		"Missing NAME arg":                      getCLIContext(t, nil, flags{"interval": 2}),
+	}
+
+	loadBalancerCommand := NewLoadBalancerCommand(base.Command())
+	for name, c := range contexts {
+		if err := loadBalancerCommand.healthcheck(c); err == nil {
```
zpatrick: We're using the `t.Run` function to do table tests now. It gives us a bunch of cool benefits - we didn't know about it before. So this would be:
```go
    t.Run(name, func(t *testing.T) {
        if err := loadBalancerCommand...
   })
}
```

cli/command/load_balancer_test.go
```diff
@@ -0,0 +1,491 @@
...
+	if err := loadBalancerCommand.list(c); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestReadLoadBalancer(t *testing.T) {
+	base, ctrl := newTestCommand(t)
+	defer ctrl.Finish()
+
+	base.Client.EXPECT().
```
zpatrick: This will need to be updated - needs to resolve a slice of ids

cli/command/load_balancer_test.go
```diff
@@ -0,0 +1,491 @@
...
+		},
+	}
+
+	for input, expected := range cases {
+		model, err := parsePort(input, "cert_name")
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if !reflect.DeepEqual(model, expected) {
```
zpatrick: use the `assert` package for this:  `assert.Equal(t, expected, result)`

cli/command/load_balancer_test.go
```diff
@@ -0,0 +1,491 @@
...
+		"Missing HOST_PORT":          ":80/tcp",
+		"Missing CONTAINER_PORT":     "80:/tcp",
+		"Missing PROTOCOL":           "80:80",
+		"Non-integer HOST_PORT":      "80p:80/tcp",
+		"Non-integer CONTAINER_PORT": "80:80p/tcp",
+	}
+
+	for name, input := range cases {
+		if _, err := parsePort(input, ""); err == nil {
+			t.Fatalf("%s: error was nil!", name)
```
zpatrick: use `t.Run` for this

cli/command/load_balancer.go
```diff
@@ -10,6 +18,497 @@ func NewLoadBalancerCommand(b *CommandBase) *LoadBalancerCommand {
...
+	if err != nil {
+		return err
+	}
+
+	loadBalancer, err := l.client.ReadLoadBalancer(loadBalancerID)
+	if err != nil {
+		return err
+	}
+
+	loadBalancer.Ports = append(loadBalancer.Ports, *port)
```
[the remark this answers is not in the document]
jparsons04: If `models.UpdateLoadBalancerRequest`'s `Ports` field is a `*[]models.Port`, how would I capture the address of the return of `append()` without first assigning it to a variable first?

cli/command/load_balancer.go
```diff
@@ -10,6 +18,497 @@ func NewLoadBalancerCommand(b *CommandBase) *LoadBalancerCommand {
...
+		interval, err := strconv.Atoi(interval)
+		if err != nil {
+			return err
+		}
+
+		requiresUpdate = true
+		healthCheck.Interval = interval
+	}
+
+	if timeout := c.String("healthcheck-timeout"); timeout != "" {
```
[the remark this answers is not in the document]
jparsons04: How do we check for an invalid input then given:
```
func TestUpdateHealthCheck_UserInputError(t *testing.T) {
     defer ctrl.Finish()
 
     contexts := map[string]*cli.Context{
         "Non-int '--healthcheck-interval' flag": getCLIContext(t, args{"lname"}, flags{"healthcheck-interval": "two"}),
         "Non-int '--healthcheck-timeout' flag":  getCLIContext(t, args{"lname"}, flags{"healthcheck-timeout": "two"}),
         "Non-int '--healthy-threshold' flag":    getCLIContext(t, args{"lname"}, flags{"healthy-threshold": "two"}),
         "Non-int '--unhealthy-threshold' flag":  getCLIContext(t, args{"lname"}, flags{"unhealthy-threshold": "two"}),
         "Missing NAME arg":                      getCLIContext(t, nil, flags{"interval": 2}),
     }
 
     loadBalancerCommand := NewLoadBalancerCommand(base.Command())
         if err := loadBalancerCommand.healthcheck(c); err == nil {
             t.Fatalf("%s: error was nil!", name)
         }
     }
 }
```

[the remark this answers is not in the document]
jparsons04: I ask this because c.Int() would return only an int (in this case, 0) if the value isn't found (as opposed to being able to return a value and an `error`.

cli/command/load_balancer.go
jparsons04: This was there because we were running into an issue where the spinner message would not print a new line after completing the job, and displaying the results. So we were forcing l.printer.StopSpinner() before calling the printer's method.

cli/command/load_balancer.go
jparsons04: This was fleshed out explicitly mainly because of what the help message was prescribing:

https://github.com/quintilesims/layer0/blob/develop/cli/command/load_balancer_command.go#L108
`view or update the health check for a load balancer`

I think there could be room for discussion about how to make both of these options available for the user.

cli/command/load_balancer.go
jparsons04: I missed that case: that a user can specify multiple `--port` arguments in a single command. I'm going to just implement the error logic in that case rather than the warning logic.

cli/command/load_balancer.go
```diff
@@ -10,6 +18,497 @@ func NewLoadBalancerCommand(b *CommandBase) *LoadBalancerCommand {
...
+			},
+			{
+				Name:      "create",
+				Usage:     "Create a new load balancer LOADBALANCER_NAME in the environment specified in ENVIRONMENT_TARGET",
+				Action:    l.create,
+				ArgsUsage: "ENVIRONMENT_TARGET LOADBALANCER_NAME",
+				Flags: []cli.Flag{
+					cli.StringSliceFlag{
+						Name:  "port",
+						Usage: "Port configuration in format 'HOST_PORT:CONTAINER_PORT/PROTOCOL' (default 80:80/tcp)",
```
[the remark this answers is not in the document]
jparsons04: `Value` is a `*cli.StringSlice`, so I would need to somehow pass the address of this literal into `Value`, which I cannot do directly.

cli/command/load_balancer.go
```diff
@@ -10,6 +18,497 @@ func NewLoadBalancerCommand(b *CommandBase) *LoadBalancerCommand {
...
+	})
+}
+
+func (l *LoadBalancerCommand) dropport(c *cli.Context) error {
+	args, err := extractArgs(c.Args(), "LOADBALANCER_NAME", "HOST_PORT")
+	if err != nil {
+		return err
+	}
+
+	port, err := strconv.ParseInt(args["HOST_PORT"], 10, 64)
```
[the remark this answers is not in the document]
jparsons04: :white_check_mark:

cli/command/load_balancer.go
```diff
@@ -10,6 +18,497 @@ func NewLoadBalancerCommand(b *CommandBase) *LoadBalancerCommand {
...
+	loadBalancer, err := l.client.ReadLoadBalancer(loadBalancerID)
+	if err != nil {
+		return err
+	}
+
+	var exists bool
+	for i, p := range loadBalancer.Ports {
+		if p.HostPort == port {
+			loadBalancer.Ports = append(loadBalancer.Ports[:i], loadBalancer.Ports[i+1:]...)
+			exists = true
```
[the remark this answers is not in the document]
jparsons04: :white_check_mark:

cli/command/load_balancer.go
jparsons04: :white_check_mark:

cli/command/load_balancer.go
```diff
@@ -10,6 +18,497 @@ func NewLoadBalancerCommand(b *CommandBase) *LoadBalancerCommand {
...
+			loadBalancer.HealthCheck.HealthyThreshold = healthCheck.HealthyThreshold
+		}
+
+		if healthCheck.UnhealthyThreshold != 0 {
+			loadBalancer.HealthCheck.UnhealthyThreshold = healthCheck.UnhealthyThreshold
+		}
+
+		req := models.UpdateLoadBalancerRequest{
+			LoadBalancerID: loadBalancerID,
+			Ports:          &loadBalancer.Ports,
```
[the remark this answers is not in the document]
jparsons04: :white_check_mark:

cli/command/load_balancer.go
jparsons04: :white_check_mark:

cli/command/load_balancer.go
jparsons04: :white_check_mark:

cli/command/load_balancer_test.go
jparsons04: :white_check_mark:

cli/command/load_balancer_test.go
jparsons04: :check:

cli/command/load_balancer_test.go
jparsons04: :white_check_mark:

cli/command/load_balancer_test.go
jparsons04: :white_check_mark:

cli/command/load_balancer_test.go
jparsons04: So here, I need to split this into two tests, because the missing name exits immediately, while the flag analysis happens after the load balancer entity is resolved and the load balancer is read.

jparsons04: Actually, I figured out a way to test this in one test by adding some checking logic up front in the `healthcheck()` method

cli/command/load_balancer.go
jparsons04: It is just for validation in this case.

cli/command/load_balancer_test.go
jparsons04: :white_check_mark:

cli/command/load_balancer_test.go
jparsons04: Ended up just removing this entirely.

cli/command/load_balancer.go
```diff
@@ -10,6 +17,449 @@ func NewLoadBalancerCommand(b *CommandBase) *LoadBalancerCommand {
...
+		port, err := parsePort(p, c.String("certificate"))
+		if err != nil {
+			return err
+		}
+
+		ports = append(ports, *port)
+	}
+
+	if c.String("healthcheck-target") != "" {
+		if _, err := parseTarget(c.String("healthcheck-target")); err != nil {
```
zpatrick: [nit] probably more appropriate to name this `validateTarget` and change the signature to only return an error

cli/command/load_balancer.go
```diff
@@ -10,6 +17,449 @@ func NewLoadBalancerCommand(b *CommandBase) *LoadBalancerCommand {
...
+
+		l.printer.StopSpinner()
+		return l.printer.PrintLoadBalancers(loadBalancer)
+	}
+
+	return l.waitOnJobHelper(c, jobID, "Creating Load Balancer", onCompleteFn)
+}
+
+func (l *LoadBalancerCommand) delete(c *cli.Context) error {
+	return l.deleteHelper(c, "loadbalancer", func(loadBalancerID string) (string, error) {
```
zpatrick: the `type` needs to be `load_balancer` when resolving

cli/command/load_balancer.go
jparsons04: :white_check_mark:

cli/command/load_balancer.go
jparsons04: :white_check_mark:

cli/command/load_balancer.go
```diff
@@ -10,6 +17,447 @@ func NewLoadBalancerCommand(b *CommandBase) *LoadBalancerCommand {
...
+		loadBalancer, err := l.client.ReadLoadBalancer(loadBalancerID)
+		if err != nil {
+			return err
+		}
+
+		l.printer.StopSpinner()
+		return l.printer.PrintLoadBalancers(loadBalancer)
+	}
+
+	return l.waitOnJobHelper(c, jobID, "adding port", onCompleteFN)
```
diemonster: [nit] personally find defining the function `onCompleteFN()` as an anonymous function within the call to `waitOnJobHelper()` to be more readable

cli/command/load_balancer.go
```diff
@@ -10,6 +17,447 @@ func NewLoadBalancerCommand(b *CommandBase) *LoadBalancerCommand {
...
+			},
+			{
+				Name:      "get",
+				Usage:     "Describe load balancer LOADBALANCER_NAME",
+				Action:    l.read,
+				ArgsUsage: "LOADBALANCER_NAME",
+			},
+			{
+				Name:      "healthcheck",
+				Usage:     "View or update the health check of load balancer LOADBALANCER_NAME. If no options are specified, the health check is displayed.",
```
diemonster: [nit] this line wraps oddly for your average terminal screen, consider shorting the line. I think the 'if' statement is a bit redundant.

PR #396, opened 2017-10-12
Added mock resolver and base test for CLI

sesh-kebab:
Adds base test and a mock resolver which would be common across all command tests.


scripts/Makefile
```diff
@@ -19,6 +20,9 @@ job:
 client:
 	mockgen github.com/quintilesims/layer0/client Client > ../client/mock_client/mock_client.go &
 
+command:
```
zpatrick: command is subset of cli

PR #398, opened 2017-10-12
add GetCLIContext func

diemonster:
Includes the former `testutils` function `getCLIContext`, which is used for unit testing `command`

### DIFF
--- a/cli/command/test_helpers.go
+++ b/cli/command/test_helpers.go
@@ -1,0 +1,37 @@
+package command
+
+import (
+	"flag"
+	"testing"
+	"time"
+
+	"github.com/urfave/cli"
+)
+
+var TEST_TIMEOUT = time.Minute * 15
+
+func GetCLIContext(t *testing.T, args []string, flags map[string]interface{}) *cli.Context {
+	flagSet := &flag.FlagSet{}
+
+	for key, val := range flags {
+		switch v := val.(type) {
+		case bool:
+			flagSet.Bool(key, v, "")
+		case string:
+			flagSet.String(key, v, "")
+		case []string:
+			slice := cli.StringSlice(v)
+			flagSet.Var(&slice, key, "")
+		case int:
+			flagSet.Int(key, v, "")
+		default:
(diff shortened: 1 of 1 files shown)


cli/command/test_helpers.go
```diff
@@ -0,0 +1,37 @@
...
+			flagSet.Var(&slice, key, "")
+		case int:
+			flagSet.Int(key, v, "")
+		default:
+			t.Errorf("Cannot generate CLI context: unknown flag type for '%s'", key)
+		}
+	}
+
+	// add default global flags
+	flagSet.String("output", "text", "")
```
zpatrick: this should use the consts in the `config` package. I think it's something like:
```
flagSet.String(config.FLAG_TIMEOUT, "15m", "")
```

diemonster: yep good guess ;). 

What's the point of `flagSet.String("output", "text", "")`?

zpatrick: `l0 -output=json|text ...`

diemonster: ✅

cli/command/test_helpers.go
```diff
@@ -0,0 +1,36 @@
...
```
zpatrick: should do it with `output` too

cli/command/test_helpers.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"flag"
+	"testing"
+	"time"
+
+	"github.com/quintilesims/layer0/common/config"
+	"github.com/urfave/cli"
+)
+
+func GetCLIContext(t *testing.T, args []string, flags map[string]interface{}) *cli.Context {
```
zpatrick: should be lowercase

cli/command/test_helpers.go
diemonster: ala 	`flagSet.String(config.FLAG_OUTPUT, "text", "")`?

cli/command/test_helpers.go
diemonster: ✅

PR #400, opened 2017-10-12
move base_test to test_helpers

diemonster:
consolidate the code from `base_test.go` to `test_helpers.go`, as placing all test utils there makes more sense


sesh-kebab:
There's an incoming change from zpatrick that uppercases all the properties e.g. client -> Client etc. If that is how we want it, might be a good chance to update the casing in this PR.

```
type TestCommandBase struct {		
 	client   *mock_client.MockClient		
 	printer  *printer.TestPrinter		
 	resolver *mock_resolver.MockResolver		
 }
```

diemonster:
@sesh-kebab I'm fine with changing, although I'm not clear on why? Are other types expected to need access to those fields?

zpatrick:
@diemonster yes, other types are expected to use those fields

diemonster:
@zpatrick @sesh-kebab Hmmm ok. That prevents merging this for the moment then as changing `CommandBase` has cascading effects to everyone else.

PR #401, opened 2017-10-16
Environment link api

sesh-kebab:
Adds implementation for Environment linking

Via swagger endpoint

**Checklist**
- [x] Unit tests


api/controllers/swagger.go
```diff
@@ -171,6 +187,36 @@ func (s *SwaggerController) ServeSwaggerSpec(c *fireball.Context) (fireball.Resp
 					},
 				},
 			},
+			"/environment/{id}/link": map[string]swagger.Method{
+				"post": {
+					Summary: "Create an Environment Link",
+					Tags:    []string{"Environment"},
+					Parameters: []swagger.Parameter{
+						swagger.NewStringPathParam("id", "ID of the source environment", true),
```
zpatrick: [nit] should we keep `source_id` here to map to `source_id` in delete?

api/provider/aws/environment_link.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return err
+	}
+
+	destGroup, err := readSG(e.AWS.EC2, getEnvironmentSGName(fqDestEnvID))
+	if err != nil {
+		return err
+	}
+
+	createIngressInput := func(sourceGroupID, destGroupID *string) *ec2.AuthorizeSecurityGroupIngressInput {
```
zpatrick: we should match the existing patterns; this should be a helper function that takes source/dest as params.

api/provider/aws/environment_link.go
```diff
@@ -0,0 +1,153 @@
...
+			},
+		}
+	}
+
+	_, err = e.AWS.EC2.AuthorizeSecurityGroupIngress(createIngressInput(sourceGroup.GroupId, destGroup.GroupId))
+	if err != nil {
+		return err
+	}
+
+	_, err = e.AWS.EC2.AuthorizeSecurityGroupIngress(createIngressInput(destGroup.GroupId, sourceGroup.GroupId))
```
zpatrick: keep errors in smallest scope: `if _, err := x(); err != nil { ... }`

api/provider/aws/environment_link.go
```diff
@@ -0,0 +1,153 @@
...
+	for _, tag := range tags {
+		if err := e.TagStore.Insert(tag); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (e *EnvironmentProvider) Unlink(sourceEnvironmentID, destEnvironmentID string) error {
```
zpatrick: Link and Unlink should be in separate files

api/provider/aws/environment_link.go
```diff
@@ -0,0 +1,153 @@
...
+	fqSourceEnvID := addLayer0Prefix(e.Config.Instance(), sourceEnvironmentID)
+	fqDestEnvID := addLayer0Prefix(e.Config.Instance(), destEnvironmentID)
+
+	sourceGroup, err := readSG(e.AWS.EC2, getEnvironmentSGName(fqSourceEnvID))
+	if err != nil {
+		return err
+	}
+
+	if sourceGroup == nil {
```
zpatrick: This path will not execute. `readSG` will return an error if the sg doesn't exist. You need to capture and handle that error above.

api/provider/aws/environment_link.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return err
+	}
+
+	if destGroup == nil {
+		//log.Warnf("Skipping environment unlink since security group '%s' does not exist", getEnvironmentSGName(fqDestEnvID))
+		return nil
+	}
+
+	removeIngressRule := func(group *ec2.SecurityGroup, groupIDToRemove string) error {
```
zpatrick: ditto helper function

api/provider/aws/environment_link.go
```diff
@@ -0,0 +1,153 @@
...
+
+	if err := removeIngressRule(sourceGroup, *destGroup.GroupId); err != nil {
+		return err
+	}
+
+	if err := removeIngressRule(destGroup, *sourceGroup.GroupId); err != nil {
+		return err
+	}
+
+	deleteLinkTag := func(sourceEnvironmentID, destEnvironmentID string) error {
```
zpatrick: ditto helper function

api/provider/aws/environment_link.go
```diff
@@ -0,0 +1,153 @@
...
+		return err
+	}
+
+	destGroup, err := readSG(e.AWS.EC2, getEnvironmentSGName(fqDestEnvID))
+	if err != nil {
+		return err
+	}
+
+	createIngressInput := func(sourceGroupID, destGroupID *string) *ec2.AuthorizeSecurityGroupIngressInput {
+		return &ec2.AuthorizeSecurityGroupIngressInput{
```
zpatrick: please follow existing patterns for aws object creation/validation/calls

api/provider/aws/environment_link.go
```diff
@@ -0,0 +1,153 @@
...
+	if destGroup == nil {
+		//log.Warnf("Skipping environment unlink since security group '%s' does not exist", getEnvironmentSGName(fqDestEnvID))
+		return nil
+	}
+
+	removeIngressRule := func(group *ec2.SecurityGroup, groupIDToRemove string) error {
+		for _, permission := range group.IpPermissions {
+			for _, pair := range permission.UserIdGroupPairs {
+				if *pair.GroupId == groupIDToRemove {
+					groupPermission := &ec2.IpPermission{
```
zpatrick: ditto: please follow existing patterns for aws object creation/validation/calls
also, always safe de-reference pointers: `aws.StringValue(pair.GroupId)`

api/controllers/swagger.go
sesh-kebab: ✅

api/provider/aws/environment_link.go
sesh-kebab: ✅

api/provider/aws/environment_link.go
sesh-kebab: ✅

api/provider/aws/environment_link.go
sesh-kebab: ✅

api/provider/aws/environment_link.go
sesh-kebab: ✅

api/provider/aws/environment_link.go
sesh-kebab: This is code from the existing implementation. Wasn't sure if we wanted to still log an error or if returning the original error should suffice.

https://github.com/quintilesims/layer0/blob/develop/api/backend/ecs/environment_manager.go#L370-L388

api/provider/aws/environment_link.go
sesh-kebab: ✅

api/provider/aws/environment_link.go
sesh-kebab: ✅ (issue from existing implementation)

api/provider/aws/environment_link.go
sesh-kebab: ✅

api/provider/aws/environment_link.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		return err
+	}
+
+	destGroup, err := readSG(e.AWS.EC2, getEnvironmentSGName(fqDestEnvID))
+	if err != nil {
+		return err
+	}
+
+	ingressInput := e.createIngressInput(sourceGroup.GroupId, destGroup.GroupId)
```
zpatrick: The current pattern is wrapping all aws calls in helpers with more sensible param types. 
Please refer to the [rules of thumb page](https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/README.md)

 So this code would look something like:
```
sourceGroup, err := ...
destGroup, err := ...

sourceGroupID := aws.StringValue(sourceGroup.GroupId)
destGroupID := aws.StringValue(destGroup.GroupId)
if err := e.createIngressInput(sourceGroupID, destGroupID); err != nil { ... }

if err := e.createIngressInput(destrGroupID, sourceGroupID); err != nil { ... }
```

api/provider/aws/environment_link.go
```diff
@@ -0,0 +1,71 @@
...
+	if _, err := e.AWS.EC2.AuthorizeSecurityGroupIngress(ingressInput); err != nil {
+		return err
+	}
+
+	ingressInput = e.createIngressInput(destGroup.GroupId, sourceGroup.GroupId)
+	if _, err := e.AWS.EC2.AuthorizeSecurityGroupIngress(ingressInput); err != nil {
+		return err
+	}
+
+	tags := models.Tags{
```
zpatrick: insert tags should be in helper

api/provider/aws/environment_link.go
```diff
@@ -0,0 +1,71 @@
...
+		if err := e.TagStore.Insert(tag); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (e *EnvironmentProvider) createIngressInput(sourceGroupID, destGroupID *string) *ec2.AuthorizeSecurityGroupIngressInput {
+	return &ec2.AuthorizeSecurityGroupIngressInput{
```
zpatrick: good example on how we generate aws-sdk-go objects to be more readable: https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/environment_create.go#L111

zpatrick: also, don't forget to call `Validate()` on the request object

api/provider/aws/environment_unlink.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+)
+
+func (e *EnvironmentProvider) Unlink(sourceEnvironmentID, destEnvironmentID string) error {
+	fqSourceEnvID := addLayer0Prefix(e.Config.Instance(), sourceEnvironmentID)
+	fqDestEnvID := addLayer0Prefix(e.Config.Instance(), destEnvironmentID)
+
+	sourceGroup, err := readSG(e.AWS.EC2, getEnvironmentSGName(fqSourceEnvID))
+	if err != nil {
```
zpatrick: delete operations are idempotent; you'll need to catch the 'does not exist' error and return nil in that scenario

api/provider/aws/environment_unlink.go
```diff
@@ -0,0 +1,81 @@
...
+		//todo: not sure if we want to still log these warning messages here
+		//log.Warnf("Skipping environment unlink since security group '%s' does not exist", getEnvironmentSGName(fqDestEnvID))
+		return err
+	}
+
+	destGroup, err := readSG(e.AWS.EC2, getEnvironmentSGName(fqDestEnvID))
```
zpatrick: ditto - idempotency

api/provider/aws/environment_unlink.go
```diff
@@ -0,0 +1,81 @@
...
+				groupPermission := &ec2.IpPermission{
+					IpProtocol:       permission.IpProtocol,
+					UserIdGroupPairs: []*ec2.UserIdGroupPair{pair},
+				}
+
+				input := &ec2.RevokeSecurityGroupIngressInput{
+					GroupId:       group.GroupId,
+					IpPermissions: []*ec2.IpPermission{groupPermission},
+				}
+				_, err := e.AWS.EC2.RevokeSecurityGroupIngress(input)
```
zpatrick: `if _, err := ...`

api/provider/aws/environment_unlink.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func (e *EnvironmentProvider) deleteLinkTag(sourceEnvironmentID, destEnvironmentID string) error {
+	tags, err := e.TagStore.SelectByTypeAndID("environment", sourceEnvironmentID)
+	if err != nil {
+		return err
+	}
+
+	for _, tag := range tags.WithKey("link").WithValue(destEnvironmentID) {
+		if err := e.TagStore.Delete(tag.EntityType, tag.EntityID, tag.Key); err != nil {
```
zpatrick: we can call this function directly w/o doing the select above - we already know the type, id, and key for the tag

api/provider/aws/environment_link.go
sesh-kebab: ✅

api/provider/aws/environment_link.go
sesh-kebab: Validate at the controller layer right? If so 🆗

api/provider/aws/environment_unlink.go
sesh-kebab: ✅

api/provider/aws/environment_unlink.go
sesh-kebab: ✅

api/provider/aws/environment_unlink.go
sesh-kebab: ✅

api/provider/aws/environment_unlink.go
```diff
@@ -0,0 +1,86 @@
...
+	if err := models.EnvironmentLinkRequest(req).Validate(); err != nil {
+		return err
+	}
+
+	fqSourceEnvID := addLayer0Prefix(e.Config.Instance(), req.SourceEnvironmentID)
+	fqDestEnvID := addLayer0Prefix(e.Config.Instance(), req.DestEnvironmentID)
+
+	sourceSecurityGroup, err := readSG(e.AWS.EC2, getEnvironmentSGName(fqSourceEnvID))
+	if err != nil {
+		log.Printf("[WARN] skipping environment unlink since security group '%s' does not exist\n", getEnvironmentSGName(fqDestEnvID))
```
zpatrick: we need to catch the “does not exist” error explicitly here and below

api/provider/aws/environment_unlink.go
```diff
@@ -0,0 +1,86 @@
...
+		return nil
+	}
+
+	if err := e.removeIngressRule(sourceSecurityGroup, *destSecurityGroup.GroupId); err != nil {
+		return err
+	}
+
+	if err := e.removeIngressRule(destSecurityGroup, *sourceSecurityGroup.GroupId); err != nil {
```
zpatrick: Need to safe dereference

cli/command/base_test.go
```diff
@@ -0,0 +1,36 @@
+package command
```
zpatrick: can remove this file

cli/command/environment.go
```diff
@@ -13,5 +18,251 @@ func NewEnvironmentCommand(b *CommandBase) *EnvironmentCommand {
...
+	return e.printer.PrintEnvironmentSummaries(environmentSummaries...)
+}
+
+func (e *EnvironmentCommand) read(c *cli.Context) error {
+	args, err := extractArgs(c.Args(), "NAME")
+	if err != nil {
+		return err
+	}
+
+	entityIDs, err := e.resolver.Resolve("environment", args["NAME"])
```
zpatrick: [nit] environmentIDs

cli/command/environment.go
```diff
@@ -13,5 +18,251 @@ func NewEnvironmentCommand(b *CommandBase) *EnvironmentCommand {
...
+	args, err := extractArgs(c.Args(), "NAME", "COUNT")
+	if err != nil {
+		return err
+	}
+
+	count, err := strconv.Atoi(args["COUNT"])
+	if err != nil {
+		return fmt.Errorf("'%s' is not a valid integer", args["COUNT"])
+	}
+	minClusterCount := int(count)
```
zpatrick: Don’t need this var, count us already an int

api/provider/aws/environment_link.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	destGroup, err := readSG(e.AWS.EC2, getEnvironmentSGName(fqDestEnvID))
+	if err != nil {
+		return err
+	}
+
+	sourceGroupID := sourceGroup.GroupId
+	destGroupID := destGroup.GroupId
+	ingressInput := e.createIngressInput(sourceGroupID, destGroupID)
```
zpatrick: [nit] this is still breaking the ROT pattern(s)

cli/command/environment.go
```diff
@@ -13,5 +18,251 @@ func NewEnvironmentCommand(b *CommandBase) *EnvironmentCommand {
...
+	id2, err := e.resolveSingleEntityIDHelper("environment", args["DESTINATION"])
+	if err != nil {
+		return err
+	}
+
+	if id1 == id2 {
+		return fmt.Errorf("Cannot link an environment to itself")
+	}
+
+	if err := e.client.CreateLink(id1, id2); err != nil {
```
zpatrick: Need to update for job model

cli/command/environment.go
```diff
@@ -13,5 +18,251 @@ func NewEnvironmentCommand(b *CommandBase) *EnvironmentCommand {
...
+			return err
+		}
+
+		environments[i] = environment
+	}
+
+	return e.printer.PrintEnvironments(environments...)
+}
+
+func (e *EnvironmentCommand) update(c *cli.Context) error {
```
zpatrick: This function is a duplicate of setMinCount

cli/command/environment_test.go
```diff
@@ -0,0 +1,381 @@
...
+	testCases := []struct {
+		name    string
+		command func(*cli.Context) error
+		args    []string
+	}{
+		{
+			name:    "create",
+			command: command.create,
+		},
+		{
```
zpatrick: These are many test cases missing here: one for each missing arg per command, another for invalid ints,  etc.

zpatrick: Also, the rest of the open prs have one test per command instead of a single one for all

cli/command/environment_test.go
```diff
@@ -0,0 +1,381 @@
...
+			c := GetCLIContext(t, tc.args, nil)
+
+			if err := tc.command(c); err == nil {
+				t.Fatalf("%s: error was nil!", tc.name)
+			}
+		})
+	}
+}
+
+func TestCreateEnvironment(t *testing.T) {
```
zpatrick: why not use the wait test helper?

cli/command/environment_test.go
```diff
@@ -0,0 +1,381 @@
...
+		Printer:  &printer.TestPrinter{},
+		Resolver: mock_resolver.NewMockResolver(ctrl),
+	}
+
+	envCmd := NewEnvironmentCommand(tc.Command())
+
+	return tc.Client, tc.Resolver, envCmd, ctrl
+}
+
+func tempFile(t *testing.T, content string) (*os.File, func()) {
```
zpatrick: this should exist in a common helper file

client/environment.go
```diff
@@ -52,3 +52,21 @@ func (c *APIClient) UpdateEnvironment(req models.UpdateEnvironmentRequest) (stri
 
 	return resp.JobID, nil
 }
+
+func (c *APIClient) CreateLink(req models.CreateEnvironmentLinkRequest) error {
+	path := fmt.Sprintf("/environment/%s/link", req.SourceEnvironmentID)
```
zpatrick: Shouldn’t we change the path since the source is in the request body now?

api/controllers/environment.go
```diff
@@ -39,6 +39,18 @@ func (e *EnvironmentController) Routes() []*fireball.Route {
 				"DELETE": e.DeleteEnvironment,
 			},
 		},
+		{
+			Path: "/environment/:id/link",
```
zpatrick: This path seems odd given the two ids exist in the request body. maybe `link` should be its own path, like `POST /link` and `DELETE /link`? I'm not sure.

api/provider/aws/environment_unlink.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/aws/aws-sdk-go/aws/awserr"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/quintilesims/layer0/common/models"
+)
+
+func (e *EnvironmentProvider) Unlink(req models.DeleteEnvironmentLinkRequest) error {
+	log.Printf("[DEBUG] unlink called")
```
zpatrick: this look like a leftover that should be deleted?

api/provider/aws/environment_link.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	if err := e.createLinkTags(req.SourceEnvironmentID, req.DestEnvironmentID); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (e *EnvironmentProvider) createIngressInput(sourceGroupID, destGroupID *string) error {
```
zpatrick: [nit] these args should be `string` instead of `*string` to match your delete helper and ROT

api/provider/aws/environment_unlink.go
```diff
@@ -0,0 +1,100 @@
...
+				groupPermission := &ec2.IpPermission{
+					IpProtocol:       permission.IpProtocol,
+					UserIdGroupPairs: []*ec2.UserIdGroupPair{pair},
+				}
+
+				input := &ec2.RevokeSecurityGroupIngressInput{
+					GroupId:       group.GroupId,
+					IpPermissions: []*ec2.IpPermission{groupPermission},
+				}
+				if _, err := e.AWS.EC2.RevokeSecurityGroupIngress(input); err != nil {
```
zpatrick: [nit] newline closing bracket

api/provider/aws/environment_unlink.go
```diff
@@ -0,0 +1,100 @@
...
+
+	if err := e.deleteLinkTag(req.DestEnvironmentID, req.SourceEnvironmentID); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (e *EnvironmentProvider) removeIngressRule(group *ec2.SecurityGroup, groupIDToRemove string) error {
+	for _, permission := range group.IpPermissions {
```
zpatrick: this implementation seems heavy-handed. Since we know what the `IpPermission` will look like, can't we just call `RevokeSecurityGroupIngress` and catch idempotent errors? That would also make this easier to test (if we decide to unit test helpers... yet to see i guess)

api/controllers/environment.go
sesh-kebab: It is definitely odd. Redundant information is being passed. But it is consistent. For example, environment update also has `environment_id` in both the path and the body. I think this is a side effect of doing updates as jobs. We are using the same model to capture the job request as the http request body, so the entity id always exists.

If we had 2 models: one for the http put body and one for the update job, we could avoid this.

api/provider/aws/environment_unlink.go
zpatrick: Right, which I think should be handled by catching that error. 

I'm leaning towards the "shoot first ask questions later" approach on deletes to cut down on api/db calls. 
It won't make much of a difference in this example since `SelectByTypeAndID` is so cheap, so consider this a [nit] comment, but just wanted to explain my reasoning.

api/provider/aws/job_runner.go
```diff
@@ -152,6 +156,32 @@ func (r *JobRunner) createTask(jobID, request string) (string, error) {
...
+	var req models.DeleteEnvironmentLinkRequest
+	if err := json.Unmarshal([]byte(request), &req); err != nil {
+		return "", errors.New(errors.InvalidRequest, err)
+	}
+
+	if err := r.environmentProvider.Unlink(req); err != nil {
+		return "", err
+	}
+
+	return req.SourceEnvironmentID, nil
```
zpatrick: Any idea for better names than `Source` and `Dest`? The implied direction problem you brought up should probably be addressed in this pr.

cli/command/environment.go
```diff
@@ -215,21 +215,25 @@ func (e *EnvironmentCommand) link(c *cli.Context) error {
...
-	if id1 == id2 {
+	if sourceEnvironmentID == destEnvironmentID {
 		return fmt.Errorf("Cannot link an environment to itself")
 	}
 
-	if err := e.client.CreateLink(id1, id2); err != nil {
+	req := models.CreateEnvironmentLinkRequest{
+		SourceEnvironmentID: sourceEnvironmentID,
+		DestEnvironmentID:   destEnvironmentID,
+	}
```
zpatrick: [nit] newline closing bracket

cli/command/environment.go
```diff
@@ -244,21 +248,25 @@ func (e *EnvironmentCommand) unlink(c *cli.Context) error {
...
+	if sourceEnvironmentID == destEnvironmentID {
 		return fmt.Errorf("Cannot unlink an environment from itself")
 	}
 
-	if err := e.client.DeleteLink(id1, id2); err != nil {
+	req := models.DeleteEnvironmentLinkRequest{
+		SourceEnvironmentID: sourceEnvironmentID,
+		DestEnvironmentID:   destEnvironmentID,
+	}
+	if err := e.client.DeleteLink(req); err != nil {
```
zpatrick: [nit] newline closing bracket

api/controllers/environment.go
zpatrick: different models is a good idea. We should do that all at once for each entity update though. Add it to that refactor ticket perhaps?

PR #409, opened 2017-10-30
Env unit tests

zpatrick:
Adds unit tests for environment entity-verbs

**Checklist**
- [ ] read
- [ ] list
- [ ] create
- [ ] delete
- [ ] update

Link is undergoing changes by @seshi and will be addressed in a future pr


api/provider/aws/test_aws/environment_list_test.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	for _, tag := range tags {
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	// ensure we filter out clusters that don't belong to our instance
+	listClusterPagesFN := func(input *ecs.ListClustersInput, fn func(output *ecs.ListClustersOutput, lastPage bool) bool) error {
```
diemonster: Is there value in even trying to make a generic pagination helper function for testing?

api/provider/aws/test_aws/environment_read_test.go
```diff
@@ -0,0 +1,104 @@
...
+		Return(describeLCOutput, nil)
+
+	target := provider.NewEnvironmentProvider(mockAWS.Client(), tagStore, mockConfig)
+	result, err := target.Read("env_id")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// todo: Links
+	expected := &models.Environment{
```
diemonster: Would it be impractical to define `expected` near the top of the test function and use fields from that model while setting up the rest of the test?

api/provider/aws/test_aws/environment_read_test.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	for _, tag := range tags {
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	// environments' security group names are <fq environment id>-env
+	describeSecurityGroupHelper(mockAWS, "l0-test-env_id-env", "sg_id")
```
diemonster: since `describeSecurityGroupHelper` is only used for tests, would it be an "improvement" for the function itself to just take `env_id` and concat the rest of the fq name?

api/provider/aws/test_aws/environment_read_test.go
zpatrick: It's definitely not impractical, but there is a school of thought that it's better to use hardcoded values over variables/references during unit tests (within reason, of course). Reason being: you actually _don't_ want it to be too easy to change your unit tests once they have been created - if someone introduces a breaking change, you want whoever did it to see all of the (potentially) cascading effects. I also tend to favor that approach because it aligns with Go's tendency towards explicitness. 

I don't feel very strong either way on the matter, but wanted to expose the logic behind why I wrote it in this form.

api/provider/aws/test_aws/environment_read_test.go
zpatrick: Doing that would either required we pass in a `instance string` param, or we would be introducing "magic" by assuming `instance`=`"test"`. I wouldn't consider either option an improvement. 

Also, I think it would be confusing to see a mix of `env_id` and `l0-test-env_id` during the "aws mocking" section of the tests.

api/provider/aws/test_aws/environment_list_test.go
diemonster: figured as much. This is easy enough to follow, i think.

api/provider/aws/test_aws/environment_read_test.go
diemonster: OK, that's a good argument; I was looking at it strictly from an ease-of-writing perspective. If the pattern sits well with everyone else, fine by me.

PR #410, opened 2017-10-31
274: Load Balancer provider/aws unit tests

jparsons04:
Adds unit tests for load balancer entity-actions

**Checklist**
- [x] create
- [x] delete
- [x] list
- [x] read
- [x] update

Work towards #274


api/provider/aws/aws_test/load_balancer_create_test.go
```diff
@@ -0,0 +1,335 @@
...
+
+	listeners := make([]*elb.Listener, len(req.Ports))
+	for i, port := range req.Ports {
+		listener := &elb.Listener{}
+		listener.SetProtocol(port.Protocol)
+		listener.SetLoadBalancerPort(port.HostPort)
+		listener.SetInstancePort(port.ContainerPort)
+
+		if port.CertificateName != "" {
+			serverCertificateMetadataList := []*iam.ServerCertificateMetadata{
```
zpatrick: [nit] this breaks the 
```go
input := ...
input.SetField(...)
```
pattern

api/provider/aws/aws_test/load_balancer_create_test.go
```diff
@@ -0,0 +1,335 @@
...
+	configureHealthCheckInput.SetLoadBalancerName("l0-test-lb_id")
+	configureHealthCheckInput.SetHealthCheck(healthCheck)
+
+	mockAWS.ELB.EXPECT().
+		ConfigureHealthCheck(configureHealthCheckInput).
+		Return(&elb.ConfigureHealthCheckOutput{}, nil)
+
+	target := provider.NewLoadBalancerProvider(mockAWS.Client(), tagStore, mockConfig)
+	result, err := target.Create(req)
+
```
zpatrick: remove whitespace

api/provider/aws/aws_test/load_balancer_create_test.go
```diff
@@ -0,0 +1,335 @@
...
+	mockConfig.EXPECT().Region().Return("region").AnyTimes()
+	mockConfig.EXPECT().AccountID().Return("123456789012")
+	mockConfig.EXPECT().PrivateSubnets().Return([]string{"priv1", "priv2"}).AnyTimes()
+
+	defer provider.SetEntityIDGenerator("lb_id")()
+
+	req := models.CreateLoadBalancerRequest{
+		LoadBalancerName: "lb_name",
+		EnvironmentID:    "env_id",
+		IsPublic:         false,
```
zpatrick: 0-values should be considered "defaults", so this true/false should swap w/ the other test

api/provider/aws/aws_test/load_balancer_create_test.go
```diff
@@ -0,0 +1,335 @@
...
+	mockConfig.EXPECT().PrivateSubnets().Return([]string{"priv1", "priv2"}).AnyTimes()
+	mockConfig.EXPECT().PublicSubnets().Return([]string{"pub1", "pub2"}).AnyTimes()
+
+	defer provider.SetEntityIDGenerator("lb_id")()
+
+	req := models.CreateLoadBalancerRequest{
+		LoadBalancerName: "lb_name",
+		EnvironmentID:    "env_id",
+		IsPublic:         true,
+		Ports: []models.Port{
```
zpatrick: Shouldn't  `[]models.Port{}` should be default?

api/provider/aws/aws_test/load_balancer_create_test.go
```diff
@@ -0,0 +1,335 @@
...
+			{
+				ContainerPort: 80,
+				HostPort:      80,
+				Protocol:      "tcp",
+			},
+		},
+		HealthCheck: models.HealthCheck{
```
zpatrick: shouldn't `[]models.HealthCheck{}` should be default?

api/provider/aws/aws_test/load_balancer_create_test.go
```diff
@@ -0,0 +1,335 @@
...
+			Target:             "TCP:80",
+			Interval:           30,
+			Timeout:            5,
+			HealthyThreshold:   2,
+			UnhealthyThreshold: 2,
+		},
+	}
+
+	readSGHelper(mockAWS, "l0-test-env_id-env", "env_sg")
+
```
zpatrick: [nit] whitespace

api/provider/aws/aws_test/load_balancer_create_test.go
```diff
@@ -0,0 +1,335 @@
...
+	mockAWS.EC2.EXPECT().
+		AuthorizeSecurityGroupIngress(ingressInput).
+		Return(&ec2.AuthorizeSecurityGroupIngressOutput{}, nil)
+
+	iamRoleInput := &iam.CreateRoleInput{}
+	iamRoleInput.SetRoleName("l0-test-lb_id-lb")
+	iamRoleInput.SetAssumeRolePolicyDocument(provider.DEFAULT_ASSUME_ROLE_POLICY)
+
+	mockAWS.IAM.EXPECT().
```
zpatrick: this should be using `gomock.Any()` ...we only care about things that should change from the "master" test above: ports, healthcheck

api/provider/aws/aws_test/load_balancer_create_test.go
```diff
@@ -0,0 +1,335 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	putIAMPolicyInput := &iam.PutRolePolicyInput{}
+	putIAMPolicyInput.SetPolicyName("l0-test-lb_id-lb")
+	putIAMPolicyInput.SetRoleName("l0-test-lb_id-lb")
+	putIAMPolicyInput.SetPolicyDocument(renderedLBRolePolicy)
+
+	mockAWS.IAM.EXPECT().
```
zpatrick: `gomock.Any()`

api/provider/aws/aws_test/load_balancer_create_test.go
```diff
@@ -0,0 +1,335 @@
...
+
+	createLoadBalancerInput := &elb.CreateLoadBalancerInput{}
+	createLoadBalancerInput.SetLoadBalancerName("l0-test-lb_id")
+	createLoadBalancerInput.SetScheme("internet-facing")
+	createLoadBalancerInput.SetSecurityGroups([]*string{aws.String("env_sg"), aws.String("lb_sg")})
+	createLoadBalancerInput.SetSubnets([]*string{aws.String("pub1"), aws.String("pub2")})
+	createLoadBalancerInput.SetListeners(listeners)
+
+	mockAWS.ELB.EXPECT().
+		CreateLoadBalancer(createLoadBalancerInput).
```
zpatrick: I would use the `mock.EXPECT().Do(validate)` pattern to run assertions on only the listeners here.

api/provider/aws/aws_test/load_balancer_create_test.go
```diff
@@ -0,0 +1,335 @@
...
+	target := provider.NewLoadBalancerProvider(mockAWS.Client(), tagStore, mockConfig)
+	result, err := target.Create(req)
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	assert.Equal(t, "lb_id", result)
+
+	expectedTags := models.Tags{
```
zpatrick: can remove - handled in other unit test

api/provider/aws/aws_test/load_balancer_list_test.go
```diff
@@ -0,0 +1,128 @@
...
+			Key:        "environment_id",
+			Value:      "env_id1",
+		},
+		{
+			EntityID:   "env_id1",
+			EntityType: "environment",
+			Key:        "name",
+			Value:      "env_name1",
+		},
+		{
```
zpatrick: I assume this was copy+pasted from my test - for the sake of brevity, we can remove non-critical tags like `os`

api/provider/aws/aws_test/load_balancer_read_test.go
```diff
@@ -0,0 +1,96 @@
...
+		{
+			EntityID:   "env_id",
+			EntityType: "environment",
+			Key:        "name",
+			Value:      "env_name",
+		},
+		{
+			EntityID:   "env_id",
+			EntityType: "environment",
+			Key:        "os",
```
zpatrick: ditto

api/provider/aws/aws_test/load_balancer_read_test.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	describeLoadBalancersInput := &elb.DescribeLoadBalancersInput{}
+	describeLoadBalancersInput.SetLoadBalancerNames([]*string{aws.String("l0-test-lb_id")})
+	describeLoadBalancersInput.SetPageSize(1)
+
+	healthCheck := &elb.HealthCheck{}
+	listener := &elb.Listener{}
+	listenerDescription := &elb.ListenerDescription{}
+	listenerDescription.SetListener(listener)
```
zpatrick: listeners and healthcheck needs to be set to real values here - we should verify they get parsed correctly in the returned model

api/provider/aws/aws_test/load_balancer_read_test.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	expected := &models.LoadBalancer{
+		LoadBalancerID:   "lb_id",
+		LoadBalancerName: "lb_name",
+		EnvironmentID:    "env_id",
+		EnvironmentName:  "env_name",
+		Ports:            []models.Port{models.Port{}},
```
zpatrick: from above: need real ports (preferably > 1) and a populated healthcheck

api/provider/aws/aws_test/load_balancer_create_test.go
```diff
@@ -0,0 +1,335 @@
...
+
+	defer provider.SetEntityIDGenerator("lb_id")()
+
+	req := models.CreateLoadBalancerRequest{
+		LoadBalancerName: "lb_name",
+		EnvironmentID:    "env_id",
+		IsPublic:         true,
+		Ports: []models.Port{
+			{
+				ContainerPort: 80,
```
zpatrick: we should have at least 2 ports, and we shouldn't test with `ContainerPort` and `HostPort` with the same value. They should be different so we verify they map correctly to loadbalancer/instance ports

api/provider/aws/aws_test/load_balancer_update_test.go
```diff
@@ -0,0 +1,201 @@
...
+	mockAWS := awsc.NewMockClient(ctrl)
+	tagStore := tag.NewMemoryStore()
+	mockConfig := mock_config.NewMockAPIConfig(ctrl)
+
+	mockConfig.EXPECT().Instance().Return("test").AnyTimes()
+
+	requestPorts := &[]models.Port{
+		models.Port{
+			CertificateName: "cert",
+			ContainerPort:   8080,
```
zpatrick: ditto: `ContainerPort` and `HostPort` should not be the same

common/aws/mock_aws/mock_cloudtrail.go
```diff
@@ -0,0 +1,681 @@
+// Code generated by MockGen. DO NOT EDIT.
```
zpatrick: remove this file

api/provider/aws/aws_test/load_balancer_create_test.go
jparsons04: I initially took "Defaults" to mean "If you don't provide any optional arguments on the command line", which is to say that the default load balancer you create is going to be public. With this in mind, are you saying that you want me to interpret defaults as any 0-value within a given entity action's request struct?

api/provider/aws/aws_test/load_balancer_create_test.go
jparsons04: The purpose of this is to construct `listeners`, but in the process of constructing the array, `ListServerCertificates` is expected to be called if a certificate name is specified. `listeners` is used on [line 146](https://github.com/quintilesims/layer0/pull/410/files/a256730c903ef3983edfa569c55403e9284e50a1#diff-267fb5b55f4e244c24468db72e65e259R146) in the `SetListeners` method.

It looks strange, but it does follow that pattern, just in a delayed fashion due to `ListServerCertificates`.

api/provider/aws/aws_test/load_balancer_create_test.go
jparsons04: So referring to your previous comment, I'm defining what a Create load balancer request would look like if no arguments are specified on the command line.

In addition to this, the actual entity action method assumes that at least one port is specified [because here, req.Port is iterated over without a previous empty check](https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/load_balancer_create.go#L61), so I have to define at least one `models.Port`

api/provider/aws/aws_test/load_balancer_create_test.go
jparsons04: This is the `TestLoadBalancerCreateDefaults` test we're looking at here. Should I be defining 2 ports in a default checking unit test such as this?

api/provider/aws/aws_test/load_balancer_create_test.go
jparsons04: For the same reason I have to define at least one `models.Port`, I have to define a `models.HealthCheck` [because the elb.HealthCheck is constructed and needs to access the request's HealthCheck fields](https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/load_balancer_create.go#L103)

api/provider/aws/aws_test/load_balancer_create_test.go
```diff
@@ -0,0 +1,335 @@
...
+			Target:             "HTTPS:443/path/to/site",
+			Interval:           60,
+			Timeout:            60,
+			HealthyThreshold:   3,
+			UnhealthyThreshold: 3,
+		},
+	}
+
+	readSGHelper(mockAWS, "l0-test-env_id-env", "env_sg")
+
```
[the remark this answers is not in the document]
jparsons04: :white_check_mark:

api/provider/aws/aws_test/load_balancer_create_test.go
```diff
@@ -0,0 +1,335 @@
...
+		t.Fatal(err)
+	}
+
+	putIAMPolicyInput := &iam.PutRolePolicyInput{}
+	putIAMPolicyInput.SetPolicyName("l0-test-lb_id-lb")
+	putIAMPolicyInput.SetRoleName("l0-test-lb_id-lb")
+	putIAMPolicyInput.SetPolicyDocument(renderedLBRolePolicy)
+
+	mockAWS.IAM.EXPECT().
+		PutRolePolicy(putIAMPolicyInput)
```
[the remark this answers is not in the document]
jparsons04: :white_check_mark:

api/provider/aws/aws_test/load_balancer_create_test.go
jparsons04: :white_check_mark:

api/provider/aws/aws_test/load_balancer_create_test.go
jparsons04: :white_check_mark:

api/provider/aws/aws_test/load_balancer_create_test.go
jparsons04: :white_check_mark: Makes sense.

api/provider/aws/aws_test/load_balancer_create_test.go
jparsons04: :white_check_mark:

api/provider/aws/aws_test/load_balancer_create_test.go
jparsons04: :white_check_mark:

api/provider/aws/aws_test/load_balancer_list_test.go
jparsons04: :white_check_mark:

common/aws/mock_aws/mock_cloudtrail.go
jparsons04: :white_check_mark:

api/provider/aws/aws_test/load_balancer_create_test.go
jparsons04: To your other point, something I'm also immediately concerned about is the implementation of the entity method action `Create()`. The reason why I use `port.HostPort` [for both the From and To ports](https://github.com/quintilesims/layer0/blob/a256730c903ef3983edfa569c55403e9284e50a1/api/provider/aws/aws_test/load_balancer_create_test.go#L73) is because of the helper function [authorizeSGIngressFromPort](https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/load_balancer_create.go#L155) which takes in the load balancer security group ID and a single `int64` port.

When [this helper function is called](https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/load_balancer_create.go#L62), we are passing in the iterated on `req.Ports.HostPort`. 

The thing that would case this to break would be `input.SetToPort` in the helper function `authorizeSGIngressFromPort`, because this clearly would need to be the `req.Ports.ContainerPort`.

Can I conclude that this is a bug in the helper function then?

jparsons04: Actually, disregard my last comment. I'm conflating the port mappings and the ingress security group. The port mappings should be different, but the security group ingress authorization for the load balancer will always use the host port for To and From because traffic from the public will go through the load balancer and to the host port. The container will forward traffic from the host to the listening container port.

So, this is by design.

Still, given that we are looking at the Defaults unit test, I'm curious if you think I still need to define 2 ports for that test.

api/provider/aws/aws_test/load_balancer_create_test.go
jparsons04: :white_check_mark:

api/provider/aws/aws_test/load_balancer_update_test.go
jparsons04: :white_check_mark:

api/provider/aws/aws_test/load_balancer_create_test.go
jparsons04: Should I remove it from here, or remove it from the Defaults test? For now, I've opted to remove it from the Defaults one.

api/provider/aws/aws_test/load_balancer_create_test.go
zpatrick: This package should have no knowledge of the cli. In that regard, yes the latter option is better. Beyond that though, the default environment os is "linux", not an empty string - so it's not strictly 0-values

api/provider/aws/aws_test/load_balancer_create_test.go
zpatrick: Remove it from defaults - the scope of that test does not include tags

api/provider/aws/aws_test/load_balancer_create_test.go
zpatrick: [addressed first part in other comment]

You can iterate over empty or `nil` slices in a `for := range` loop w/o checking if it is empty

api/provider/aws/aws_test/load_balancer_create_test.go
zpatrick: Sorry, meant for the 2 ports to be in the 'master' test. I assume a default of 0 ports is allowed?

api/provider/aws/aws_test/load_balancer_create_test.go
zpatrick: I think we need to handle a default healthcheck if one isn't specified then. `defaults.go` is a good place for this.

api/provider/aws/aws_test/load_balancer_create_test.go
jparsons04: :white_check_mark: So I made this empty, implemented a default in `defaults.go` and made sure to use `gomock.Any()` in the expected calls in the defaults test given they were already tested in the master test.

api/provider/aws/aws_test/load_balancer_create_test.go
jparsons04: :white_check_mark:

api/provider/aws/aws_test/load_balancer_create_test.go
zpatrick: i mean it would be something like:
```
serverMeta := &iam.ServerCertMeta{}
serverMeta.SetArn(...)
serverMeta.SetServerCertificatName(...)

serverMetaList := []*iam.ServerMeta{serverMeta}
...
```

api/provider/aws/aws_test/load_balancer_create_test.go
```diff
@@ -0,0 +1,257 @@
...
+	putIAMPolicyInput.SetPolicyName("l0-test-lb_id-lb")
+	putIAMPolicyInput.SetRoleName("l0-test-lb_id-lb")
+	putIAMPolicyInput.SetPolicyDocument(renderedLBRolePolicy)
+
+	mockAWS.IAM.EXPECT().
+		PutRolePolicy(putIAMPolicyInput).
+		Return(&iam.PutRolePolicyOutput{}, nil)
+
+	listeners := make([]*elb.Listener, len(req.Ports))
+	for i, port := range req.Ports {
```
zpatrick: I'm a little torn on this because it will make an already huge unit test even larger - but there's a lot of value in  avoiding logic like this to create our expected objects and instead build them by hand. For example, if someone made a change/mistake when setting up the `req` object above (e.g. remove `req.Ports` entirely), it could cause false positives.  

see bivers and my convo at https://github.com/quintilesims/layer0/pull/409/files/8c08f98d76bb3564ab00127c2b15db30bc0d67b0#diff-92d2c246bc12ec7b170dd0e92ef06122R92

api/provider/aws/aws_test/load_balancer_create_test.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	createLoadBalancerInput := &elb.CreateLoadBalancerInput{}
+	createLoadBalancerInput.SetLoadBalancerName("l0-test-lb_id")
+	createLoadBalancerInput.SetScheme("internet-facing")
+	createLoadBalancerInput.SetSecurityGroups([]*string{aws.String("env_sg"), aws.String("lb_sg")})
+	createLoadBalancerInput.SetSubnets([]*string{aws.String("pub1"), aws.String("pub2")})
+	createLoadBalancerInput.SetListeners(listeners)
+
+	validateFN := func(input *elb.CreateLoadBalancerInput) {
```
zpatrick: this is redundant - the `Listeners` field is already checked for equality below at `CreateLoadBalancer(createLoadBalancerInput).`

api/provider/aws/aws_test/load_balancer_create_test.go
```diff
@@ -0,0 +1,257 @@
...
+	mockAWS.IAM.EXPECT().
+		CreateRole(gomock.Any()).
+		Return(&iam.CreateRoleOutput{}, nil)
+
+	mockAWS.IAM.EXPECT().
+		PutRolePolicy(gomock.Any()).
+		Return(&iam.PutRolePolicyOutput{}, nil)
+
+	mockAWS.ELB.EXPECT().
+		CreateLoadBalancer(gomock.Any()).
```
zpatrick: Don't we need to check for default ports (if that exists?) and the default healthcheck?

api/provider/aws/aws_test/load_balancer_list_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+		output := &elb.DescribeLoadBalancersOutput{}
+		output.SetLoadBalancerDescriptions(loadBalancerDescriptions)
+		fn(output, true)
+
+		return nil
+	}
+
+	mockAWS.ELB.EXPECT().
+		DescribeLoadBalancersPages(gomock.Any(), gomock.Any()).
```
zpatrick: we should be using an `*elb.DescribeLoadBalancersInput` object here, not `gomock.Any()`. If that input object has fields set, you'll need to do that as well

api/provider/aws/aws_test/load_balancer_list_test.go
```diff
@@ -0,0 +1,116 @@
...
+		},
+	}
+
+	for _, tag := range tags {
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	describeLoadBalancerPagesFN := func(input *elb.DescribeLoadBalancersInput, fn func(output *elb.DescribeLoadBalancersOutput, lastPage bool) bool) error {
```
zpatrick: Do we expect only loadbalancers for a particular instance get returned from the api call? If not, we should throw in loadbalancers that don't belong to the layer0 instance to ensure they get filtered out.

zpatrick: edit: just checked, the api call can return anything, so you should definitely add those extras

api/provider/aws/aws_test/load_balancer_read_test.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+	}
+
+	describeLoadBalancersInput := &elb.DescribeLoadBalancersInput{}
+	describeLoadBalancersInput.SetLoadBalancerNames([]*string{aws.String("l0-test-lb_id")})
+	describeLoadBalancersInput.SetPageSize(1)
+
+	healthCheck := models.HealthCheck{
+		Target:             "HTTPS:443/path/to/site",
+		Interval:           5,
```
zpatrick: ditto: different ints

api/provider/aws/aws_test/load_balancer_read_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+	elbHealthCheck := &elb.HealthCheck{}
+	elbHealthCheck.SetTarget(healthCheck.Target)
+	elbHealthCheck.SetInterval(int64(healthCheck.Interval))
+	elbHealthCheck.SetTimeout(int64(healthCheck.Timeout))
+	elbHealthCheck.SetHealthyThreshold(int64(healthCheck.HealthyThreshold))
+	elbHealthCheck.SetUnhealthyThreshold(int64(healthCheck.UnhealthyThreshold))
+
+	listenerDescriptions := make([]*elb.ListenerDescription, len(ports))
+	for i, port := range ports {
```
zpatrick: ditto on my comment above about using too much logic

api/provider/aws/aws_test/load_balancer_update_test.go
```diff
@@ -0,0 +1,201 @@
...
+			ContainerPort:   4444,
+			HostPort:        444,
+			Protocol:        "https",
+		},
+	}
+
+	req := models.UpdateLoadBalancerRequest{
+		LoadBalancerID: "lb_name",
+		Ports:          requestPorts,
+		HealthCheck: &models.HealthCheck{
```
zpatrick: you need to format your code

api/provider/aws/aws_test/load_balancer_update_test.go
```diff
@@ -0,0 +1,201 @@
...
+			Target:             "HTTPS:444/path/to/site",
+			Interval:           75,
+			Timeout:            75,
+			HealthyThreshold:   4,
```
zpatrick: ditto

api/provider/aws/aws_test/load_balancer_update_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+	configureHealthCheckOutput := &elb.ConfigureHealthCheckOutput{}
+	configureHealthCheckOutput.SetHealthCheck(healthCheck)
+
+	mockAWS.ELB.EXPECT().
+		ConfigureHealthCheck(configureHealthCheckInput).
+		Return(configureHealthCheckOutput, nil)
+
+	listeners := make([]*elb.Listener, len(*requestPorts))
+	for i, port := range *requestPorts {
```
zpatrick: the more i read these the more i think hardcoding would be the better approach

api/provider/aws/aws_test/load_balancer_update_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+	createLoadBalancerListenersInput := &elb.CreateLoadBalancerListenersInput{}
+	createLoadBalancerListenersInput.SetLoadBalancerName("l0-test-lb_name")
+	createLoadBalancerListenersInput.SetListeners(listeners)
+
+	mockAWS.ELB.EXPECT().
+		CreateLoadBalancerListeners(createLoadBalancerListenersInput).
+		Return(&elb.CreateLoadBalancerListenersOutput{}, nil)
+
+	for _, listener := range listeners {
```
zpatrick: this test is massive - maybe we can break some of it out into helpers?

api/provider/aws/defaults.go
```diff
@@ -121,3 +123,19 @@ const DEFAULT_LB_ROLE_POLICY_TEMPLATE = `{
         }
     ]
 }`
+
+var DEFAULT_CREATE_LB_PORT_MODEL = []models.Port{
```
zpatrick: I was thinking about this today - why do we have a default port? It doesn't make much sense to me. It's pretty rare that you actually want a `80:80/tcp` port... seems better to just not have a default. It also make a ton of sense when you think of it from a terraform configuration perspective.

api/provider/aws/defaults.go
```diff
@@ -121,3 +123,19 @@ const DEFAULT_LB_ROLE_POLICY_TEMPLATE = `{
...
+	models.Port{
+		ContainerPort: 80,
+		HostPort:      80,
+		Protocol:      "tcp",
+	},
+}
+
+var DEFAULT_CREATE_LB_HEALTHCHECK_MODEL = models.HealthCheck{
```
zpatrick: You didn't update the `cli/command` package to reference these values. 
While you're doing that, it would be great if you could knock out all default values. 

* Environment OS == "linux"
* Environment Size = "m3.medium"
* Service Scale = 1 (this is new), see my `terraform-update` branch

api/provider/aws/defaults.go
```diff
@@ -1,5 +1,7 @@
 package aws
 
+import "github.com/quintilesims/layer0/common/models"
+
 const (
 	DEFAULT_INSTANCE_SIZE  = "m3.medium"
```
zpatrick: This issue was originally brought on by me - I don't think Go uses all caps for constants. I've just seen upper-camel case

api/provider/aws/defaults.go
jparsons04: That makes sense, but should I encapsulate that in a different PR/open a different issue for fixing that specific problem?

api/provider/aws/defaults.go
jparsons04: As with the previous comment, I wonder if it would be worth it to open a new issue and open a new PR to address the cli/command that would be required here.

zpatrick: The cli portion sure, but you'll need those to exist to finish these unit tests

jparsons04:
So I've tried to implement quite a few things given the last PR comments:

1) A couple of helper functions in `helpers.go` to try to clean up some of the size of the unit tests
2) I've tried to favor hardcoding expected objects where possible
3) I've changed the casing within the `defaults.go` file

Unit tests still look good to me, but it could use another styling/formatting review pass @zpatrick @diemonster

api/provider/aws/defaults.go
```diff
@@ -121,3 +124,19 @@ const DEFAULT_LB_ROLE_POLICY_TEMPLATE = `{
         }
     ]
 }`
+
+var DefaultCreateLBPortModel = []models.Port{
```
zpatrick: I'm still up in the air about having a default port - it just seems 'lucky' that someone would need `80:80/tcp` for their service. If we do keep it, it should probably be just `var DefaultLBPort = models.Port`, not a slice (for easy conversion in regards to string in the cli and elb listeners objects in unit tests)

api/provider/aws/test_aws/helpers.go
```diff
@@ -52,3 +55,55 @@ func deleteSGHelper(mockAWS *awsc.MockClient, securityGroupID string) {
...
+	elbHealthCheck.SetInterval(int64(healthCheck.Interval))
+	elbHealthCheck.SetTimeout(int64(healthCheck.Timeout))
+	elbHealthCheck.SetHealthyThreshold(int64(healthCheck.HealthyThreshold))
+	elbHealthCheck.SetUnhealthyThreshold(int64(healthCheck.UnhealthyThreshold))
+
+	return elbHealthCheck
+}
+
+func listenerHelper(port *models.Port) *elb.Listener {
+	if port == nil {
```
zpatrick: this doesn't seem intuitive/too 'magical' to me; converting nil to the default seems like something the caller should handle

api/provider/aws/test_aws/helpers.go
```diff
@@ -52,3 +55,55 @@ func deleteSGHelper(mockAWS *awsc.MockClient, securityGroupID string) {
...
+	listener.SetInstancePort(port.ContainerPort)
+
+	switch strings.ToLower(port.Protocol) {
+	case "http", "https":
+		listener.SetInstanceProtocol("http")
+	case "tcp", "ssl":
+		listener.SetInstanceProtocol("tcp")
+	}
+
+	if port.CertificateName != "" {
```
zpatrick: Doesn't the certificate name need to be resolved first? I'm guessing this is passing only because you specify the certificate name == arn below. That's too much magic, we need to be more explicit about that mapping and ensure the caller uses the arn/name in the correct places.

api/provider/aws/test_aws/load_balancer_create_test.go
```diff
@@ -0,0 +1,275 @@
...
+	putIAMPolicyInput.SetPolicyName("l0-test-lb_id-lb")
+	putIAMPolicyInput.SetRoleName("l0-test-lb_id-lb")
+	putIAMPolicyInput.SetPolicyDocument(renderedLBRolePolicy)
+
+	mockAWS.IAM.EXPECT().
+		PutRolePolicy(putIAMPolicyInput).
+		Return(&iam.PutRolePolicyOutput{}, nil)
+
+	serverCertificateMetadata := &iam.ServerCertificateMetadata{}
+	serverCertificateMetadata.SetArn("cert")
```
zpatrick: the cert arn and name need to be different values; we need to ensure proper use of name vs. arn elsewhere

api/provider/aws/test_aws/load_balancer_create_test.go
```diff
@@ -0,0 +1,275 @@
...
+
+	listServerCertificatesOutput := &iam.ListServerCertificatesOutput{}
+	listServerCertificatesOutput.SetServerCertificateMetadataList(serverCertificateMetadataList)
+
+	mockAWS.IAM.EXPECT().
+		ListServerCertificates(&iam.ListServerCertificatesInput{}).
+		Return(listServerCertificatesOutput, nil).
+		AnyTimes()
+
+	listener1 := &elb.Listener{}
```
zpatrick: shouldn't you use your helper here?

api/provider/aws/test_aws/load_balancer_create_test.go
```diff
@@ -0,0 +1,275 @@
...
+
+	listeners := []*elb.Listener{listener1, listener2}
+	createLoadBalancerInput := &elb.CreateLoadBalancerInput{}
+	createLoadBalancerInput.SetLoadBalancerName("l0-test-lb_id")
+	createLoadBalancerInput.SetScheme("internet-facing")
+	createLoadBalancerInput.SetSecurityGroups([]*string{aws.String("env_sg"), aws.String("lb_sg")})
+	createLoadBalancerInput.SetSubnets([]*string{aws.String("pub1"), aws.String("pub2")})
+	createLoadBalancerInput.SetListeners(listeners)
+
+	validateFN := func(input *elb.CreateLoadBalancerInput) {
```
zpatrick: you can remove this - it's redundant validation with `CreateLoadBalancer(createLoadBalancerInput).` below

api/provider/aws/test_aws/load_balancer_create_test.go
```diff
@@ -0,0 +1,275 @@
...
+			assert.NoError(t, err)
+		}
+	}
+
+	mockAWS.ELB.EXPECT().
+		CreateLoadBalancer(createLoadBalancerInput).
+		Do(validateFN).
+		Return(&elb.CreateLoadBalancerOutput{}, nil)
+
+	healthCheck := &elb.HealthCheck{}
```
zpatrick: helper?

api/provider/aws/test_aws/load_balancer_create_test.go
```diff
@@ -0,0 +1,275 @@
...
+
+	mockAWS.IAM.EXPECT().
+		CreateRole(gomock.Any()).
+		Return(&iam.CreateRoleOutput{}, nil)
+
+	mockAWS.IAM.EXPECT().
+		PutRolePolicy(gomock.Any()).
+		Return(&iam.PutRolePolicyOutput{}, nil)
+
+	listeners := []*elb.Listener{listenerHelper(nil)}
```
zpatrick: Going off of my other comment - we should definitely call something like `listenerHelper(defaults.DefaultLBPort`) instead of using `nil`

api/provider/aws/test_aws/load_balancer_read_test.go
```diff
@@ -0,0 +1,125 @@
...
+		DescribeLoadBalancers(describeLoadBalancersInput).
+		Return(describeLoadBalancersOutput, nil)
+
+	target := provider.NewLoadBalancerProvider(mockAWS.Client(), tagStore, mockConfig)
+	result, err := target.Read("lb_id")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	expected := &models.LoadBalancer{
```
zpatrick: This needs the `ServiceID` and `ServiceName` fields tested (I believe you only need to populate some tags)

api/provider/aws/test_aws/helpers.go
```diff
@@ -52,3 +55,55 @@ func deleteSGHelper(mockAWS *awsc.MockClient, securityGroupID string) {
...
+	elbHealthCheck.SetTarget(healthCheck.Target)
+	elbHealthCheck.SetInterval(int64(healthCheck.Interval))
+	elbHealthCheck.SetTimeout(int64(healthCheck.Timeout))
+	elbHealthCheck.SetHealthyThreshold(int64(healthCheck.HealthyThreshold))
+	elbHealthCheck.SetUnhealthyThreshold(int64(healthCheck.UnhealthyThreshold))
+
+	return elbHealthCheck
+}
+
+func listenerHelper(port *models.Port) *elb.Listener {
```
zpatrick: [nit] it's a little odd to have `*models.Port` as the param type when we typically work with `models.Port` objects

api/provider/aws/test_aws/load_balancer_read_test.go
```diff
@@ -0,0 +1,125 @@
...
+			Protocol:        "https",
+		},
+		models.Port{
+			ContainerPort: 88,
+			HostPort:      80,
+			Protocol:      "tcp",
+		},
+	}
+
+	elbHealthCheck := &elb.HealthCheck{}
```
zpatrick: helper?

api/provider/aws/test_aws/load_balancer_read_test.go
```diff
@@ -0,0 +1,125 @@
...
+	elbHealthCheck.SetTarget(healthCheck.Target)
+	elbHealthCheck.SetInterval(int64(healthCheck.Interval))
+	elbHealthCheck.SetTimeout(int64(healthCheck.Timeout))
+	elbHealthCheck.SetHealthyThreshold(int64(healthCheck.HealthyThreshold))
+	elbHealthCheck.SetUnhealthyThreshold(int64(healthCheck.UnhealthyThreshold))
+
+	listener1 := listenerHelper(&ports[0])
+	listenerDescription1 := &elb.ListenerDescription{}
+	listenerDescription1.SetListener(listener1)
```
zpatrick: can probably remove the first line in this:
```
listenerDescription1 := &elb.ListenerDescription{}
listenerDescription1.SetListener(listenerHelper(ports[0]))
```

api/provider/aws/test_aws/load_balancer_update_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+	configureHealthCheckOutput := &elb.ConfigureHealthCheckOutput{}
+	configureHealthCheckOutput.SetHealthCheck(healthCheck)
+
+	mockAWS.ELB.EXPECT().
+		ConfigureHealthCheck(configureHealthCheckInput).
+		Return(configureHealthCheckOutput, nil)
+
+	serverCertificateMetadata := &iam.ServerCertificateMetadata{}
+	serverCertificateMetadata.SetArn("cert")
```
zpatrick: ditto - need different arns/names

api/provider/aws/test_aws/load_balancer_update_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+	mockAWS.IAM.EXPECT().
+		ListServerCertificates(&iam.ListServerCertificatesInput{}).
+		Return(listServerCertificatesOutput1, nil)
+
+	serverCertificateMetadataList2 := []*iam.ServerCertificateMetadata{serverCertificateMetadata}
+	listServerCertificatesOutput2 := &iam.ListServerCertificatesOutput{}
+	listServerCertificatesOutput2.SetServerCertificateMetadataList(serverCertificateMetadataList2)
+
+	mockAWS.IAM.EXPECT().
```
zpatrick: this seems redundant - you can just use the `Times(n)` or `.AnyTimes()`  option to the mock, e.g.:
```
mockAWS.IAM.EXPECT().
    ListServerCertificates(&iam.ListServerCertificatesInput{}).
    Return(listServerCertificatesOutput, nil).
    AnyTimes()
```

api/provider/aws/test_aws/load_balancer_update_test.go
```diff
@@ -0,0 +1,172 @@
...
+		ListServerCertificates(&iam.ListServerCertificatesInput{}).
+		Return(listServerCertificatesOutput2, nil)
+
+	readSGHelper(mockAWS, "l0-test-lb_name-lb", "lb_sg")
+
+	describeLoadBalancersInput := &elb.DescribeLoadBalancersInput{}
+	describeLoadBalancersInput.SetLoadBalancerNames([]*string{aws.String("l0-test-lb_name")})
+	describeLoadBalancersInput.SetPageSize(1)
+
+	healthCheck = healthCheckHelper(nil)
```
zpatrick: ditto - the `nil` values should be replaced with the default variables

api/provider/aws/test_aws/load_balancer_update_test.go
```diff
@@ -0,0 +1,172 @@
...
+	listener := listenerHelper(nil)
+	listenerDescription := &elb.ListenerDescription{}
+	listenerDescription.SetListener(listener)
```
zpatrick: ditto - can probably save a line here

api/provider/aws/test_aws/load_balancer_update_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+	createLoadBalancerListenersInput := &elb.CreateLoadBalancerListenersInput{}
+	createLoadBalancerListenersInput.SetLoadBalancerName("l0-test-lb_name")
+	createLoadBalancerListenersInput.SetListeners(listeners)
+
+	mockAWS.ELB.EXPECT().
+		CreateLoadBalancerListeners(createLoadBalancerListenersInput).
+		Return(&elb.CreateLoadBalancerListenersOutput{}, nil)
+
+	ingressInput1 := &ec2.AuthorizeSecurityGroupIngressInput{}
```
zpatrick: helpers for ingress?

api/provider/aws/test_aws/load_balancer_create_test.go
jparsons04: So, I could use the `req.Ports` object to send to the helper or hardcode the expected objects as we discussed in previous comments.

api/provider/aws/defaults.go
zpatrick: after discussion, I think we should keep a default port due to the security group requirements. However, I still think this should be a single object, not a slice. Also, variable name is over-specific - 
```
var DefaultLoadBalancerPort = models.Port{ ... }
```

api/provider/aws/defaults.go
```diff
@@ -121,3 +124,19 @@ const DEFAULT_LB_ROLE_POLICY_TEMPLATE = `{
...
+	models.Port{
+		ContainerPort: 80,
+		HostPort:      80,
+		Protocol:      "tcp",
+	},
+}
+
+var DefaultCreateLBHealthCheckModel = models.HealthCheck{
```
zpatrick: `var DefaultHealthCheck = models.HealthCheck{...}`

api/provider/aws/test_aws/helpers.go
```diff
@@ -52,3 +55,55 @@ func deleteSGHelper(mockAWS *awsc.MockClient, securityGroupID string) {
 		DeleteSecurityGroup(input).
 		Return(&ec2.DeleteSecurityGroupOutput{}, nil)
 }
+
+func healthCheckHelper(healthCheck *models.HealthCheck) *elb.HealthCheck {
+	if healthCheck == nil {
```
zpatrick: this block needs to be removed - too much 'magic'. The caller should be responsible for passing in the default

api/provider/aws/defaults.go
jparsons04: Based on our previous conversation, because `CreateLoadBalancerInput` requires that you provide at least one `Listener` in a slice, we're opting to keep the notion of a default port unless we figure out a better UX pattern when a user wants to create a load balancer without providing a port.

api/provider/aws/test_aws/load_balancer_read_test.go
jparsons04: :white_check_mark:

api/provider/aws/defaults.go
jparsons04: :white_check_mark: Both fixed.

api/provider/aws/test_aws/load_balancer_update_test.go
```diff
@@ -0,0 +1,149 @@
...
+		Return(describeLoadBalancersOutput, nil)
+
+	revokeIngressInput := revokeSGIngressHelper(provider.DefaultLoadBalancerPort)
+	revokeIngressInput.SetGroupId("lb_sg")
+
+	mockAWS.EC2.EXPECT().
+		RevokeSecurityGroupIngress(revokeIngressInput).
+		Return(&ec2.RevokeSecurityGroupIngressOutput{}, nil)
+
+	port := int64(80)
```
zpatrick: this should probably reference the default port for clarity, e.g. `DefaultLBPort.HostPort`

jparsons04: Ah, good point. :white_check_mark:

PR #411, opened 2017-11-01
Environment links changes

sesh-kebab:

Summary:
-  update endpoints from PUT to PATCH requests
-  update environment links to be uni-directional
-  update entity update request models to not include the entity id (as this is already included in the path e.g., PATCH /environment/:id { minCount: 0, links: [] }
-  update swagger to match the change to above
-  client update signatures have been updated to require entityID as well the request object

Detailed:

Updates L0 API endpoints that deal with the creation and deletion of environment links. This will no longer be exposed by a POST and DELETE request to endpoint `/environment/:id/link` but will be now exposed via a common PUT request used for environment updates.

Note that a link is now uni-directional. Backwards compatibility is maintained by making bi-directional flag/attribute true by default.

This also fixes existing behavior where a single environment, linked to more than 1 environment, was not correctly represented in the tag store. Multiple links are now stored as a comma-delimited list of environment ids and are rendered correctly when creating an environment model from the information stored in the tag store.

Also, the changes from PUT to PATCH also additionally changes update request endpoints
```
from:
PUT /environment/
{ id: 1, minCount: 0, links: [] }

to:
PATCH /environment/:id/
{ minCount: 0, links: [] }
```

Note the omission of the id in the body of the patch request. This is to avoid redundancy and with having the same data be specified in two places.

- Run unit tests.
- Run link and unlink commands via the CLI
- Run link and unlink commands via Swagger

**Checklist**
- [x] Unit tests

Todo:
- [ ] Terraform Layer0 Provider changes

closes #401 #416 and the update entity update paths to be restful in the API refactor issue #274


cli/command/environment.go
```diff
@@ -236,18 +252,41 @@ func (e *EnvironmentCommand) link(c *cli.Context) error {
 		return fmt.Errorf("Cannot link an environment to itself")
 	}
 
-	req := models.CreateEnvironmentLinkRequest{
-		SourceEnvironmentID: sourceEnvironmentID,
-		DestEnvironmentID:   destEnvironmentID,
+	updateEnvReq, err := e.generateAddLinkRequest(sourceEnvironmentID, destEnvironmentID)
```
zpatrick: I think we can make this quite a bit more readable and DRY with some helper functions. Maybe something like:
```
func (e *EnvironmentCommand) addLink(c *cli.Context) error {
	// ignore duplicates for bi-directional idempotence
	generateRequest := func(links []string, destEnvironmentID string) models.UpdateEnvironmentRequest {
		links = append(links, destEnvironmentID)

		return models.UpdateEnvironmentRequest{
			Links: &links,
		}
	}

	return e.updateEnvironmentLinksHelper(c, generateRequest)
}

func (e *EnvironmentCommand) removeLink(c *cli.Context) error {
	// ignore missing for bi-directional idempotence
	generateRequest := func(links []string, destEnvironmentID string) models.UpdateEnvironmentRequest {
		for i, link := range links {
			if link == destEnvironmentID {
				links = append(links[:i], links[i+1:]...)
				i--
			}
		}

		return models.UpdateEnvironmentRequest{
			Links: &links,
		}
	}

	return e.updateEnvironmentLinksHelper(c, generateRequest)
}

func (e *EnvironmentCommand) updateEnvironmentLinksHelper(
	c *cli.Context,
	generateRequest func(links []string, destEnvironmentID string) models.UpdateEnvironmentRequest,
) error {
	args, err := parseArgs(c, "SOURCE", "DEST")
	if err != nil {
		return err
	}

	sourceEnvironmentID, err := resolveSingle(args["SOURCE"])
	if err != nil {
		return err
	}

	destEnvironmentID, err := resolveSingle(args["DEST"])
	if err != nil {
		return err
	}

	updateFN := func(sourceEnvironmentID, destEnvironmentID string) error {
		sourceEnvironment, err := c.Client.ReadEnvironment(sourceEnvironmentID)
		if err != nil {
			return err
		}

		req := generateRequest(sourceEnvironment.Links, destEnvironmentID)
		jobID, err := e.Client.UpdateEnvironment(sourceEnvironmentID, req)
		if err != nil {
			return
		}

		return waitOnJobHelper(c, jobID, "updating environment", func(string) error {
			e.printer.Printf("done")
			return nil
		})
	}

	if err := updateFN(sourceEnvironmentID, destEnvironmentID); err != nil {
		return err
	}

	if c.Bool("uni-directional") {
		return nil
	}

	return updateFN(destEnvironmentID, sourceEnvironmentID)
}
```

cli/command/environment_test.go
```diff
@@ -338,19 +298,97 @@ func TestEnvironmentSetMinCount(t *testing.T) {
...
+			Return(&models.Environment{
+				EnvironmentID: "env_id1",
+				Links:         []string{},
+			}, nil)
+
+		base.Client.EXPECT().
+			ReadEnvironment("env_id2").
+			Return(&models.Environment{
+				EnvironmentID: "env_id2",
+				Links:         []string{},
```
zpatrick: Let's make this `[]string{"env_id3"}` so we are testing that we do idempotence correctly

cli/command/environment_test.go
```diff
@@ -338,19 +298,97 @@ func TestEnvironmentSetMinCount(t *testing.T) {
...
+
+		base.Resolver.EXPECT().
+			Resolve("environment", "env_name2").
+			Return([]string{"env_id2"}, nil)
+
+		base.Client.EXPECT().
+			ReadEnvironment("env_id1").
+			Return(&models.Environment{
+				EnvironmentID: "env_id1",
+				Links:         []string{},
```
zpatrick: ditto, `[]string{"env_id3"}`

cli/command/environment_test.go
```diff
@@ -380,19 +432,97 @@ func TestEnvironmentLink(t *testing.T) {
...
+func TestEnvironmentUnlinkBiDirectional(t *testing.T) {
 	testWaitHelper(t, func(t *testing.T, wait bool) {
 		base, ctrl := newTestCommand(t)
 		defer ctrl.Finish()
 
 		command := NewEnvironmentCommand(base.Command())
 
-		job := &models.Job{
-			JobID:  "job_id",
+		links1 := []string{}
```
zpatrick: ditto, `[]string{"env_id3"}`

client/environment.go
```diff
@@ -46,27 +46,8 @@ func (c *APIClient) ReadEnvironment(environmentID string) (*models.Environment,
...
-	}
-
-	return resp.JobID, nil
-}
-
-func (c *APIClient) DeleteLink(req models.DeleteEnvironmentLinkRequest) (string, error) {
-	var resp models.CreateJobResponse
-	path := fmt.Sprintf("/environment/%s/link/%s", req.SourceEnvironmentID, req.DestEnvironmentID)
-	if err := c.client.Delete(path, nil, &resp); err != nil {
+	path := fmt.Sprintf("/environment/%s", req.EnvironmentID)
```
zpatrick: `EnvironmentID` needs to be removed from this object. The `UpdateEnvironment` signature should be: 
```
UpdateEnvironment(environmentID string, req models.UpdateEnvironmentRequest) (string, error)
```

common/models/update_environment_request.go
```diff
@@ -7,8 +7,9 @@ import (
 )
 
 type UpdateEnvironmentRequest struct {
-	EnvironmentID   string `json:"environment_id"`
-	MinClusterCount *int   `json:"min_cluster_count"`
+	EnvironmentID   string    `json:"environment_id"`
```
zpatrick: Need to remove `EnvironmentID` field here. We will need a new object to pass into the job runner.

As a pattern maybe we swap the `-Request` suffix with `-Job` and wrap the original request?

```
type UpdateEnvironmentJob struct {
    EnvironmentID string
    Req           UpdateEnvironmentRequest
}
```

api/provider/aws/common.go
```diff
@@ -243,3 +243,14 @@ func listClusterTaskARNs(ecsapi ecsiface.ECSAPI, clusterName, startedBy, status
 
 	return taskARNs, nil
 }
+
+func removeEmptyEntries(input []string) []string {
```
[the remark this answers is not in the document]
sesh-kebab: I agree there are different ways to do this but I wouldn't say this is overkill. `string.Replace` gets tricky when you have are removing comma prefixes and suffixes.

api/provider/aws/environment_update.go
```diff
@@ -52,3 +139,175 @@ func (e *EnvironmentProvider) updateASGSize(autoScalingGroupName string, minSize
 
 	return nil
 }
+
+func (e *EnvironmentProvider) createIngressInput(sourceGroupID, destGroupID string) error {
+	if sourceGroupID == "" || destGroupID == "" {
```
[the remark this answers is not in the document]
sesh-kebab: ✅ agreed this was intended for testing during development

api/provider/aws/environment_update.go
```diff
@@ -52,3 +139,175 @@ func (e *EnvironmentProvider) updateASGSize(autoScalingGroupName string, minSize
...
+		}
+
+		return err
+	}
+
+	return nil
+}
+
+func (e *EnvironmentProvider) createLinkTags(sourceEnvironmentID, destEnvironmentID string) error {
+	if sourceEnvironmentID == "" || destEnvironmentID == "" {
```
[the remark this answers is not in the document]
sesh-kebab: ✅ agreed this was intended for testing during development

api/provider/aws/environment_update.go
```diff
@@ -52,3 +139,175 @@ func (e *EnvironmentProvider) updateASGSize(autoScalingGroupName string, minSize
...
+		if err := e.TagStore.Insert(tag); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (e *EnvironmentProvider) removeIngressRule(groupID, groupIDToRemove string) error {
+	if groupID == "" || groupIDToRemove == "" {
```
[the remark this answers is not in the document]
sesh-kebab: ✅ agreed this was intended for testing during development

api/provider/aws/environment_update.go
```diff
@@ -52,3 +139,175 @@ func (e *EnvironmentProvider) updateASGSize(autoScalingGroupName string, minSize
...
+	for _, tag := range updatedTags {
+		if err := e.TagStore.Insert(tag); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (e *EnvironmentProvider) describeSG(groupID string) (*ec2.SecurityGroup, error) {
```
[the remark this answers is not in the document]
sesh-kebab: readSG expects a `group name` as an argument. I only have  security group id at this point.

cli/command/environment.go
```diff
@@ -82,12 +86,24 @@ func (e *EnvironmentCommand) Command() cli.Command {
 				Usage:     "links two environments together",
 				Action:    e.link,
 				ArgsUsage: "SOURCE_ENVIRONMENT_NAME DESTINATION_ENVIRONMENT_NAME",
+				Flags: []cli.Flag{
+					cli.BoolTFlag{
+						Name:  FLAG_BI_DIRECTIONAL,
```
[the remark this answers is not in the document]
sesh-kebab: In this case though, I would prefer having to set a flag to false to override the default behavior as we don't actually intend the usage of this flag. Needing to set something to false would require more confidence on the user's part to know what they are doing.

cli/command/environment_test.go
sesh-kebab: ✅  sure.

cli/command/environment_test.go

cli/command/environment_test.go

client/environment.go
sesh-kebab: Not sure what you mean. It doesn't make sense to remove the entity id in each of the entity update request models? It is consistent with all the other Update functions in the current API refactor branch:
```
type Client interface {
	UpdateEnvironment(req models.UpdateEnvironmentRequest) (string, error)
	UpdateLoadBalancer(req models.UpdateLoadBalancerRequest) (string, error)
	UpdateService(req models.UpdateServiceRequest) (string, error)
        ...
}
```

I think we might've visited this with the original discussion with env links change. If we remove the entity id, we would need to create job request specific models. Which would be super tedious to translate from one model to another almost identical looking model just for the job runner.

common/models/update_environment_request.go
sesh-kebab: See above.

api/provider/aws/environment_update.go
```diff
@@ -33,6 +39,87 @@ func (e *EnvironmentProvider) Update(req models.UpdateEnvironmentRequest) error
...
+				envID := delLayer0Prefix(e.Config.Instance(), aws.StringValue(sg.GroupName))
+				envID = strings.TrimRight(envID, "-env")
+
+				secGrpIDtoEnvID[groupID] = envID
+				addRemoveLinks[groupID] = false
+			}
+		}
+
+		// diff request with current linked environments
+		for _, environmentID := range *req.Links {
```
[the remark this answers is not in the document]
sesh-kebab: Definitely agree it is hard to follow - will refactor. ✅

api/provider/aws/environment_update.go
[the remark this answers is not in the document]
sesh-kebab: ✅ agreed - was being used during development.

cli/command/environment.go
sesh-kebab: I like this 👍

cli/command/environment.go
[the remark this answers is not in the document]
zpatrick: I bring it up because that was the logic for using `--no-wait` over `--wait=true`

client/environment.go
zpatrick: right - we hit on this in slack quite a bit so will refer to that

sesh-kebab: Yes, we could do this to be more RESful but PATCH requests would need to be constructed differently to the PUT requests as a PATCH contains a description of the changes rather than the changed state i.e., the request for adding a link would contain just the addition and look something like
```
[
  {
    "action": "addLink",
    "link": "environment_id"
  }
]

```

That would require significant changes server-side and as you mentioned previously, add more responsibilities to the Terraform Provider. Instead we could change the path to the below to keep supporting PUT and be more RESTful.
Instead of 
```
PATCH /environment/:id
{}
```

```
PUT /environment/:id/links/
["eid1", "eid2", "eid3"]

PUT /environment/:id/minclustercount/
{"1"}
```

And could lead to:
```
type UpdateEnvironmentLinksJob struct {
    EnvironmentID string
    Req                   []string
}

type UpdateEnvironmentMinCountJob struct {
    EnvironmentID string
    Req                   int
}
```

Making it more complicated by trying to be more RESTful? @zpatrick

api/provider/aws/environment_update.go
```diff
@@ -33,6 +39,87 @@ func (e *EnvironmentProvider) Update(req models.UpdateEnvironmentRequest) error
...
+		addRemoveLinks := map[string]bool{}
+		secGrpIDtoEnvID := map[string]string{}
+
+		environmentSG, err := readSG(e.AWS.EC2, getEnvironmentSGName(fqEnvironmentID))
+		if err != nil {
+			return err
+		}
+
+		// check which environments are currently linked
+		for _, ipPermission := range environmentSG.IpPermissions {
```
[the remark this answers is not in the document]
sesh-kebab: If we rely on the just the tags however, it could lead to a situation where the actual state can drift from our tags store. We need at least a way to verify the current state is as is right? If so I would have thought we can rely on making the assumptions being made here as we appending custom rules to security groups linking other security groups created via l0 is something that we expose (apart from the links).

api/provider/aws/common.go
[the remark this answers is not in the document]
sesh-kebab: ✅

api/provider/aws/environment_update.go
```diff
@@ -52,3 +139,175 @@ func (e *EnvironmentProvider) updateASGSize(autoScalingGroupName string, minSize
...
+			return nil
+		}
+
+		return err
+	}
+
+	return nil
+}
+
+func (e *EnvironmentProvider) createLinkTags(sourceEnvironmentID, destEnvironmentID string) error {
```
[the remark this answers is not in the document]
sesh-kebab: Too true ✅

api/provider/aws/environment_update.go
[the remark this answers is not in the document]
zpatrick: There's pros and cons to each, for sure. I just think it would be too costly/complex to verify/map an environment's security group ingress rules to links vs. just going with tags. If it's any consolation, we are relying on tags as the source of truth for task and deploy listing (via the `arn` tags)

api/provider/aws/environment_update.go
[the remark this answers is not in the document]
zpatrick: ah i see - maybe as a [nit] we rename this to `describeSGByID` to avoid potential confusion in the future?

client/environment.go
zpatrick: hmmm - we need to also consider keeping this consistent w/ the other entity updates. What if we we keep the method + paths you laid out, but the controller layer does the conversion of turning those request into generic`UpdateEntityJob` structs (which are just renamed versions of `UpdateEntityRequest`). That way we are keeping the `EntityUpdate` method the same in the `provider` layer.

sesh-kebab: Sure - I think that isn't too much responsibility at for the Controller to handle. ✅

api/provider/aws/environment_update.go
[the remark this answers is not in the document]
sesh-kebab: ok - will update it to rely on tags instead. ✅

api/provider/aws/environment_update.go
```diff
@@ -33,6 +39,93 @@ func (e *EnvironmentProvider) Update(req models.UpdateEnvironmentRequest) error
...
+		}
+
+		// check which environments are currently linked
+		currentTags, err := e.TagStore.SelectByTypeAndID("environment", environmentID)
+		if err != nil {
+			return err
+		}
+
+		currentLinkedEnvIDs := []string{}
+		for _, tag := range currentTags.WithKey("link") {
```
zpatrick: This would overwrite `currentLinkedEnvIDs`. Since tag keys are unique, you can write this as:
```
if tag, ok := currentTags.WithKey("link").First(); ok {
    currentLinkedEnvIDs = strings.Split(tag.Value, ",")
}
```

api/provider/aws/environment_update.go
```diff
@@ -33,6 +39,93 @@ func (e *EnvironmentProvider) Update(req models.UpdateEnvironmentRequest) error
...
+		checkForAWSError := func(err error) error {
+			if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "DoesNotExist" {
+				log.Printf("[WARN] skipping environment link/unlink since security group '%s' does not exist\n", getEnvironmentSGName(fqEnvironmentID))
+				return nil
+			}
+
+			return err
+		}
+
+		contains := func(value string, values []string) bool {
```
zpatrick: this should probably be a standalone somewhere else. Maybe bottom of this file or in `helpers.go`

api/provider/aws/environment_update.go
```diff
@@ -33,6 +39,93 @@ func (e *EnvironmentProvider) Update(req models.UpdateEnvironmentRequest) error
...
+		if err != nil {
+			return err
+		}
+
+		currentLinkedEnvIDs := []string{}
+		for _, tag := range currentTags.WithKey("link") {
+			currentLinkedEnvIDs = strings.Split(tag.Value, ",")
+		}
+
+		checkForAWSError := func(err error) error {
```
zpatrick: you probably don't want to check for errors this way, will explain more below

api/provider/aws/environment_update.go
```diff
@@ -33,6 +39,93 @@ func (e *EnvironmentProvider) Update(req models.UpdateEnvironmentRequest) error
...
+
+		// add - check if current link state contains links in request
+		for _, destEnvironmentID := range *req.Links {
+			if contains(destEnvironmentID, currentLinkedEnvIDs) {
+				continue
+			}
+
+			fqDestEnvID := addLayer0Prefix(e.Config.Instance(), destEnvironmentID)
+			sg, err := readSG(e.AWS.EC2, getEnvironmentSGName(fqDestEnvID))
+
```
zpatrick: remove newline

api/provider/aws/environment_update.go
```diff
@@ -33,6 +39,93 @@ func (e *EnvironmentProvider) Update(req models.UpdateEnvironmentRequest) error
...
+			if err != nil {
+				return checkForAWSError(err)
```
zpatrick: this would return prematurely, when we actually don't want to return if `checkForAWSError()` return `nil` - we want to continue. Also, the check for idempontence should actually not occur here, since we do want to error when we can't add a security group that doesn't exist.

api/provider/aws/environment_update.go
```diff
@@ -33,6 +39,93 @@ func (e *EnvironmentProvider) Update(req models.UpdateEnvironmentRequest) error
...
+			}
+
+			sourceSGID := aws.StringValue(sourceEnvSG.GroupId)
+			destSGID := aws.StringValue(sg.GroupId)
+
+			if err := e.createIngressInput(sourceSGID, destSGID); err != nil {
+				return err
+			}
+
+			if err := e.setLinkTags(req.EnvironmentID, *req.Links); err != nil {
```
zpatrick: this should only be called once, not within the loop. If you do keep this within the loop, then (contrary to what I said in the previous PR), this should only append to the existing `link` tag.

api/provider/aws/environment_update.go
```diff
@@ -33,6 +39,93 @@ func (e *EnvironmentProvider) Update(req models.UpdateEnvironmentRequest) error
...
+
+		// remove - check if links in request are missing from current links
+		for _, destEnvironmentID := range currentLinkedEnvIDs {
+			if contains(destEnvironmentID, *req.Links) {
+				continue
+			}
+
+			fqDestEnvID := addLayer0Prefix(e.Config.Instance(), destEnvironmentID)
+			sg, err := readSG(e.AWS.EC2, getEnvironmentSGName(fqDestEnvID))
+
```
zpatrick: remove newline

api/provider/aws/environment_update.go
```diff
@@ -33,6 +39,93 @@ func (e *EnvironmentProvider) Update(req models.UpdateEnvironmentRequest) error
...
+			if err != nil {
+				return checkForAWSError(err)
```
zpatrick: again, don't want to return here if err is nil. Probably would be best to not use a helper for this.

api/provider/aws/environment_update.go
```diff
@@ -33,6 +39,93 @@ func (e *EnvironmentProvider) Update(req models.UpdateEnvironmentRequest) error
...
+			}
+
+			sourceSGID := aws.StringValue(sourceEnvSG.GroupId)
+			destSGID := aws.StringValue(sg.GroupId)
+
+			if err := e.removeIngressRule(sourceSGID, destSGID); err != nil {
+				return err
+			}
+
+			if err := e.setLinkTags(req.EnvironmentID, *req.Links); err != nil {
```
zpatrick: relates to comment above

cli/command/environment.go
```diff
@@ -192,10 +208,8 @@ func (e *EnvironmentCommand) update(c *cli.Context) error {
 		return err
 	}
 
-	req := models.UpdateEnvironmentRequest{
-		EnvironmentID:   id,
-		MinClusterCount: &minClusterCount,
-	}
+	req := models.UpdateEnvironmentRequest{EnvironmentID: id}
```
zpatrick: This struct should not have the `EnvironentID` field

cli/command/environment.go
```diff
@@ -217,42 +231,52 @@ func (e *EnvironmentCommand) update(c *cli.Context) error {
...
+	return e.updateEnvironmentLinksHelper(c, generateAddLinkRequest)
+}
 
-	req := models.CreateEnvironmentLinkRequest{
-		SourceEnvironmentID: sourceEnvironmentID,
-		DestEnvironmentID:   destEnvironmentID,
-	}
+func (e *EnvironmentCommand) unlink(c *cli.Context) error {
+	generateRemoveLinkRequest := func(sourceEnvironmentID, destEnvironmentID string) (models.UpdateEnvironmentRequest, error) {
+		env, err := e.client.ReadEnvironment(sourceEnvironmentID)
```
zpatrick: my old example didn't require the helper call `ReadEnvironment`

cli/command/environment.go
```diff
@@ -217,42 +231,52 @@ func (e *EnvironmentCommand) update(c *cli.Context) error {
...
+		if err != nil {
+			return models.UpdateEnvironmentRequest{}, err
+		}
 
-	jobID, err := e.client.CreateLink(req)
-	if err != nil {
-		return err
+		destEnvIDIndex := -1
+		for i, link := range env.Links {
```
zpatrick: see old example for how to delete while iterating over a list

cli/command/environment.go
```diff
@@ -217,42 +231,52 @@ func (e *EnvironmentCommand) update(c *cli.Context) error {
...
+				destEnvIDIndex = i
+				break
+			}
+		}
+
+		if destEnvIDIndex >= 0 {
+			env.Links = append(env.Links[:destEnvIDIndex], env.Links[destEnvIDIndex+1:]...)
+		}
+
+		req := models.UpdateEnvironmentRequest{EnvironmentID: sourceEnvironmentID}
```
zpatrick: `EnvironmentID` field should not exist

cli/command/environment.go
```diff
@@ -217,42 +231,52 @@ func (e *EnvironmentCommand) update(c *cli.Context) error {
 }
 
 func (e *EnvironmentCommand) link(c *cli.Context) error {
```
zpatrick: this is really hard to read - i'll try to comb through this but will need to come back to it later

cli/command/environment_test.go
```diff
@@ -338,19 +296,93 @@ func TestEnvironmentSetMinCount(t *testing.T) {
...
+		base.Client.EXPECT().
+			ReadEnvironment("env_id1").
+			Return(&models.Environment{
+				EnvironmentID: "env_id1",
+				Links:         []string{},
+			}, nil)
+
+		base.Client.EXPECT().
+			ReadEnvironment("env_id2").
+			Return(&models.Environment{
```
zpatrick: we typically break-out these objects into their own variables if they become multi-line. It's pretty hard to read in the current format.

cli/command/environment_test.go
```diff
@@ -338,19 +296,93 @@ func TestEnvironmentSetMinCount(t *testing.T) {
...
 			Status: job.Completed.String(),
-			Result: "entity_id",
+			Result: "env_id1",
+		}
+
+		req1 := models.UpdateEnvironmentRequest{EnvironmentID: "env_id1"}
+		req1.Links = &links1
+
+		base.Client.EXPECT().
+			UpdateEnvironment(req1).
```
zpatrick: Updates should take `entityID string` as first param.

cli/command/environment_test.go
```diff
@@ -338,19 +296,93 @@ func TestEnvironmentSetMinCount(t *testing.T) {
...
+
+			base.Client.EXPECT().
+				ReadJob(job2.JobID).
+				Return(job2, nil)
+		}
+
+		f := Flags{
+			"bi-directional": true,
+		}
+		c := NewContext(t, []string{"env_name1", "env_name2"}, f, SetNoWait(!wait))
```
zpatrick: [nit] newline after closing bracket

api/provider/aws/environment_update.go
sesh-kebab: There would only ever be a single `link` tag for an environment, so that loop would only ever execute once. But First makes that more explicit so ✅

api/provider/aws/environment_update.go
sesh-kebab: Moved to `helpers.go` ✅

api/provider/aws/environment_update.go
sesh-kebab: line 86? [nit]? if not why?

api/provider/aws/environment_update.go
sesh-kebab: I think we should actually return the error here without checking for the DoesNotExist as we're adding links in this loop. If the destination environment doesn't have an associated security group, we can't actually proceed so not sure what you mean when you say "we want to continue"?

I've updated it to
```
if err := e.createIngressInput(sourceSGID, destSGID); err != nil {
	return err
}
```

api/provider/aws/environment_update.go
sesh-kebab: moved out of the loop ✅

api/provider/aws/environment_update.go
sesh-kebab: line 105? [nit]? if not why?

api/provider/aws/environment_update.go
sesh-kebab: Ok - I think I was only thinking of a single link being deleted but in the case of multiple links being deleted this should be a continue. Moved out of helper and updated to continue ✅

api/provider/aws/environment_update.go
sesh-kebab: ✅

cli/command/environment.go
sesh-kebab: ? can you elaborate - not having environment id here would require changing the client interface for all the entities.

cli/command/environment.go
sesh-kebab: ? can you elaborate - not having environment id here would require changing the client interface for all the entities.

cli/command/environment_test.go
sesh-kebab: That's new. That would require changing the client interface for all the entities.

cli/command/environment.go
sesh-kebab: ok - not sure if you're asking me to copy your example verbatim? The change to refactor link and unlink in the command and provider layer were underway and I referenced your example on the WIP PR but I didn't copy it.

api/provider/aws/environment_update.go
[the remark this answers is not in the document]
sesh-kebab: :white

cli/command/environment.go
sesh-kebab: Oh yes! missed this :(. didn't really understand how your example was working so wanted to try it out in go playground before updating to it. pretty slick but still not sure how this works.
```
for i, link := range env.Links {
	if link == destEnvironmentID {
		links = append(env.Links[:i], env.Links[i+1:]...)
		i--
	}
}
```

https://play.golang.org/p/eWULErPdTD wouldn't deleting the last item always generate index out of bounds error?

cli/command/environment_test.go

cli/command/environment.go
sesh-kebab: I think it is an improvement over what was there during WIP stage, but always open to iterate on it to make it more readable.

cli/command/environment_test.go

cli/command/environment.go
sesh-kebab: Oh I see now - looks like your example had an error. It should be 
```
links = append(env.Links[i:], env.Links[i+1:]...)

// instead of 
links = append(env.Links[:i], env.Links[i+1:]...)
```

cli/command/environment.go
```diff
@@ -217,42 +231,47 @@ func (e *EnvironmentCommand) update(c *cli.Context) error {
...
 
-	return e.waitOnJobHelper(c, jobID, "linking", func(environmentID string) error {
-		e.printer.Printf("Environment successfully linked\n")
-		return nil
-	})
+	return e.updateEnvironmentLinksHelper(c, generateRemoveLinkRequest)
 }
 
-func (e *EnvironmentCommand) unlink(c *cli.Context) error {
+func (e *EnvironmentCommand) updateEnvironmentLinksHelper(c *cli.Context,
```
zpatrick: when we do line wrapping with params, give each param a line please:
```
func foo(
    p1 int,
    p2 bool,
) error {
    ....
}
```

the last line is especially important as it gives a clear indentation difference between the return signature and body

cli/command/environment.go
```diff
@@ -269,21 +288,31 @@ func (e *EnvironmentCommand) unlink(c *cli.Context) error {
 	}
 
 	if sourceEnvironmentID == destEnvironmentID {
-		return fmt.Errorf("Cannot unlink an environment from itself")
+		return fmt.Errorf("Cannot link an environment to itself")
```
zpatrick: Shouldn't this be `link/unlink ...to/from` since this can happen in both scenarios?

cli/command/environment.go
zpatrick: It's doesn't have to be verbatim, but it does save you the read call in the "generateRequest" helpers

cli/command/environment.go
```diff
@@ -217,42 +231,47 @@ func (e *EnvironmentCommand) update(c *cli.Context) error {
...
+		env, err := e.client.ReadEnvironment(sourceEnvironmentID)
+		if err != nil {
+			return models.UpdateEnvironmentRequest{}, err
+		}
 
-	jobID, err := e.client.CreateLink(req)
-	if err != nil {
-		return err
+		updatedLinks := []string{}
+		for _, link := range env.Links {
```
zpatrick: [nit] you can save a few lines by removing the destEnvironmentID links in place

cli/command/environment.go
```diff
@@ -269,21 +288,31 @@ func (e *EnvironmentCommand) unlink(c *cli.Context) error {
...
 	if sourceEnvironmentID == destEnvironmentID {
-		return fmt.Errorf("Cannot unlink an environment from itself")
+		return fmt.Errorf("Cannot link an environment to itself")
 	}
 
-	req := models.DeleteEnvironmentLinkRequest{
-		SourceEnvironmentID: sourceEnvironmentID,
-		DestEnvironmentID:   destEnvironmentID,
-	}
+	addLinkFN := func(sourceEnvID, destEnvID string) error {
```
zpatrick: a more appropriate name would probably be something like `updateLinkFN`, given that this could add or remove links

cli/command/environment.go
```diff
@@ -269,21 +288,31 @@ func (e *EnvironmentCommand) unlink(c *cli.Context) error {
...
+		updateEnvReq, err := generateReq(sourceEnvID, destEnvID)
+		if err != nil {
+			return err
+		}
 
-	jobID, err := e.client.DeleteLink(req)
-	if err != nil {
-		return err
+		jobID1, err := e.client.UpdateEnvironment(sourceEnvID, updateEnvReq)
```
zpatrick: [nit] there's only one job in scope, so `jobID` is probably a more appropriate var name here.

cli/command/environment.go
```diff
@@ -217,42 +231,47 @@ func (e *EnvironmentCommand) update(c *cli.Context) error {
...
+		if err != nil {
+			return models.UpdateEnvironmentRequest{}, err
+		}
 
-	sourceEnvironmentID, err := e.resolveSingleEntityIDHelper("environment", args["SOURCE_ENVIRONMENT_NAME"])
-	if err != nil {
-		return err
-	}
+		if !client.Contains(destEnvironmentID, env.Links) {
+			appendedLinks := append(env.Links, destEnvironmentID)
```
zpatrick: Wait, I'm a little confused here. Does this build?

```
if !client.Contains(destEnvironmentID, env.Links) {
    appendedLinks := append(env.Links, destEnvironmentID)
}

req := models.UpdateEnvironmentRequest{Links: &appendedLinks}
```

since `appendedLinks` is declared within the `if` statement, isn't is out of scope when you try to reference it in `models.UpdateEnvironmentRequest`?

cli/command/environment_test.go
zpatrick: looks like this still needs to be added

client/service.go
```diff
@@ -56,9 +56,10 @@ func (c *APIClient) ReadServiceLogs(serviceID string, query url.Values) ([]*mode
 	return logs, nil
 }
 
-func (c *APIClient) UpdateService(req models.UpdateServiceRequest) (string, error) {
+func (c *APIClient) UpdateService(serviceID string, req models.UpdateServiceRequest) (string, error) {
 	var resp models.CreateJobResponse
-	if err := c.client.Put("/service", req, &resp); err != nil {
+	path := fmt.Sprintf("/service/%s", serviceID)
+	if err := c.client.Put(path, req, &resp); err != nil {
```
tlake: This should be `c.client.Patch()`, right?

cli/command/environment_test.go
```diff
@@ -401,18 +519,29 @@ func TestEnvironmentUnlink(t *testing.T) {
 			Resolve("environment", "env_name2").
 			Return([]string{"env_id2"}, nil)
 
-		req := models.DeleteEnvironmentLinkRequest{
-			SourceEnvironmentID: "env_id1",
-			DestEnvironmentID:   "env_id2",
+		base.Client.EXPECT().
+			ReadEnvironment("env_id1").
+			Return(&models.Environment{
```
zpatrick: I'll add this to ROT, but if variables span multiple lines in our `gomock` chains, let's declare them outside of the chain.

client/load_balancer.go
```diff
@@ -44,9 +44,10 @@ func (c *APIClient) ReadLoadBalancer(loadBalancerID string) (*models.LoadBalance
 	return loadbalancer, nil
 }
 
-func (c *APIClient) UpdateLoadBalancer(req models.UpdateLoadBalancerRequest) (string, error) {
+func (c *APIClient) UpdateLoadBalancer(loadBalancerID string, req models.UpdateLoadBalancerRequest) (string, error) {
 	var resp models.CreateJobResponse
-	if err := c.client.Put("/loadbalancer", req, &resp); err != nil {
+	path := fmt.Sprintf("/loadbalancer/%s", loadBalancerID)
+	if err := c.client.Put(path, req, &resp); err != nil {
```
tlake: This should be `c.client.Patch()`, right?

sesh-kebab: ✅ oops

client/service.go
sesh-kebab: ✅

client/helpers_test.go
```diff
@@ -177,3 +177,36 @@ func TestWaitForJobTimeout(t *testing.T) {
 		t.Fatal("Error was nil!")
 	}
 }
+
+func TestContains(t *testing.T) {
```
zpatrick: good test! - but- for future reference, we probably don't need unit tests for helpers that are this simple

client/service.go
zpatrick: Should this be `Patch`, or is that coming in a future pr?

api/provider/aws/environment_update.go
```diff
@@ -33,6 +38,75 @@ func (e *EnvironmentProvider) Update(req models.UpdateEnvironmentRequest) error
...
+		}
+
+		currentLinkedEnvIDs := []string{}
+		if tag, ok := currentTags.WithKey("link").First(); ok {
+			currentLinkedEnvIDs = strings.Split(tag.Value, ",")
+		}
+
+		// add - check if current link state contains links in request
+		for _, destEnvironmentID := range *req.Links {
+			if client.Contains(destEnvironmentID, currentLinkedEnvIDs) {
```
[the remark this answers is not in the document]
sesh-kebab: I thought you meant move it to client\helper.go? I'm using it in the command package as well. But also to your point below, will move it to `layer0/common/models/tags.go`
✅

api/provider/aws/environment_update.go
```diff
@@ -33,6 +38,75 @@ func (e *EnvironmentProvider) Update(req models.UpdateEnvironmentRequest) error
...
+			if client.Contains(destEnvironmentID, *req.Links) {
+				continue
+			}
+
+			fqDestEnvID := addLayer0Prefix(e.Config.Instance(), destEnvironmentID)
+			sg, err := readSG(e.AWS.EC2, getEnvironmentSGName(fqDestEnvID))
+
+			if err != nil {
+				if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "DoesNotExist" {
+					log.Printf("[WARN] skipping environment link/unlink since security group '%s' does not exist\n", getEnvironmentSGName(fqEnvironmentID))
```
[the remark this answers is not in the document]
sesh-kebab: ✅

common/models/update_environment_request.go
```diff
@@ -6,16 +6,17 @@ import (
 	swagger "github.com/zpatrick/go-plugin-swagger"
 )
 
+type UpdateEnvironmentRequestJob struct {
+	EnvironmentID string `json:"environment_id"`
```
diemonster: Just commenting on this decision as it runs slightly contrary to what I had previously understood about the implementation. Looks good!

client/helpers_test.go
sesh-kebab: Yea I thought it was overkill writing a test for a basically a for loop too. Making up for not writing tests all those years 😅

api/provider/aws/environment_update.go
```diff
@@ -52,3 +126,78 @@ func (e *EnvironmentProvider) updateASGSize(autoScalingGroupName string, minSize
...
+		}
+
+		return err
+	}
+
+	return nil
+}
+
+func (e *EnvironmentProvider) setLinkTags(environmentID string, links []string) error {
+	uniqueLinks := []string{}
```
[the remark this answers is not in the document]
sesh-kebab: I thought the same thing when I found `StringSlice` in the `sort` package. But that seemed overkill too. But as you suggested it, it would make it better. Helper methods like Contains doesn't really belong anywhere (on a sidenote: kind of shocking that none go standard package doesn't contain a `contains` on an unordered list).

cli/command/environment.go
sesh-kebab: woah - this is really bad. I thought I was making a minor change (apparently not). Fixed! ✅

cli/command/environment.go
sesh-kebab: yup ✅

cli/command/environment.go
sesh-kebab: Agreed. ✅

cli/command/environment.go

cli/command/environment_test.go

common/models/update_environment_request.go
sesh-kebab: 😈

cli/command/environment.go
sesh-kebab: that looks soo much better! ✅

api/provider/aws/environment_update.go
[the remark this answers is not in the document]
sesh-kebab: ✅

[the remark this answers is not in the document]
zpatrick: i know right? oh well 🤷‍♂️

common/models/link_tag.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+func (l LinkTags) Distinct() []string {
+	m := map[string]int{}
+
+	for idx, k := range l {
+		m[k] = idx
+	}
+
+	d := []string{}
```
zpatrick: [super nit] for future reference, if we needed to be super high perf we'd use `d := make([]string, len(m))` and use `d[i] = k` instead of `append`

sesh-kebab: oops - i commented on @nehayward review about the same thing yesterday and then ended up doing the same thing myself. will fix ✅

PR #414, opened 2017-11-03
Unit test task

nehayward:
Initial pass at unit test for aws provider.

**Checklist**
- [x] Unit tests

closes #

### DIFF
--- a/api/provider/aws/common.go
+++ b/api/provider/aws/common.go
@@ -82,7 +82,7 @@ func lookupTaskDefinitionARNFromDeployID(store tag.Store, deployID string) (stri
 func lookupEntityEnvironmentID(store tag.Store, entityType, entityID string) (string, error) {
 	tags, err := store.SelectByTypeAndID(entityType, entityID)
 	if err != nil {
-		return "ts", err
+		return "", err
 	}
 
 	if len(tags) == 0 {
--- a/api/provider/aws/common.go
+++ b/api/provider/aws/common.go
@@ -82,7 +82,7 @@ func lookupTaskDefinitionARNFromDeployID(store tag.Store, deployID string) (stri
 func lookupEntityEnvironmentID(store tag.Store, entityType, entityID string) (string, error) {
 	tags, err := store.SelectByTypeAndID(entityType, entityID)
 	if err != nil {
-		return "", err
+		return "ts", err
 	}
 
 	if len(tags) == 0 {
--- a/api/provider/aws/task_delete.go
+++ b/api/provider/aws/task_delete.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/quintilesims/layer0/common/errors"
 )
 
 // Delete stops an ECS Task using the specified taskID. The taskID is used to look up the name of
@@ -15,13 +16,19 @@ func (t *TaskProvider) Delete(taskID string) error {
 	environmentID, err := lookupEntityEnvironmentID(t.TagStore, "task", taskID)
 	if err != nil {
 		log.Printf("[WARN] Environment not found\n")
-		return nil
+		if err, ok := err.(*errors.ServerError); ok && err.Code == errors.TaskDoesNotExist {
+			return nil
+		}
+		return err
 	}
 
 	taskARN, err := t.lookupTaskARN(taskID)
 	if err != nil {
 		log.Printf("[WARN] Task not found\n")
-		return nil
+		if err, ok := err.(*errors.ServerError); ok && err.Code == errors.TaskDoesNotExist {
+			return nil
+		}
+		return err
 	}
 
 	fqEnvironmentID := addLayer0Prefix(t.Config.Instance(), environmentID)
--- a/api/provider/aws/task_delete.go
+++ b/api/provider/aws/task_delete.go
@@ -15,8 +15,8 @@ import (
 func (t *TaskProvider) Delete(taskID string) error {
 	environmentID, err := lookupEntityEnvironmentID(t.TagStore, "task", taskID)
 	if err != nil {
-		log.Printf("[WARN] Environment not found\n")
 		if err, ok := err.(*errors.ServerError); ok && err.Code == errors.TaskDoesNotExist {
+			log.Printf("[WARN] Environment not found\n")
 			return nil
 		}
 		return err
@@ -24,8 +24,8 @@ func (t *TaskProvider) Delete(taskID string) error {
 
 	taskARN, err := t.lookupTaskARN(taskID)
 	if err != nil {
-		log.Printf("[WARN] Task not found\n")
 		if err, ok := err.(*errors.ServerError); ok && err.Code == errors.TaskDoesNotExist {
+			log.Printf("[WARN] Task not found\n")
 			return nil
 		}
 		return err
--- a/api/provider/aws/task_delete.go
+++ b/api/provider/aws/task_delete.go
@@ -15,11 +15,13 @@ func (t *TaskProvider) Delete(taskID string) error {
 	environmentID, err := lookupEntityEnvironmentID(t.TagStore, "task", taskID)
 	if err != nil {
 		log.Printf("[WARN] Environment not found\n")
+		return nil
 	}
 
 	taskARN, err := t.lookupTaskARN(taskID)
 	if err != nil {
 		log.Printf("[WARN] Task not found\n")
+		return nil
 	}
 
 	fqEnvironmentID := addLayer0Prefix(t.Config.Instance(), environmentID)
--- a/api/provider/aws/task_delete.go
+++ b/api/provider/aws/task_delete.go
@@ -16,7 +16,7 @@ func (t *TaskProvider) Delete(taskID string) error {
 	environmentID, err := lookupEntityEnvironmentID(t.TagStore, "task", taskID)
 	if err != nil {
 		if err, ok := err.(*errors.ServerError); ok && err.Code == errors.TaskDoesNotExist {
-			log.Printf("[WARN] Environment not found\n")
+			log.Printf("[WARN] Task not found\n")
 			return nil
 		}
 		return err
--- a/api/provider/aws/task_delete.go
+++ b/api/provider/aws/task_delete.go
@@ -1,6 +1,7 @@
 package aws
 
 import (
+	"log"
 	"strings"
 
 	"github.com/aws/aws-sdk-go/aws/awserr"
@@ -13,12 +14,12 @@ import (
 func (t *TaskProvider) Delete(taskID string) error {
 	environmentID, err := lookupEntityEnvironmentID(t.TagStore, "task", taskID)
 	if err != nil {
-		return err
+		log.Printf("[WARN] Environment not found\n")
 	}
 
 	taskARN, err := t.lookupTaskARN(taskID)
 	if err != nil {
-		return err
+		log.Printf("[WARN] Task not found\n")
 	}
 
 	fqEnvironmentID := addLayer0Prefix(t.Config.Instance(), environmentID)
--- a/api/provider/aws/test_aws/task_create_test.go
+++ b/api/provider/aws/test_aws/task_create_test.go
@@ -1,0 +1,150 @@
+package test_aws
+
+import (
+	"testing"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/golang/mock/gomock"
+	provider "github.com/quintilesims/layer0/api/provider/aws"
+	"github.com/quintilesims/layer0/api/tag"
+	awsc "github.com/quintilesims/layer0/common/aws"
+	"github.com/quintilesims/layer0/common/config/mock_config"
+	"github.com/quintilesims/layer0/common/models"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTaskCreate(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockAWS := awsc.NewMockClient(ctrl)
+	tagStore := tag.NewMemoryStore()
+	mockConfig := mock_config.NewMockAPIConfig(ctrl)
+
+	// todo: setup helper for config
+	mockConfig.EXPECT().Instance().Return("test").AnyTimes()
+
+	tags := models.Tags{
+		{
+			EntityID:   "tsk_id1",
+			EntityType: "task",
+			Key:        "name",
+			Value:      "tsk_name1",
+		},
+		{
+			EntityID:   "tsk_id1",
+			EntityType: "task",
+			Key:        "arn",
+			Value:      "arn1",
+		},
+		{
+			EntityID:   "tsk_id1",
+			EntityType: "task",
+			Key:        "id",
+			Value:      "env_id1",
+		},
+		{
+			EntityID:   "tsk_id2",
+			EntityType: "task",
+			Key:        "name",
+			Value:      "tsk_name2",
+		},
+		{
+			EntityID:   "tsk_id2",
+			EntityType: "task",
+			Key:        "arn",
+			Value:      "arn2",
+		},
+		{
+			EntityID:   "tsk_id2",
+			EntityType: "task",
+			Key:        "id",
+			Value:      "env_id2",
+		},
+		{
+			EntityID:   "env_id1",
+			EntityType: "environment",
+			Key:        "name",
+			Value:      "env_name1",
+		},
+		{
+			EntityID:   "env_id2",
+			EntityType: "environment",
+			Key:        "name",
+			Value:      "env_name2",
+		},
+	}
+
+	for _, tag := range tags {
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	// ensure we filter out clusters that don't belong to our instance
+	// TODo: Make Helper?
+	listClusterPagesFN := func(input *ecs.ListClustersInput, fn func(output *ecs.ListClustersOutput, lastPage bool) bool) error {
+		clusterARNs := []*string{
+			aws.String("arn:aws:ecs:region:012345678910:cluster/l0-test-env_id1"),
+			aws.String("arn:aws:ecs:region:012345678910:cluster/l0-test-env_id2"),
+			aws.String("arn:aws:ecs:region:012345678910:cluster/l0-bad-env_id1"),
+			aws.String("arn:aws:ecs:region:012345678910:cluster/bad2"),
+		}
+
+		output := &ecs.ListClustersOutput{}
+		output.SetClusterArns(clusterARNs)
+		fn(output, true)
+
+		return nil
+	}
+
+	mockAWS.ECS.EXPECT().
+		ListClustersPages(gomock.Any(), gomock.Any()).
+		Do(listClusterPagesFN).
+		Return(nil)
+
+	// ensure we filter out tasks that don't belong to our instance
+	listTaskPagesFN := func(input *ecs.ListTasksInput, fn func(output *ecs.ListTasksOutput, lastPage bool) bool) error {
+		taskARNs := []*string{
+			aws.String("arn:aws:ecs:region:012345678910:task/l0-test-tsk_id1"),
+			aws.String("arn:aws:ecs:region:012345678910:task/l0-test-tsk_id2"),
+			aws.String("arn:aws:ecs:region:012345678910:task/l0-bad-tsk_id1"),
+			aws.String("arn:aws:ecs:region:012345678910:task/bad2"),
+		}
+
+		output := &ecs.ListTasksOutput{}
+		output.SetTaskArns(taskARNs)
+		fn(output, true)
+
+		return nil
+	}
+
+	mockAWS.ECS.EXPECT().
+		ListTasksPages(gomock.Any(), gomock.Any()).
+		Do(listTaskPagesFN).
+		Return(nil)
+
+	target := provider.NewTaskProvider(mockAWS.Client(), tagStore, mockConfig)
+	result, err := target.List()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	expected := []models.TaskSummary{
+		{
+			TaskID:          "tsk_id1",
+			TaskName:        "tsk_name1",
+			EnvironmentID:   "env_id1",
+			EnvironmentName: "env_name1",
+		},
+		{
+			TaskID:          "tsk_id2",
+			TaskName:        "tsk_name2",
+			EnvironmentID:   "env_id2",
+			EnvironmentName: "env_name2",
+		},
+	}
+
+	assert.Equal(t, expected, result)
+}
--- a/api/provider/aws/test_aws/task_create_test.go
+++ b/api/provider/aws/test_aws/task_create_test.go
@@ -3,8 +3,6 @@ package test_aws
 import (
 	"testing"
 
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/service/ecs"
 	"github.com/golang/mock/gomock"
 	provider "github.com/quintilesims/layer0/api/provider/aws"
 	"github.com/quintilesims/layer0/api/tag"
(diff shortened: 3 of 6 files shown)


api/provider/aws/test_aws/task_create_test.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/golang/mock/gomock"
+	provider "github.com/quintilesims/layer0/api/provider/aws"
+	"github.com/quintilesims/layer0/api/tag"
+	awsc "github.com/quintilesims/layer0/common/aws"
+	"github.com/quintilesims/layer0/common/config/mock_config"
+	"github.com/quintilesims/layer0/common/models"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTaskCreate(t *testing.T) {
```
zpatrick: This file/function suggest a `Create` test, but this is a `List` test

api/provider/aws/test_aws/task_create_test.go
```diff
@@ -0,0 +1,150 @@
...
+		},
+	}
+
+	for _, tag := range tags {
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	// ensure we filter out clusters that don't belong to our instance
```
zpatrick: we don't need to assert this behavior in this test. This is handled by environment list test

api/provider/aws/test_aws/task_create_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+	mockAWS.ECS.EXPECT().
+		ListClustersPages(gomock.Any(), gomock.Any()).
+		Do(listClusterPagesFN).
+		Return(nil)
+
+	// ensure we filter out tasks that don't belong to our instance
+	listTaskPagesFN := func(input *ecs.ListTasksInput, fn func(output *ecs.ListTasksOutput, lastPage bool) bool) error {
+		taskARNs := []*string{
+			aws.String("arn:aws:ecs:region:012345678910:task/l0-test-tsk_id1"),
```
zpatrick: these are not what task arns look like, and they also don't match the arns you specified in tags above. They need to match because every arn returned here that doesn't have a matching tag gets filtered out. I'm not sure how this test is passing tbh because it should return an empty slice of task summaries with the current way it is setup.

api/provider/aws/test_aws/task_create_test.go
```diff
@@ -0,0 +1,150 @@
...
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	// ensure we filter out clusters that don't belong to our instance
+	// TODo: Make Helper?
+	listClusterPagesFN := func(input *ecs.ListClustersInput, fn func(output *ecs.ListClustersOutput, lastPage bool) bool) error {
+		clusterARNs := []*string{
+			aws.String("arn:aws:ecs:region:012345678910:cluster/l0-test-env_id1"),
```
zpatrick: you need a way to relate the clusters returned here to the `ListTaskPages` below. In the actual implementation, we call `ListTaskPages` for **each** cluster (not just once)

api/provider/aws/test_aws/task_delete_test.go
```diff
@@ -0,0 +1,97 @@
...
+		{
+			EntityID:   "tsk_id",
+			EntityType: "task",
+			Key:        "arn",
+			Value:      "arn1",
+		},
+		{
+			EntityID:   "tsk_id",
+			EntityType: "task",
+			Key:        "id",
```
zpatrick: `environment_id`

api/provider/aws/test_aws/task_create_test.go
```diff
@@ -0,0 +1,150 @@
...
+		{
+			EntityID:   "tsk_id2",
+			EntityType: "task",
+			Key:        "arn",
+			Value:      "arn2",
+		},
+		{
+			EntityID:   "tsk_id2",
+			EntityType: "task",
+			Key:        "id",
```
zpatrick: `environment_id`

api/provider/aws/test_aws/task_create_test.go
```diff
@@ -0,0 +1,150 @@
...
+		{
+			EntityID:   "tsk_id1",
+			EntityType: "task",
+			Key:        "arn",
+			Value:      "arn1",
+		},
+		{
+			EntityID:   "tsk_id1",
+			EntityType: "task",
+			Key:        "id",
```
zpatrick: `environment_id`

api/provider/aws/test_aws/task_delete_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	for _, tag := range tags {
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	stopTaskInput := &ecs.StopTaskInput{}
+	stopTaskInput.SetCluster("l0-test-env_id")
+	stopTaskInput.SetTask("l0-test-tsk_id")
```
zpatrick: We pass in the task arn, not the task id into `StopTask` (https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/task_delete.go#L26). I'm not sure how this test is passing

api/provider/aws/test_aws/task_delete_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	mockAWS.ECS.EXPECT().
+		StopTask(stopTaskInput).
+		Return(nil, nil)
```
zpatrick: `Return(&ecs.StopTaskOutput{}, nil)`

api/provider/aws/test_aws/task_delete_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	target := provider.NewTaskProvider(mockAWS.Client(), tagStore, mockConfig)
+	err := target.Delete("tsk_id")
```
zpatrick: `if err := ...; err != nil { ... }`

api/provider/aws/test_aws/task_delete_test.go
```diff
@@ -0,0 +1,97 @@
...
+	mockAWS := awsc.NewMockClient(ctrl)
+	tagStore := tag.NewMemoryStore()
+	mockConfig := mock_config.NewMockAPIConfig(ctrl)
+
+	// todo: setup helper for config
+	mockConfig.EXPECT().Instance().Return("test").AnyTimes()
+
+	mockAWS.ECS.EXPECT().
+		StopTask(gomock.Any()).
+		Return(nil, awserr.New("TaskDoesNotExist", "", nil))
```
zpatrick: This doesn't check that error code, it checks the message: https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/task_delete.go#L47

api/provider/aws/test_aws/task_list_test.go
```diff
@@ -0,0 +1,153 @@
...
+		Do(listClusterPagesFN).
+		Return(nil)
+
+	for _, environmentID := range []string{"l0-test-env_id1", "l0-test-env_id2"} {
+		listTaskPagesFN := func(input *ecs.ListTasksInput, fn func(output *ecs.ListTasksOutput, lastPage bool) bool) error {
+			taskARNs := []*string{
+				aws.String("arn:aws:ecs:region:012345678910:task/arn1"),
+				aws.String("arn:aws:ecs:region:012345678910:task/arn2"),
+			}
+			output := &ecs.ListTasksOutput{}
```
zpatrick: [nit] newline after closing bracket

api/provider/aws/test_aws/task_list_test.go
```diff
@@ -0,0 +1,153 @@
...
+		return nil
+	}
+
+	mockAWS.ECS.EXPECT().
+		ListClustersPages(gomock.Any(), gomock.Any()).
+		Do(listClusterPagesFN).
+		Return(nil)
+
+	for _, environmentID := range []string{"l0-test-env_id1", "l0-test-env_id2"} {
+		listTaskPagesFN := func(input *ecs.ListTasksInput, fn func(output *ecs.ListTasksOutput, lastPage bool) bool) error {
```
zpatrick: This is on the right track, but a little wonky if you think about it. The scenario setup here is we have 2 environments in our layer0 instance, and each environment has 2 tasks but they have the exact same arns. This could never happen in real life (each task's arn is unique), so we should consider changing this a bit, e.g. cluster1 returns arn1, cluster2 returns arn2

api/provider/aws/test_aws/task_list_test.go
```diff
@@ -0,0 +1,153 @@
...
+				aws.String("arn:aws:ecs:region:012345678910:task/arn2"),
+			}
+			output := &ecs.ListTasksOutput{}
+			output.SetTaskArns(taskARNs)
+			fn(output, true)
+
+			return nil
+		}
+
+		for _, status := range []string{ecs.DesiredStatusRunning, ecs.DesiredStatusStopped} {
```
zpatrick: 👍 nice

api/provider/aws/test_aws/task_read_test.go
```diff
@@ -0,0 +1,119 @@
...
+		{
+			EntityID:   "tsk_id1",
+			EntityType: "task",
+			Key:        "arn",
+			Value:      "arn1",
+		},
+		{
+			EntityID:   "tsk_id1",
+			EntityType: "task",
+			Key:        "id",
```
zpatrick: `environment_id`

api/provider/aws/test_aws/task_read_test.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	for _, tag := range tags {
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	// environments' asg names are the same as the fq environment id
+	describeASGInput := &autoscaling.DescribeAutoScalingGroupsInput{}
```
zpatrick: this doesn't make any sense to me - we don't describe autoscaling groups in a task read?

api/provider/aws/test_aws/task_read_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	lc := &autoscaling.LaunchConfiguration{}
+	lc.SetLaunchConfigurationName("lc_name")
+	lc.SetInstanceType("m3.small")
+	lc.SetImageId("ami_id")
+
+	describeLCOutput := &autoscaling.DescribeLaunchConfigurationsOutput{}
+	describeLCOutput.SetLaunchConfigurations([]*autoscaling.LaunchConfiguration{lc})
+
+	taskARNs := []*string{
```
zpatrick: we should only be expecting the task arn from the tag above

api/provider/aws/test_aws/task_read_test.go
```diff
@@ -0,0 +1,119 @@
...
+		aws.String("arn:aws:ecs:region:012345678910:task/l0-bad-tsk_id1"),
+		aws.String("arn:aws:ecs:region:012345678910:task/bad2"),
+	}
+
+	input := &ecs.DescribeTasksInput{}
+	// input.SetCluster()
+	input.SetTasks(taskARNs)
+
+	output := &ecs.DescribeTasksOutput{}
+	// output.SetTasks()
```
zpatrick: This definitely needs the output set https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/task_read.go#L66

api/provider/aws/test_aws/task_read_test.go
```diff
@@ -0,0 +1,119 @@
...
+		Return(output, nil)
+
+	target := provider.NewTaskProvider(mockAWS.Client(), tagStore, mockConfig)
+	result, err := target.Read("tsk_id")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// todo: Links
+	expected := &models.Task{}
```
zpatrick: You need a fully populated model (including a non-empty `Containers` field)

api/provider/aws/test_aws/task_delete_test.go
```diff
@@ -0,0 +1,97 @@
...
+		StopTask(stopTaskInput).
+		Return(nil, nil)
+
+	target := provider.NewTaskProvider(mockAWS.Client(), tagStore, mockConfig)
+	err := target.Delete("tsk_id")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	assert.Len(t, tagStore.Tags(), 0)
```
zpatrick: shouldn't the environment's name tag still exist?

api/provider/aws/test_aws/task_list_test.go
```diff
@@ -0,0 +1,153 @@
...
+		},
+	}
+
+	for _, tag := range tags {
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	//[l0-nick-demo91e6e283 l0-nick-api l0-nick-demoenve36e9 l0-nick-endtoenb8169]
```
zpatrick: remove debug stuff

api/provider/aws/test_aws/task_list_test.go
```diff
@@ -0,0 +1,153 @@
...
+		output := &ecs.ListClustersOutput{}
+		output.SetClusterArns(clusterARNs)
+
+		fn(output, true)
+
+		return nil
+	}
+
+	mockAWS.ECS.EXPECT().
+		ListClustersPages(gomock.Any(), gomock.Any()).
```
zpatrick: `ListClustersPages(&ecs.ListClustersPagesInput{}, gomock.Any()).`

api/provider/aws/test_aws/task_read_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	taskARNs := []*string{
+		aws.String("arn:aws:ecs:region:012345678910:task/l0-test-tsk_id1"),
+		aws.String("arn:aws:ecs:region:012345678910:task/l0-test-tsk_id2"),
+		aws.String("arn:aws:ecs:region:012345678910:task/l0-bad-tsk_id1"),
+		aws.String("arn:aws:ecs:region:012345678910:task/bad2"),
+	}
+
+	input := &ecs.DescribeTasksInput{}
+	// input.SetCluster()
```
zpatrick: needs to be set

api/provider/aws/test_aws/task_delete_test.go
jparsons04: I see only Task entity type tags defined above.

api/provider/aws/test_aws/task_delete_test.go
```diff
@@ -0,0 +1,94 @@
...
+	stopTaskInput.SetTask("arn:aws:ecs:region:012345678910:task/arn1")
+
+	mockAWS.ECS.EXPECT().
+		StopTask(stopTaskInput).
+		Return(nil, nil)
+
+	target := provider.NewTaskProvider(mockAWS.Client(), tagStore, mockConfig)
+	err := target.Delete("tsk_id")
+	if err != nil {
+		// t.Fatal(err)
```
jparsons04: This also needs the `if err := ...; err != nil { ... }` pattern. Also, it seems that this should be uncommented.

nehayward: This is something I wanted to ask about if you look at the task delete it reads the aws error message. I’m wondering if I should do the same or if we should change both. 
@zpatrick

jparsons04: Ah, I see now. So, yes, since this test is testing for idempotence, we're mainly trying to test the graceful handling of an error returned from the SDK. For instance, the SDK helper function [deleteLoadBalancer](https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/load_balancer_delete.go#L54) checks for the presence of a "NoSuchEntity" error code from the SDK and exits gracefully. Likewise [my EXPECT() call](https://github.com/quintilesims/layer0/pull/410/files#diff-ed65c448f68e560743c6c2a4999f9b22R93) ensures that this is happened.

The other helper functions that are called have similar error code checks. In your case, the idempotent check [looks for a specific string in the error message](https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/task_delete.go#L47). @zpatrick has a good example of how to handle this [here](https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/test_aws/environment_delete_test.go#L101)

api/provider/aws/test_aws/task_create_test.go
```diff
@@ -0,0 +1,106 @@
...
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	req := models.CreateTaskRequest{
+		DeployID:           "dpl_id",
+		EnvironmentID:      "env_id",
+		TaskName:           "tsk_name",
+		ContainerOverrides: []models.ContainerOverride{},
```
zpatrick: this should be a non zero-value/empty object so we make sure we map these params correctly

api/provider/aws/test_aws/task_delete_test.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+	}
+
+	stopTaskInput := &ecs.StopTaskInput{}
+	stopTaskInput.SetCluster("l0-test-env_id")
+	stopTaskInput.SetTask("arn:aws:ecs:region:012345678910:task/arn")
+
+	mockAWS.ECS.EXPECT().
+		StopTask(stopTaskInput).
+		Return(nil, nil)
```
zpatrick: `Return(&ecs.StopTaskOutput{}, nil)`

api/provider/aws/test_aws/task_list_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+		taskARNs := []*string{
+			aws.String("arn:aws:ecs:region:012345678910:task/arn2"),
+			aws.String("arn:aws:ecs:region:012345678910:task/arn1"),
+		}
+
+		taskARNs = append(taskARNs[:i], taskARNs[i+1:]...)
+
+		listTaskPagesFN := func(input *ecs.ListTasksInput, fn func(output *ecs.ListTasksOutput, lastPage bool) bool) error {
+
```
zpatrick: [nit] remove newline

api/provider/aws/test_aws/task_list_test.go
```diff
@@ -0,0 +1,156 @@
...
+		Return(nil)
+
+	for i, environmentID := range []string{"l0-test-env_id1", "l0-test-env_id2"} {
+
+		taskARNs := []*string{
+			aws.String("arn:aws:ecs:region:012345678910:task/arn2"),
+			aws.String("arn:aws:ecs:region:012345678910:task/arn1"),
+		}
+
+		taskARNs = append(taskARNs[:i], taskARNs[i+1:]...)
```
zpatrick: this was hard to parse, maybe we can simplify a bit:
```
for i, environmentID := range []string{"l0-test-env_id1", "l0-test-env_id2"} {
    taskARNs := []string{"arn:aws:ecs:region:012345678910:task/arn1"}
    if i == 1 {
        taskARNs = []string{"arn:aws:ecs:region:012345678910:task/arn2"}
    }
 
    ...
}
```

api/provider/aws/test_aws/task_read_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+	expected := &models.Task{
+		TaskID:        "tsk_id",
+		TaskName:      "tsk_name",
+		EnvironmentID: "env_id",
+		DeployID:      "dpl_id",
+		DeployName:    "dpl_name",
+		DeployVersion: "version",
+		Status:        "RUNNING",
+		Containers:    []models.Container{},
```
zpatrick: we need a zero-value/nil field here to ensure we map correctly.

api/provider/aws/test_aws/task_list_test.go
nehayward: Yeah I might have over complicated it.

api/provider/aws/test_aws/task_delete_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	mockAWS := awsc.NewMockClient(ctrl)
+	tagStore := tag.NewMemoryStore()
+	mockConfig := mock_config.NewMockAPIConfig(ctrl)
+
+	mockConfig.EXPECT().Instance().Return("test").AnyTimes()
+
+	mockAWS.ECS.EXPECT().
+		StopTask(gomock.Any()).
+		Return(nil, awserr.New("TaskDoesNotExist", "Taks 'tsk_id' does not exist", nil))
```
diemonster: [nit] typo "Taks"

api/provider/aws/test_aws/task_list_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+		return nil
+	}
+
+	mockAWS.ECS.EXPECT().
+		ListClustersPages(&ecs.ListClustersInput{}, gomock.Any()).
+		Do(listClusterPagesFN).
+		Return(nil)
+
+	var taskARNs = map[string]map[string][]*string{}
```
zpatrick: I think using a 4-dimensional map is overthinking this. We expect 4 calls and want to return a different arn each time - sounds like this can just be accomplished with a `switch` statement within whatever function you pass to `.Do(...)` (see below)

api/provider/aws/test_aws/task_list_test.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	for _, environmentID := range []string{"l0-test-env_id1", "l0-test-env_id2"} {
+		for _, status := range []string{ecs.DesiredStatusRunning, ecs.DesiredStatusStopped} {
+			listTasksInput := &ecs.ListTasksInput{}
+			listTasksInput.SetCluster(environmentID)
+			listTasksInput.SetDesiredStatus(status)
+			listTasksInput.SetStartedBy("test")
+
+			mockAWS.ECS.EXPECT().
```
zpatrick:
```go
newListTaskPagesFN := func(i int) func(input *ecs.ListTasksInput, fn func(output *ecs.ListTasksOutput, lastPage bool) bool) error {
    return func(input *ecs.ListTasksInput, fn func(output *ecs.ListTasksOutput, lastPage bool) bool) error {
        var arn string
        switch i {
        case 0:
            arn = "../arn1"
        ...
        default:
            arn = ".../arn4"
        }

        output := &ecs.ListTasksOutput{}
        output.SetTaskArns([]*string{arn})
        fn(output, true)
 
        return nil
    })
}

for i := range  environments {
    for j := range statuses {
        mockAWS.ECS.EXPECT().
            ListTasksPages(listTasksInput, gomock.Any()).
            Do(newListTaskPagesFN(i+j)).
            Return(nil)
    }
}
```

api/provider/aws/test_aws/task_read_test.go
```diff
@@ -0,0 +1,108 @@
...
+			EntityID:   "dpl_id",
+			EntityType: "deploy",
+			Key:        "name",
+			Value:      "dpl_name",
+		},
+		{
+			EntityID:   "dpl_id",
+			EntityType: "deploy",
+			Key:        "version",
+			Value:      "version",
```
zpatrick: this should be a different value than the key

api/provider/aws/test_aws/task_read_test.go
```diff
@@ -0,0 +1,108 @@
...
+		Return(describeTaskOutput, nil)
+
+	target := provider.NewTaskProvider(mockAWS.Client(), tagStore, mockConfig)
+	result, err := target.Read("tsk_id")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	container := models.Container{
+		ContainerName: "container",
```
zpatrick: We need to check the `Status` and `ExitCode` fields

api/provider/aws/test_aws/task_list_test.go
zpatrick: you could avoid the `switch` statement since each arn is just a prefix + integer suffix, e.g. 
```
arn := fmt.Sprintf("aws:arn:..../arn%d", i+1)
```

api/provider/aws/common.go
```diff
@@ -82,7 +82,7 @@ func lookupTaskDefinitionARNFromDeployID(store tag.Store, deployID string) (stri
 func lookupEntityEnvironmentID(store tag.Store, entityType, entityID string) (string, error) {
 	tags, err := store.SelectByTypeAndID(entityType, entityID)
 	if err != nil {
-		return "", err
+		return "ts", err
```
jparsons04: This seems curious to me. I didn't see any references to the string `"ts"` for each reference I could find for this function. Is this string being explicitly tested somewhere that I'm not seeing?

api/provider/aws/test_aws/task_list_test.go
```diff
@@ -0,0 +1,154 @@
...
+			Key:        "name",
+			Value:      "env_name1",
+		},
+		{
+			EntityID:   "env_id2",
+			EntityType: "environment",
+			Key:        "name",
+			Value:      "env_name2",
+		},
+	}
```
jparsons04: I'm curious what happens in the case where a layer0 environment not named `test` (e.g. `l0-someotherenvironment-env_id1`) is included in the tags db). Would list still return only the clusters with the environment name `test`?

api/provider/aws/common.go
nehayward: Good catch. Hmm I don't recall adding this. I don't think it's necessary.

api/provider/aws/test_aws/task_list_test.go
nehayward: So if we add something like this. 
```go
	{
		EntityID:   "env_id3",
		EntityType: "environment",
		Key:        "name",
		Value:      "env_name3",
	},
```
It'll still only return the list of tasks in "test" environment (set environment). At least in testing, however this functionality is achieved by cli via the `l0 task list --all`

api/provider/aws/task_delete.go
```diff
@@ -13,12 +14,12 @@ import (
 func (t *TaskProvider) Delete(taskID string) error {
 	environmentID, err := lookupEntityEnvironmentID(t.TagStore, "task", taskID)
 	if err != nil {
-		return err
+		log.Printf("[WARN] Environment not found\n")
```
zpatrick: We need to do type assertions on the error and ensure that the error code is `TaskDoesNotExist`. Only in that scenario do we `return nil` early (also, you forgot to `return nil` after logging) .

api/provider/aws/task_delete.go
```diff
@@ -13,12 +14,12 @@ import (
...
 	}
 
 	taskARN, err := t.lookupTaskARN(taskID)
 	if err != nil {
-		return err
+		log.Printf("[WARN] Task not found\n")
```
zpatrick: ditto

api/provider/aws/test_aws/task_delete_test.go
```diff
@@ -0,0 +1,135 @@
...
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockAWS := awsc.NewMockClient(ctrl)
+	tagStore := tag.NewMemoryStore()
+	mockConfig := mock_config.NewMockAPIConfig(ctrl)
+
+	mockConfig.EXPECT().Instance().Return("test").AnyTimes()
+
+	mockAWS.ECS.EXPECT().
```
zpatrick: we shouldn't get this far if the tags don't exist, correct?

api/provider/aws/test_aws/task_list_test.go
zpatrick: @jparsons04: I think you're conflating environment ids with the layer0 instance name, e.g. `env_id1` vs. `test`

api/provider/aws/test_aws/task_list_test.go
```diff
@@ -0,0 +1,154 @@
...
+	for i, environmentID := range []string{"l0-test-env_id1", "l0-test-env_id2"} {
+		for j, status := range []string{ecs.DesiredStatusRunning, ecs.DesiredStatusStopped} {
+			listTasksInput := &ecs.ListTasksInput{}
+			listTasksInput.SetCluster(environmentID)
+			listTasksInput.SetDesiredStatus(status)
+			listTasksInput.SetStartedBy("test")
+
+			mockAWS.ECS.EXPECT().
+				ListTasksPages(listTasksInput, gomock.Any()).
+				Do(newListTaskPagesFN(i + j)).
```
zpatrick: doesn't this create a `aws:arn.../arn0` on the first iteration? Seems like we should be testing that 4 tasks are returned: 2 for each environment

api/provider/aws/test_aws/task_read_test.go
```diff
@@ -0,0 +1,112 @@
...
+		Status:        "status",
+		ExitCode:      0,
```
zpatrick: We should avoid using 0-values in tests; this should probably be a `1`

api/provider/aws/test_aws/task_list_test.go
nehayward: Yeah I'll fix that. I agree, is that just add 4 task tags and assert 4 exist.

api/provider/aws/task_delete.go
nehayward: Okay and this behavior is already done here right https://github.com/quintilesims/layer0/blob/e6cf881ab637f2f5cd57848acc639fa262251b74/api/provider/aws/task_delete.go#L48

zpatrick: similar behavior, but you should be checking for an `*errors.ServerError` and checking the `.StatusCode`, not the message

api/provider/aws/test_aws/task_list_test.go
```diff
@@ -0,0 +1,205 @@
...
+			output.SetTaskArns(arn)
+			fn(output, true)
+
+			return nil
+		}
+
+		return listTaskPagesFN
+	}
+
+	taskARNCount := 1
```
zpatrick: [nit] could just used `i + j + 1`

api/provider/aws/test_aws/task_list_test.go
jparsons04: Yep, sorry. I think my original comment can be discarded.

api/provider/aws/test_aws/task_list_test.go
nehayward: Not quite, since it won't hit every case, then if you try to add 1 it just complicates things and still doesn't hit every case unless you apply some conditional.
0 + 0 = 0
0 + 1 = 1
1 + 0 = 1
1 + 1 = 2

api/provider/aws/task_delete.go
```diff
@@ -13,11 +15,19 @@ import (
...
+		log.Printf("[WARN] Environment not found\n")
+		if err, ok := err.(*errors.ServerError); ok && err.Code == errors.TaskDoesNotExist {
+			return nil
+		}
 		return err
 	}
 
 	taskARN, err := t.lookupTaskARN(taskID)
 	if err != nil {
+		log.Printf("[WARN] Task not found\n")
```
zpatrick: This needs to be inside the `if err, ok` block

api/provider/aws/test_aws/task_list_test.go
zpatrick: oops, my bad!

api/provider/aws/test_aws/task_read_test.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	describeTaskInput := &ecs.DescribeTasksInput{}
+	describeTaskInput.SetCluster("l0-test-env_id")
+	describeTaskInput.SetTasks([]*string{aws.String("arn:aws:ecs:region:012345678910:task/arn")})
+
+	containerECS := &ecs.Container{}
+	containerECS.SetName("container")
+	containerECS.SetLastStatus("status")
+	containerECS.SetExitCode(0)
```
zpatrick: let's set this to 1 so we aren't using 0-values

api/provider/aws/task_delete.go
```diff
@@ -13,11 +15,19 @@ import (
 func (t *TaskProvider) Delete(taskID string) error {
 	environmentID, err := lookupEntityEnvironmentID(t.TagStore, "task", taskID)
 	if err != nil {
+		if err, ok := err.(*errors.ServerError); ok && err.Code == errors.TaskDoesNotExist {
+			log.Printf("[WARN] Environment not found\n")
```
diemonster: `task` not `environment`

PR #417, opened 2017-11-07
Terraform update

zpatrick:
The main point of this PR is to update the terraform plugin. I've updated all of the code to work with the update client / job model. I've also slightly changed the unit test patterns for this package. While testing this, I encountered quite a few bugs/missing features needed to properly test these changes. With that, I've also fixed the following bugs / added the missing functionality: 

* Adds tags for jobs
* Adds a custom error handler to the api, so all uncaught errors are still returned as a `ServerError`. It also maps our error codes to meaningful http status codes. 
* Fixes cli text printer tests and implementation for tasks
* Adds retry logic to service create, since that would sometimes fail with the "failed to assume role and validate ports" error, due to the loadbalancer still coming up. 
* Adds an optional `scale` param to service create 
* Adds the missing job command to the cli 
* Adds shorthand flag names for `output` and `timeout`, e.g. `l0 -o json ...`

Unit test all of the package where I have made changes. 
To test the terraform provider, I usually just create a `main.tf` with every available resource and data source. I add/update/remove each of the resources' options until i've exhausted all possible configurations. You also need to test that it gracefully handles when things no longer exist - e.g. it was created in terraform but deleted by the CLI. 

**Checklist**
- [x] Unit tests


plugins/terraform/layer0/data_source_layer0_service.go
```diff
@@ -0,0 +1,69 @@
...
+	query.Set(client.TagQueryParamName, d.Get("name").(string))
+	query.Set(client.TagQueryParamEnvironmentID, d.Get("environment_id").(string))
+
+	tags, err := apiClient.ListTags(query)
+	if err != nil {
+		return err
+	}
+
+	if len(tags) == 0 {
+		return fmt.Errorf("No service found matching the specified name and environment id")
```
tlake: [nit] There might be debugging value in providing things like the service name and environment ID in these errors.

api/provider/aws/service_create.go
```diff
@@ -12,8 +12,6 @@ func (s *ServiceProvider) Create(req models.CreateServiceRequest) (string, error
 	fqEnvironmentID := addLayer0Prefix(s.Config.Instance(), req.EnvironmentID)
 	cluster := fqEnvironmentID
 
-	desiredCount := 1
```
diemonster: is this a terraform-specific change or just a fix in general?

common/models/create_service_request.go
```diff
@@ -11,6 +11,7 @@ type CreateServiceRequest struct {
 	EnvironmentID  string `json:"environment_id"`
 	LoadBalancerID string `json:"load_balancer_id"`
 	ServiceName    string `json:"service_name"`
+	Scale          int    `json:"scale"`
```
diemonster: [rambling] I think I see why you did this now... easier to define the scale upfront for use with terraform. I do see some edge cases, however, with systems like Consul where you still need to set scale later on after the first service is officially up.

plugins/terraform/layer0/data_source_layer0_api.go
```diff
@@ -32,19 +36,19 @@ func dataSourceLayer0API() *schema.Resource {
 }
 
 func dataSourceLayer0APIRead(d *schema.ResourceData, meta interface{}) error {
-	client := meta.(*Layer0Client)
+	apiClient := meta.(client.Client)
```
diemonster: [nit] the naming of `apiClient` caught me for a second here --- what api? are there other clients? Knowing the context of the function name makes it more clear, but it's still feels like an assumption that it's related to layer0.

plugins/terraform/layer0/data_source_layer0_service.go
diemonster: agreed

plugins/terraform/layer0/resource_layer0_service.go
```diff
@@ -105,48 +105,51 @@ func resourceLayer0ServiceRead(d *schema.ResourceData, meta interface{}) error {
 }
 
 func resourceLayer0ServiceUpdate(d *schema.ResourceData, meta interface{}) error {
-	client := meta.(*Layer0Client)
-	serviceID := d.Id()
+	apiClient := meta.(client.Client)
+	//serviceID := d.Id()
```
diemonster: [nit] remove?

diemonster:
Why are there so many vendored code changes in this PR? Are they necessary for terraform or were they auto-updated as part of something else?

plugins/terraform/layer0/resource_layer0_load_balancer.go
```diff
@@ -150,30 +155,37 @@ func resourceLayer0LoadBalancerRead(d *schema.ResourceData, meta interface{}) er
...
 	d.Set("url", loadBalancer.URL)
 
 	return nil
 }
 
 func resourceLayer0LoadBalancerUpdate(d *schema.ResourceData, meta interface{}) error {
-	client := meta.(*Layer0Client)
-	loadBalancerID := d.Id()
+	apiClient := meta.(client.Client)
+	//loadBalancerID := d.Id()
```
diemonster: [nit] unneeded?

api/provider/aws/service_create.go
zpatrick: fix in general

common/models/create_service_request.go
zpatrick: yeah this doesn't remove that behavior, the default scale is still 1

zpatrick:
> Why are there so many vendored code changes in this PR? Are they necessary for terraform or were they auto-updated as part of something else?

I believe this was because I ran `go format` on the entire repo

plugins/terraform/layer0/data_source_layer0_api.go
```diff
@@ -32,19 +36,19 @@ func dataSourceLayer0API() *schema.Resource {
...
 
-	apiConfig, err := client.API.GetConfig()
+	config, err := apiClient.ReadConfig()
 	if err != nil {
 		return err
 	}
 
-	d.SetId(apiConfig.Prefix)
+	d.SetId(config.Instance)
+	d.Set("instance", config.Instance)
```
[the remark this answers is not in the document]
zpatrick: So, looking through all of the other examples of modules as https://github.com/terraform-providers, it seems like they never do it in their code (https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_elb.go#L315)

sesh-kebab:
@zpatrick  would you please be able to update develop.md with development workflow information for the terraform provider also?

diemonster:
@sesh-kebab @zpatrick to seshi's point, I think what we need detailed here eventually is "How should this be tested?". This is still WIP though so I don't think it's fully fair to have that detailed yet.

api/controllers/deploy.go
```diff
@@ -51,12 +54,12 @@ func (d *DeployController) CreateDeploy(c *fireball.Context) (fireball.Response,
 		return nil, errors.New(errors.InvalidRequest, err)
 	}
 
-	return createJob(d.JobStore, job.CreateDeployJob, req)
+	return createJob(d.TagStore, d.JobStore, job.CreateDeployJob, req)
```
diemonster: was this a bug in the prior implementation to not include `TagStore`?

cli/command/job.go
```diff
@@ -10,6 +13,76 @@ func NewJobCommand(b *CommandBase) *JobCommand {
 	return &JobCommand{b}
 }
 
-func (e *JobCommand) Command() cli.Command {
-	return cli.Command{}
+func (j *JobCommand) Command() cli.Command {
```
diemonster: We should break this into a separate PR

api/main.go
```diff
@@ -84,16 +84,21 @@ func main() {
...
+		routes = append(routes, controllers.NewDeployController(deployProvider, jobStore, tagStore).Routes()...)
+		routes = append(routes, controllers.NewEnvironmentController(environmentProvider, jobStore, tagStore).Routes()...)
+		routes = append(routes, controllers.NewJobController(jobStore, tagStore).Routes()...)
+		routes = append(routes, controllers.NewLoadBalancerController(loadBalancerProvider, jobStore, tagStore).Routes()...)
+		routes = append(routes, controllers.NewServiceController(serviceProvider, jobStore, tagStore).Routes()...)
 		routes = append(routes, controllers.NewTagController(tagStore).Routes()...)
-		routes = append(routes, controllers.NewTaskController(taskProvider, jobStore).Routes()...)
+		routes = append(routes, controllers.NewTaskController(taskProvider, jobStore, tagStore).Routes()...)
+
+		// todo: add auth decroator
```
diemonster: [mega nit] typo

api/controllers/deploy.go
zpatrick: yes, we weren't creating tags when we would create a job, so we weren't able to resolve job tags.

diemonster: ah right, now I remember this being mentioned previously. 🏅

cli/printer/text.go
```diff
@@ -370,14 +365,14 @@ func (t *TextPrinter) PrintTasks(tasks ...*models.Task) error {
 		return strings.Replace(t.DeployID, ".", ":", 1)
 	}
 
-	rows := []string{"TASK ID | TASK NAME | ENVIRONMENT | DEPLOY | SCALE "}
+	rows := []string{"TASK ID | TASK NAME | ENVIRONMENT | DEPLOY | STATUS "}
```
sesh-kebab: This would require documentation where we show the output of a task to be updated. For example, http://layer0.ims.io/guides/one_off_task/ .

cli/printer/text_test.go
```diff
@@ -356,45 +345,40 @@ func ExampleTextPrintTasks() {
 			TaskName:        "tsk1",
 			EnvironmentID:   "eid1",
 			EnvironmentName: "ename1",
-			RunningCount:    1,
-			DesiredCount:    1,
 			DeployName:      "d1",
 			DeployVersion:   "1",
+			Status:          "RUNNING",
```
sesh-kebab: Is there any other status apart from RUNNING?

common/errors/error.go
```diff
@@ -33,21 +36,26 @@ func (s *ServerError) Model() models.ServerError {
...
-		return Newf(JobDoesNotExist, "Job '%s' does not exist", entityID)
-	case "load_balancer":
-		return Newf(LoadBalancerDoesNotExist, "Load balancer '%s' does not exist", entityID)
-	case "service":
-		return Newf(ServiceDoesNotExist, "Service '%s' does not exist", entityID)
-	case "task":
-		return Newf(TaskDoesNotExist, "Task '%s' does not exist", entityID)
+func (s *ServerError) Write(w http.ResponseWriter, r *http.Request) {
+	switch s.Code {
+	case InvalidRequest,
```
sesh-kebab: [nit] I would expect InvalidRequest and DoesNotExist errors to return different http status codes. 400 & 404 respectively.

common/errors/helpers.go
```diff
@@ -0,0 +1,39 @@
+package errors
+
+import (
+	"fmt"
+)
+
+func MultiError(errors []error) error {
+	if count := len(errors); count == 0 {
```
sesh-kebab: did not know you could do this 👍

plugins/terraform/layer0/data_source_layer0_load_balancer_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+	if err := dataSourceLayer0LoadBalancerRead(d, mockClient); err != nil {
+		t.Fatal(err)
+	}
+
+	assert.Equal(t, "lb_id", d.Id())
+	assert.Equal(t, "lb_name", d.Get("name"))
+	assert.Equal(t, "env_id", d.Get("environment_id"))
+	assert.Equal(t, "env_name", d.Get("environment_name"))
+	assert.Equal(t, true, d.Get("private"))
```
sesh-kebab: [nit] odd that our model used `isPublic` but we expose `private` via terraform.

api/controllers/error_test.go
```diff
@@ -0,0 +1,91 @@
...
+		{
+			Name:           "ServerError: InvalidRequest",
+			Error:          errors.Newf(errors.InvalidRequest, ""),
+			ExpectedStatus: http.StatusBadRequest,
+			ExpectedCode:   errors.InvalidRequest,
+		},
+		{
+			Name:           "ServerError: DeployDoesNotExist",
+			Error:          errors.Newf(errors.DeployDoesNotExist, ""),
+			ExpectedStatus: http.StatusBadRequest,
```
sesh-kebab: [nit] I would expect a `http.StatusNotFound` here.

api/provider/aws/service_create.go
```diff
@@ -53,10 +51,20 @@ func (s *ServiceProvider) Create(req models.CreateServiceRequest) (string, error
 				}
 			}
 		}
+	}
 
+	scale := req.Scale
+	if req.Scale == 0 {
```
sesh-kebab: What if I wanted the scale to be 0 initially?

plugins/terraform/layer0/data_source_layer0_api_test.go
```diff
@@ -0,0 +1,43 @@
...
+		ReadConfig().
+		Return(config, nil)
+
+	apiDataSource := Provider().(*schema.Provider).DataSourcesMap["layer0_api"]
+	d := schema.TestResourceDataRaw(t, apiDataSource.Schema, map[string]interface{}{})
+	if err := dataSourceLayer0APIRead(d, mockClient); err != nil {
+		t.Fatal(err)
+	}
+
+	assert.Equal(t, "test", d.Id())
```
sesh-kebab: I don't understand how these asserts are working. How does `d` get the values from `config`?

cli/printer/text_test.go
```diff
@@ -83,43 +84,31 @@ func ExampleTextPrintJobs() {
...
-	// JOB ID  TASK ID  TYPE                  STATUS       CREATED
-	// id1     t1       Delete Environment    Pending      0001-01-01 00:00:00
-	// id2     t2       Delete Service        In Progress  0001-01-01 00:00:00
-	// id3     t3       Delete Load Balancer  Completed    0001-01-01 00:00:00
-	// id4     t4       Delete Environment    Error        0001-01-01 00:00:00
-
+	// JOB ID   TYPE                STATUS      CREATED
+	// job_id1  DeleteEnvironment   Pending     0001-01-01 00:00:00
+	// job_id2  CreateService       InProgress  0001-01-01 00:00:00
+	// job_id3  UpdateLoadBalancer  Error       0001-01-01 00:00:00
```
jparsons04: I feel like the fourth `Status` `Completed` should still be tested.

plugins/terraform/layer0/resource_layer0_environment_test.go
```diff
@@ -0,0 +1,100 @@
...
+		Return(&models.Job{Status: job.Completed.String()}, nil)
+
+	environmentResource := Provider().(*schema.Provider).ResourcesMap["layer0_environment"]
+	d := schema.TestResourceDataRaw(t, environmentResource.Schema, map[string]interface{}{})
+	d.SetId("env_id")
+
+	if err := resourceLayer0EnvironmentDelete(d, mockClient); err != nil {
+		t.Fatal(err)
+	}
+}
```
jparsons04: I think we need to test the `resourceLayer0EnvironmentUpdate` in this test file.

plugins/terraform/layer0/resource_layer0_environment.go
```diff
@@ -121,21 +143,21 @@ func resourceLayer0EnvironmentUpdate(d *schema.ResourceData, meta interface{}) e
...
 
 		return err
 	}
 
-	if err := waitForJobWithContext(client, jobID); err != nil {
-		return err
-	}
-
 	return nil
 }
```
jparsons04: I know you have a TODO in the ResourceProvider, but want to explicitly call out here that we do need to implement Environment Link, particularly being able to Update Environment links.

api/provider/aws/job_runner.go
```diff
@@ -121,7 +121,16 @@ func (r *JobRunner) createService(jobID, request string) (string, error) {
 		return "", errors.New(errors.InvalidRequest, err)
 	}
 
-	serviceID, err := r.serviceProvider.Create(req)
+	serviceID, err := catchAndRetry(time.Minute*5, func() (result string, err error, shouldRetry bool) {
+		log.Printf("[DEBUG] [JobRunner] Creating service %s", req.ServiceName)
+		serviceID, err := r.serviceProvider.Create(req)
+		if err != nil {
```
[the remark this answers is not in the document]
zpatrick: The "unable to assume role" error: https://github.com/Shippable/support/issues/3390

api/controllers/error_test.go
zpatrick: I went w/ StatusBadRequest because we can hit something like EnvironmentDoesNotExist on service create.

api/provider/aws/service_create.go
zpatrick: makes more sense to me to just now allow that.

cli/printer/text.go
zpatrick: 👍

cli/printer/text_test.go
zpatrick: http://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_life_cycle.html

common/errors/error.go
zpatrick: addressed above

plugins/terraform/layer0/data_source_layer0_api_test.go
zpatrick: requires some knowledge of terraform providers: https://www.terraform.io/guides/writing-custom-terraform-providers.html

plugins/terraform/layer0/data_source_layer0_load_balancer_test.go
zpatrick: This is a UX decision, the same reason we expose `--private` in the CLI; we want the "default" behavior (omission of a flag/field) to be public.

plugins/terraform/layer0/resource_layer0_environment_test.go
zpatrick: I purposely removed update unit tests as our current understanding of the`d` object doesn't lend itself for unit testing very well. If we do this, it should be a separate task for after the refactor IMO.

diemonster:
Is there any use for the `admin` API in the terraform plugin?

api/provider/aws/service_create.go
sesh-kebab: Not sure what you mean? Makes more sense to allow 0? or just now allow a scale of 0?

DEVELOP.md
```diff
@@ -64,24 +64,22 @@ make build
 make release
 ```
 
-## Layer0 Runner
-The Layer0 Runner is a tool that runs asynchronous jobs.
-The [Layer0 API](#layer0-api) creates job for the Layer0 Runner.
+## Layer0 Terraform plugin
```
diemonster: I know we mentioned terraform version dependency as a client-facing requirement, but I still think we should call out what version we're using for our own development as well.

plugins/terraform/layer0/resource_layer0_environment_test.go
```diff
@@ -0,0 +1,100 @@
...
+	assert.Equal(t, "env_id", d.Id())
+	assert.Equal(t, "env_name", d.Get("name").(string))
+	assert.Equal(t, "m3.large", d.Get("size").(string))
+	assert.Equal(t, 1, d.Get("cluster_count").(int))
+	assert.Equal(t, "sgid", d.Get("security_group_id").(string))
+	assert.Equal(t, "linux", d.Get("os").(string))
+	assert.Equal(t, "ami123", d.Get("ami").(string))
+}
+
+func TestResourceEnvironmentDelete(t *testing.T) {
```
diemonster: We should have a rainy day test for Environment Delete

jparsons04:
Can you explain the test pattern where you create a resource in terraform and then delete the resource via the cli. Are you saying that if you were to `terraform plan` after the deletion that terraform would gracefully handle that situation? I didn't see what the specific testing pattern was supposed to be from terraform's perspective for gracefully handling that.

plugins/terraform/layer0/resource_layer0_environment.go
```diff
@@ -1,10 +1,14 @@
...
 
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/quintilesims/layer0/api/provider/aws"
+	"github.com/quintilesims/layer0/client"
+	"github.com/quintilesims/layer0/common/config"
 	"github.com/quintilesims/layer0/common/errors"
+	"github.com/quintilesims/layer0/common/models"
 )
 
 func resourceLayer0Environment() *schema.Resource {
```
jparsons04: The [Exists](https://www.terraform.io/docs/plugins/provider.html#exists) Resource CRUD operation needs to be implemented here.

I created a simple environment with terraform (v0.11)
```
provider "layer0" {
  skip_ssl_verify = true
}

resource "layer0_environment" "demo" {
  name      = "demo"
  min_count = 1
  os        = "linux"
  size      = "t2.micro"
}
```
Then I removed this environment from the CLI. Running `terraform refresh` yields a 500 response from the layer0 API due to the security group of the environment not existing. It seems that this is due to `Exists` not being present within the Resouce's operations.

plugins/terraform/layer0/resource_layer0_deploy.go
```diff
@@ -1,10 +1,13 @@
...
 	"log"
 
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/quintilesims/layer0/client"
+	"github.com/quintilesims/layer0/common/config"
 	"github.com/quintilesims/layer0/common/errors"
+	"github.com/quintilesims/layer0/common/models"
 )
 
 func resourceLayer0Deploy() *schema.Resource {
```
jparsons04: Ditto implementing `Exists` here.

plugins/terraform/layer0/resource_layer0_load_balancer.go
```diff
@@ -8,6 +8,9 @@ import (
 
 	"github.com/hashicorp/terraform/helper/hashcode"
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/quintilesims/layer0/api/provider/aws"
+	"github.com/quintilesims/layer0/client"
+	"github.com/quintilesims/layer0/common/config"
 	"github.com/quintilesims/layer0/common/errors"
 	"github.com/quintilesims/layer0/common/models"
 )
```
jparsons04: Ditto implementing `Exists` in `resourceLayer0LoadBalancer()`.

plugins/terraform/layer0/resource_layer0_service.go
```diff
@@ -1,10 +1,13 @@
...
 	"log"
 
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/quintilesims/layer0/client"
+	"github.com/quintilesims/layer0/common/config"
 	"github.com/quintilesims/layer0/common/errors"
+	"github.com/quintilesims/layer0/common/models"
 )
 
 func resourceLayer0Service() *schema.Resource {
```
jparsons04: Ditto implementing `Exists` here.

plugins/terraform/layer0/resource_layer0_environment_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/golang/mock/gomock"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/quintilesims/layer0/api/job"
+	"github.com/quintilesims/layer0/client/mock_client"
+	"github.com/quintilesims/layer0/common/models"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestResourceEnvironmentCreateRead(t *testing.T) {
```
diemonster: We should add some test cases for all the entities to capture the https://www.terraform.io/docs/plugins/provider.html#exists behavior Josh mentioned elsewhere -- i.e. someone deletes an entity from Swagger or the CLI, and terraform can gracefully handle it when running `apply`, etc.

diemonster: We should add some test cases for all the entities to capture the https://www.terraform.io/docs/plugins/provider.html#exists behavior Josh mentioned elsewhere -- i.e. someone deletes an entity from Swagger or the CLI, and terraform can gracefully handle it when running `apply`, etc.

api/controllers/error_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"net/http/httptest"
+	"testing"
+
+	"github.com/quintilesims/layer0/common/errors"
+	"github.com/quintilesims/layer0/common/models"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestErrorHandler(t *testing.T) {
+	cases := []struct {
```
nehayward: [nit] I think this reads better and is inline with our style. 

`
	type ErrorCases struct {
		Name           string
		Error          error
		ExpectedStatus int
		ExpectedCode   errors.ErrorCode
	}

	cases := []ErrorCases{
`

cli/command/job.go
```diff
@@ -10,6 +13,76 @@ func NewJobCommand(b *CommandBase) *JobCommand {
...
+				Name:      "read",
+				Usage:     "describe a job",
+				Action:    j.read,
+				ArgsUsage: "NAME",
+			},
+			{
+				Name:      "list",
+				Usage:     "list all jobs",
+				Action:    j.list,
+				ArgsUsage: " ",
```
nehayward: [nit] Should we omit this line?

api/provider/aws/service_create.go
diemonster: Just a guess but "now" seems like a typo... should be "not". What's a practical reason to specify a scale of 0 initially? Can't think of any off the top of my head...

plugins/terraform/layer0/data_source_layer0_service.go
```diff
@@ -0,0 +1,69 @@
...
+		return fmt.Errorf("No service found matching %#v", query)
+	}
+
+	serviceID := tags[0].EntityID
+	service, err := apiClient.ReadService(serviceID)
+	if err != nil {
+		return err
+	}
+
+	d.SetId(service.ServiceID)
```
nehayward: [nit] Is it better to compare this too a struct? I looked into it but it seems to complicate it more. Just wanted an opinion.

plugins/terraform/layer0/provider.go
```diff
@@ -0,0 +1,55 @@
...
+		},
+		ConfigureFunc: providerConfigure,
+	}
+}
+
+func providerConfigure(d *schema.ResourceData) (interface{}, error) {
+	apiClient := client.NewAPIClient(client.Config{
+		Endpoint:  d.Get("endpoint").(string),
+		Token:     d.Get("token").(string),
+		VerifySSL: !d.Get("skip_ssl_verify").(bool),
```
nehayward: Could we change this to ssl_verify and just do d.Get, omit the !

plugins/terraform/layer0/resource_layer0_load_balancer.go
```diff
@@ -76,27 +79,27 @@ func resourceLayer0LoadBalancer() *schema.Resource {
 						"target": {
 							Type:     schema.TypeString,
 							Optional: true,
-							Default:  "TCP:80",
+							Default:  aws.DefaultHealthCheck.Target,
```
nehayward: I like these default changes.

plugins/terraform/layer0/resource_layer0_environment.go
zpatrick: We actually don't need that method, as it uses `Read` when `Exist` isn't implemenated (this is what was used in previous iterations). However, you have found a bug in the API: For the Terraform code to determine when an entity does not exist, we explicitly check that the Layer0 API returns a `<Entity>DoesNotExist` error code. This needs to happen when the environment's SG does not exist.

plugins/terraform/layer0/resource_layer0_environment_test.go
zpatrick: don't need that method

plugins/terraform/layer0/resource_layer0_service.go
zpatrick: not needed

plugins/terraform/layer0/resource_layer0_load_balancer.go
zpatrick: not needed

plugins/terraform/layer0/resource_layer0_deploy.go
zpatrick: not needed

api/controllers/error_test.go
zpatrick: anonymous structs are pretty common for table tests in go: https://dave.cheney.net/2013/06/09/writing-table-driven-tests-in-go

cli/command/job.go
zpatrick: The `cli` package we use will throw up some custom help text for positional arguments if this field is an empty string - we just set it as a space here to stop it from doing that behavior.

plugins/terraform/layer0/data_source_layer0_service.go
zpatrick: I'm not sure what you mean - what's being compared here?

plugins/terraform/layer0/provider.go
zpatrick: I don't think that would be a good idea: by default (when env vars/flags are omitted), we want the configuration to be as safe as possible: e.g. no skipping ssl verification. That forces us to have an "inverse" setting, e.g. `SKIP_SSL_VERIFY=1` vs `SSL_VERIFY=1`. This is common practice with security, e.g. `curl` verifies ssl by default, but you can use `curl --insecure ...` to bypass this.

As for the `client.Config.VerifySSL` field, it's common practice to have affirmative boolean variables to try and keep things simple/consistent - e.g. `VerifySSL` vs `DontVerifySSL`

nehayward: Okay that makes sense, thanks for clarifying.

plugins/terraform/layer0/provider.go
```diff
@@ -0,0 +1,55 @@
...
+		Schema: map[string]*schema.Schema{
+			"endpoint": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The endpoint of the Layer0 API",
+			},
+			"token": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The endpoint of the Layer0 API",
```
tlake: [copypasta] token, not endpoint

DEVELOP.md
zpatrick: added

plugins/terraform/layer0/resource_layer0_environment_test.go
zpatrick: we don't have special behavior for rainy day scenarios - there isn't anything to test

PR #418, opened 2017-11-07
Deploy provider/aws unit tests

diemonster:
Adds unit tests for the deploy entity in provider/aws

Manually via `go test`

**Checklist**
- [x] Unit tests


api/provider/aws/test_aws/deploy_create_test.go
```diff
@@ -0,0 +1,128 @@
...
+	container.SetName("test_name")
+	logConfig := &ecs.LogConfiguration{
+		LogDriver: aws.String("awslogs"),
+		Options: map[string]*string{
+			"awslogs-group":         aws.String("test_group"),
+			"awslogs-region":        aws.String("test_region"),
+			"awslogs-stream-prefix": aws.String("test_prefix"),
+		},
+	}
+	container.SetLogConfiguration(logConfig)
```
jparsons04: [nit] New line after closing curly brace

api/provider/aws/test_aws/deploy_create_test.go
```diff
@@ -0,0 +1,128 @@
...
+	containers := []*ecs.ContainerDefinition{}
+	containers = append(containers, container)
```
jparsons04: Couldn't these two lines be condensed as:
`containers := []*ecs.ContainerDefinition{container}` ?

diemonster: yep

api/provider/aws/test_aws/deploy_list_test.go
```diff
@@ -0,0 +1,138 @@
...
+	for _, tag := range tags {
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	listTaskDefinitionFamiliesPagesfn := func(input *ecs.ListTaskDefinitionFamiliesInput,
+		fn func(output *ecs.ListTaskDefinitionFamiliesOutput, lastPage bool) bool) error {
+		taskDefinitionFamilies := []*string{
+			aws.String("l0-test-"),
```
jparsons04: I may be mistaken, but I believe the family is the name of the task itself without its revision numbers. So I think this would need to be `l0-test-dpl` in your case. Unless this test is already passing for you?

api/provider/aws/test_aws/deploy_list_test.go
```diff
@@ -0,0 +1,138 @@
...
+			EntityID:   "dpl_id1",
+			EntityType: "deploy",
+			Key:        "version",
+			Value:      "1",
+		},
+		{
+			EntityID:   "dpl_id1",
+			EntityType: "deploy",
+			Key:        "arn",
+			Value:      "arn:aws:ecs:region:012345678910:task-definition/l0-test-dpl:1",
```
tlake: Compare the ARNs in these tags to the ARNs that you generate below in [`listTaskDefinitionPagesfn()`](https://github.com/quintilesims/layer0/pull/418/files#diff-a3d4f7c16b15c14bcb92929c3a14dba0R97)

`"arn:aws:ecs:region:012345678910:task-definition/l0-test-dpl:1"`
vs
`"arn:aws:ecs:region:012345678910:task-definition/l0-test-dpl1:1"`

api/provider/aws/test_aws/deploy_create_test.go
```diff
@@ -0,0 +1,128 @@
...
+	defer ctrl.Finish()
+
+	mockAWS := awsc.NewMockClient(ctrl)
+	tagStore := tag.NewMemoryStore()
+	mockConfig := mock_config.NewMockAPIConfig(ctrl)
+
+	mockConfig.EXPECT().Instance().Return("test").AnyTimes()
+	defer provider.SetEntityIDGenerator("dpl_id")()
+
+	tags := models.Tags{
```
zpatrick: these tags shouldn't exist at the start of the test - we need to ensure they get created during `deployProvider.Create`

api/provider/aws/test_aws/deploy_create_test.go
```diff
@@ -0,0 +1,128 @@
...
+		},
+	}
+
+	container.SetLogConfiguration(logConfig)
+	containers := []*ecs.ContainerDefinition{container}
+
+	// define request
+	reqDeployFile := &ecs.TaskDefinition{}
+	reqDeployFile.SetContainerDefinitions(containers)
+	deployFile, _ := json.Marshal(reqDeployFile)
```
zpatrick: gotta check the error here

api/provider/aws/test_aws/deploy_create_test.go
```diff
@@ -0,0 +1,128 @@
...
+	for _, tag := range tags {
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	// define container defaults
+	container := &ecs.ContainerDefinition{}
+	container.SetName("test_name")
+	logConfig := &ecs.LogConfiguration{
```
zpatrick: we should probably have 2 containers, so we verify that one w/o log configuration will add the default layer0 log configuration.

api/provider/aws/test_aws/deploy_create_test.go
```diff
@@ -0,0 +1,128 @@
...
+	req := models.CreateDeployRequest{
+		DeployName: "dpl_id",
+		DeployFile: deployFile,
+	}
+
+	registerTaskDefinitionInput := &ecs.RegisterTaskDefinitionInput{}
+	registerTaskDefinitionInput.SetFamily("l0-test-dpl_id")
+	registerTaskDefinitionInput.SetTaskRoleArn("")
+	registerTaskDefinitionInput.SetContainerDefinitions(containers)
+	registerTaskDefinitionInput.SetVolumes(nil)
```
zpatrick: you don't need to set this to nil, they are already nil.

api/provider/aws/test_aws/deploy_create_test.go
```diff
@@ -0,0 +1,128 @@
...
+	registerTaskDefinitionInput.SetPlacementConstraints(nil)
+
+	taskDefinitionOutput := &ecs.TaskDefinition{}
```
zpatrick: You'll need to set the task definition's arn

api/provider/aws/test_aws/deploy_delete_test.go
```diff
@@ -0,0 +1,114 @@
...
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockAWS := awsc.NewMockClient(ctrl)
+	tagStore := tag.NewMemoryStore()
+	mockConfig := mock_config.NewMockAPIConfig(ctrl)
+
+	mockConfig.EXPECT().Instance().Return("test").AnyTimes()
+
+	tags := models.Tags{
```
zpatrick: we don't need to test tags for the idempotence test

api/provider/aws/test_aws/deploy_delete_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	for _, tag := range tags {
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	mockAWS.ECS.EXPECT().
+		DeregisterTaskDefinition(gomock.Any()).
+		Return(nil, errors.Newf(errors.DeployDoesNotExist, "Task Definition does not exist", nil))
```
zpatrick: shouldn't this be an `awserr.Error`?

zpatrick: There is a bug at https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/deploy_delete.go#L41. This is not idempotent: this should return `nil`

api/provider/aws/test_aws/deploy_list_test.go
zpatrick: would be `l0-test-dpl_id1` and `l0-test-dpl_id2`

api/provider/aws/test_aws/deploy_list_test.go
```diff
@@ -0,0 +1,138 @@
...
+		fn(output, true)
+
+		return nil
+	}
+
+	td := &ecs.ListTaskDefinitionsInput{}
+	td.SetFamilyPrefix("l0-test-")
+	td.SetStatus(ecs.TaskDefinitionStatusActive)
+
+	mockAWS.ECS.EXPECT().
```
zpatrick: don't you need to do this for each deploy family returned above?

api/provider/aws/test_aws/deploy_read_test.go
```diff
@@ -0,0 +1,76 @@
...
+	for _, tag := range tags {
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	taskDefinitionInput := &ecs.DescribeTaskDefinitionInput{}
+	taskDefinitionInput.SetTaskDefinition("arn:aws:ecs:region:012345678910:task/arn")
+
+	taskDefinitionOutput := &ecs.DescribeTaskDefinitionOutput{}
```
zpatrick: [nit] this doesn't needs it's own variable, can just be `.Return(&ecs.DescribeTaskDefinitionOutput{}, nil)`

api/provider/aws/test_aws/deploy_read_test.go
```diff
@@ -0,0 +1,76 @@
...
+		Return(taskDefinitionOutput, nil)
+
+	target := provider.NewDeployProvider(mockAWS.Client(), tagStore, mockConfig)
+	result, err := target.Read("dpl_id")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	expected := &models.Deploy{
+		DeployFile: []byte("null"),
```
zpatrick: this is odd - we'd never have a DeployFile look like this. The task definition output returned by the mock needs to be a non-nil value

api/provider/aws/test_aws/deploy_create_test.go
diemonster: At this point, I can't figure out a decent way to make this test capture that case -- in `deploy_create.go`, the first step a deploy goes through is to have the request rendered via unmarshaling into an `ecs.TaskDefinition`. From there, we check if the subfield `LogConfiguration` is nil, and if so, we inject the proper config. 

Using the mocks, on the other hand, forces us to use `SetLogConfiguration` when creating an `ecs.ContainerDefinition` or you get a `no matching expected call: *mock_config.MockAPIConfig.LogGroupName([])` error. Setting `LogConfiguration` to `nil` doesn't make a difference. If you have a good idea on how to get around that, LMK

api/provider/aws/test_aws/deploy_delete_test.go
diemonster: I made a separate test ala @tlake for a `no tags` idempotence test.

api/provider/aws/test_aws/deploy_delete_test.go
diemonster: ✅

api/provider/aws/test_aws/deploy_delete_test.go
diemonster: The way that `read` is currently implemented means that tags must exist or we throw an error: https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/deploy_read.go#L39

api/provider/aws/test_aws/deploy_create_test.go
```diff
@@ -0,0 +1,112 @@
...
+		},
+	}
+
+	cntr1 := &ecs.ContainerDefinition{}
+	cntr1.SetName("cntr_name_1")
+	cntr1.SetLogConfiguration(logConfig)
+
+	cntr2 := &ecs.ContainerDefinition{}
+	cntr2.SetName("cntr_name_2")
+	cntr2.SetLogConfiguration(logConfig)
```
diemonster: Calling this out in advance that I'm not sure of a way to satisfy Gomock and also test whether `LogConfiguration` is injected into a given `ContainerDefinition`. I'm not sure if this test is really the place to do that, though.

api/provider/aws/deploy_delete.go
```diff
@@ -37,24 +38,14 @@ func (d *DeployProvider) deleteDeploy(taskARN string) error {
 	}
 
 	if _, err := d.AWS.ECS.DeregisterTaskDefinition(input); err != nil {
+		log.Printf("[WARN] Deploy not found\n")
```
zpatrick: this should be printed after we check the error below. Also, only use `log` to print to stdout (maybe this is leftover debug code?)

api/provider/aws/deploy_delete_test.go
```diff
@@ -39,7 +38,7 @@ func TestDeploy_deleteDeployTags(t *testing.T) {
 		}
 	}
 
-	if err := deploy.deleteDeployTags("deploy_id"); err != nil {
+	if err := deleteEntityTags(tagStore, "deploy", "deploy_id"); err != nil {
```
zpatrick: I don't think we need this test anymore since it's just testing `deleteEntityTags`. I assume we already have a `TestDeleteEntityTags`?

api/provider/aws/test_aws/deploy_create_test.go
zpatrick: we need at one container without the `LogConfiguration` set, so we can assert that we add the default layer0 log config correctly.

api/provider/aws/test_aws/deploy_create_test.go
```diff
@@ -0,0 +1,112 @@
...
+			EntityID:   "dpl_id",
+			EntityType: "deploy",
+			Key:        "name",
+			Value:      "dpl_name",
+		},
+		{
+			EntityID:   "dpl_id",
+			EntityType: "deploy",
+			Key:        "version",
+			Value:      "0",
```
zpatrick: Why is the version 0? From the arn you return, `...:task-definition/l0-test-dpl_id:1`, shouldn't the version be 1?

api/provider/aws/test_aws/deploy_delete_test.go
zpatrick: oh I see - then do we have an idempotence test for when the tags don't exist?

api/provider/aws/test_aws/deploy_list_test.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+	}
+
+	taskDefinitionFamilies := []*string{
+		aws.String("l0-test-dpl_id1"),
+		aws.String("l0-test-dpl_id2"),
+	}
+
+	// ListTaskDefinitionsFamiliesPages Mocks
+	listTaskDefinitionFamiliesPagesfn := func(input *ecs.ListTaskDefinitionFamiliesInput,
```
zpatrick: for multiline function signature, we give each param it's own line. The last param is especially important because it gives us a different indentation level to show where the signature ends and the body starts:
```
listTaskDefinitionFamiliesPagesfn := func(
    input *ecs.ListTaskDefinitionFamiliesInput,
    fn func(output *ecs.ListTaskDefinitionFamiliesOutput, lastPage bool) bool,
) error {
    output := &ecs.ListTaskDefinitionFamiliesOutput{}
    ...
```

api/provider/aws/test_aws/deploy_list_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+	// ListTaskDefinitionsPages Mocks
+	taskDefinitionARNs := []*string{
+		aws.String("arn:aws:ecs:region:012345678910:task-definition/l0-test-dpl_id1:1"),
+		aws.String("arn:aws:ecs:region:012345678910:task-definition/l0-test-dpl_id2:1"),
+		aws.String("arn:aws:ecs:region:012345678910:task-definition/l0-bad-dpl_id1:1"),
+	}
+
+	generateTaskDefinitionPagesFN := func(taskDefinitionARN *string) func(input *ecs.ListTaskDefinitionsInput, fn func(output *ecs.ListTaskDefinitionsOutput, lastPage bool) bool) error {
+		listTaskDefinitionPagesFN := func(input *ecs.ListTaskDefinitionsInput, fn func(output *ecs.ListTaskDefinitionsOutput, lastPage bool) bool) error {
```
zpatrick: [nit] we can just `return func(input ...)` instead of declaring it as a variable and returning it below.

api/provider/aws/test_aws/deploy_list_test.go
```diff
@@ -0,0 +1,150 @@
...
+			output := &ecs.ListTaskDefinitionsOutput{}
+			output.SetTaskDefinitionArns([]*string{taskDefinitionARN})
+			fmt.Printf("%#v\n", *output)
```
zpatrick: rm debug statement

api/provider/aws/test_aws/deploy_create_test.go
zpatrick: just create a `logConfig` like you already have, but the new one has the layer0 defaults instead of the custom log group/prefix/etc. Note that you need to separate the objects you use to create the `req.DeployFile` and the one you pass to `.EXPECT().RegisterTaskDefinition(...)`

diemonster: Already called this out in a comment above but will repaste:

I'm not sure of a way to satisfy Gomock and also test whether `LogConfiguration` is injected into a given `ContainerDefinition`. Testing the `LogConfiguration` logic also seems out-of-scope for the tests in this package.

api/provider/aws/deploy_delete.go
diemonster: This was actually a debug leftover, so will remove --- is it necessary to print this to logs given the error will already be present?

api/provider/aws/test_aws/deploy_list_test.go
diemonster: nice! that helps readability quite a bit...

api/provider/aws/test_aws/deploy_delete_test.go
diemonster: ✅

api/provider/aws/deploy_delete_test.go
diemonster: I'm fine w removing this but I'm not sure what else to target in this test otherwise

zpatrick: the whole test hsould be removed

zpatrick: the whole test should be removed

api/provider/aws/test_aws/deploy_create_test.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	req := models.CreateDeployRequest{
+		DeployName: "dpl_name",
+		DeployFile: deployFile,
+	}
+
+	cntr2.SetLogConfiguration(logConfig2)
```
zpatrick: [nit] the complexity of this part probably warrants a comment

PR #420, opened 2017-11-13
Janitor task

nehayward:
This adds task and job janitor with unit tests.

Using the unit tests.

**Checklist**
- [x] Unit tests

closes #


api/tag/janitor.go
```diff
@@ -8,6 +8,26 @@ import (
...
+			return err
+		}
+
+		m := make(map[string]bool, len(tasks))
+		for _, task := range tasks {
+			m[task.TaskID] = true
+		}
+
+		for _, tag := range tags {
+			if m[tag.EntityID] {
```
sesh-kebab: Shouldn't this be
```
if !m[tag.EntityID] { //if the tag doesn't have a corresponding task
```

api/tag/janitor_test.go
```diff
@@ -15,11 +17,109 @@ func TestJanitor(t *testing.T) {
 	taskProvider := mock_provider.NewMockTaskProvider(ctrl)
 
 	// todo: put some old tags in the store, mock the ListTasks call
```
sesh-kebab: Same thing with this todo comment.

common/config/api.go
```diff
@@ -100,6 +105,7 @@ type APIConfig interface {
 	DynamoJobTable() string
 	DynamoTagTable() string
 	LogGroupName() string
+	JobExpiry() int
```
sesh-kebab: Just a thought - but because this is an int, it is unclear if it is minutes, seconds or hours.

nehayward: I'm not sure I can change it, since it can be changed right now via a flag. I'll look into it.

nehayward: I could change it to Duration.

sesh-kebab: I just meant JobExpiry variable name doesn't have any indication of of what time unit. Maybe JobExpiryInSeconds or a comment saying seconds. I'll update this to a `nit` so it doesn't require any changing. I don't know what change I would make to this myself.

api/job/janitor.go
```diff
@@ -8,7 +8,17 @@ import (
 
 func NewJanitor(jobStore Store, expiry time.Duration) *janitor.Janitor {
 	return janitor.NewJanitor("Job", func() error {
-		// todo: select all jobs, delete those older than expiry
+		jobs, err := jobStore.SelectAll()
```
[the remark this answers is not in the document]
nehayward: This would be cool, but I don't see a SelectWhere in the jobstore

common/config/api.go
nehayward: I made some revisions, and made it a durationflag

api/job/janitor.go
```diff
@@ -8,7 +8,17 @@ import (
...
+		if err != nil {
+			return err
+		}
+
+		for _, job := range jobs {
+			if job.Created.Before(time.Now().Add(-expiry)) {
+				jobStore.Delete(job.JobID)
```
zpatrick: you are swallowing the error returned by `jobStore.Delete`

api/job/janitor.go
```diff
@@ -8,7 +8,17 @@ import (
...
 func NewJanitor(jobStore Store, expiry time.Duration) *janitor.Janitor {
 	return janitor.NewJanitor("Job", func() error {
-		// todo: select all jobs, delete those older than expiry
+		jobs, err := jobStore.SelectAll()
+		if err != nil {
+			return err
+		}
+
+		for _, job := range jobs {
+			if job.Created.Before(time.Now().Add(-expiry)) {
```
zpatrick: `if time.Since(job.Created) < expiry`

api/job/janitor_test.go
```diff
@@ -3,17 +3,43 @@ package job
...
 	"time"
+
+	"github.com/quintilesims/layer0/common/models"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestJanitor(t *testing.T) {
 	store := NewMemoryStore()
 	janitor := NewJanitor(store, time.Hour*24)
+	timeNow := time.Now()
```
zpatrick: [nit] we usually just call this variable `now` in other places we use `time.Now()`

api/job/janitor_test.go
```diff
@@ -3,17 +3,43 @@ package job
...
+
+	jobs := []*models.Job{
+		{
+			JobID:   "delete",
+			Created: timeNow.Add(-24 * time.Hour),
```
zpatrick: `time.Sub(24*time.Hour)`

api/job/janitor_test.go
```diff
@@ -3,17 +3,43 @@ package job
...
 	"testing"
 	"time"
+
+	"github.com/quintilesims/layer0/common/models"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestJanitor(t *testing.T) {
 	store := NewMemoryStore()
 	janitor := NewJanitor(store, time.Hour*24)
```
zpatrick: you should probably make `expiry` a variable, and set the `Created` field below based off of those times, e.g. 
```
// delete
Created: now.Sub(expiry+1),
...
// keep
Created: now.Sub(expiry-1),
```

api/job/janitor_test.go
```diff
@@ -3,17 +3,43 @@ package job
...
+			Created: timeNow.Add(-24 * time.Hour),
+		},
+		{
+			JobID:   "keep",
+			Created: timeNow.Add(-12 * time.Hour),
+		},
+	}
 
-	// todo: put some jobs in the store
+	for _, job := range jobs {
```
zpatrick: You should use a `mock_job.MockJobStore` here instead of the memory store. That way you can define the expected behavior more clearly:
```
mockJobStore.EXPECT().
    SelectAll().
    Return(jobs, nil)

mockJobStore.EXPECT().
    DeleteJob("delete").
    Return(nil)
```

and you won't need to do the select + assert behavior below

api/main.go
```diff
@@ -102,8 +102,8 @@ func main() {
 		scalerTicker := scalerDispatcher.RunEvery(time.Minute * 5)
 		defer scalerTicker.Stop()
 
-		// todo: get job expiry from config env var
-		jobJanitor := job.NewJanitor(jobStore, time.Hour*24)
+		expiry := cfg.JobExpiryHour()
```
zpatrick: We should always deal with `time.Duration` types when dealing with time, never integers. Integers add an unnecessary level of indirection. This should be `cfg.JobExpiry()`, and it should return a `time.Duration`.

api/tag/janitor.go
```diff
@@ -7,7 +7,26 @@ import (
...
+		}
+
+		m := make(map[string]bool, len(tasks))
+		for _, task := range tasks {
+			m[task.TaskID] = true
+		}
+
+		for _, tag := range tags {
+			if !m[tag.EntityID] {
+				tagStore.Delete(tag.EntityType, tag.EntityID, tag.Key)
```
zpatrick: you are swallowing the error again

api/tag/janitor.go
```diff
@@ -7,7 +7,26 @@ import (
...
+			}
+		}
 		return nil
```
zpatrick: this line should be deleted

api/job/janitor.go
```diff
@@ -8,7 +8,17 @@ import (
...
+				jobStore.Delete(job.JobID)
+			}
+		}
+
 		return nil
```
zpatrick: this line should be deleted

api/tag/janitor_test.go
```diff
@@ -14,12 +16,109 @@ func TestJanitor(t *testing.T) {
...
+			EnvironmentName: "env_name2",
+		},
+	}
+
+	taskProvider.
+		EXPECT().
+		List().
+		Return(taskSummaries, nil)
+
+	actual, _ := taskProvider.List()
```
zpatrick: never swallow errors

api/tag/janitor_test.go
```diff
@@ -14,12 +16,109 @@ func TestJanitor(t *testing.T) {
...
+			Key:        "environment_id",
+			Value:      "env_id2",
+		},
+		{
+			EntityID:   "tsk_id2",
+			EntityType: "task",
+			Key:        "arn",
+			Value:      "arn:aws:ecs:region:012345678910:task/arn2",
+		},
+		{
```
zpatrick: why do we care about environment tags here?

api/tag/janitor_test.go
```diff
@@ -14,12 +16,109 @@ func TestJanitor(t *testing.T) {
...
+	for _, tag := range tags {
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	taskSummaries := []models.TaskSummary{
+		{
+			TaskID:          "tsk_id1",
+			TaskName:        "tsk_name1",
```
zpatrick: we only care about the `TaskID` field, this other stuff should be removed

api/tag/janitor_test.go
```diff
@@ -14,12 +16,109 @@ func TestJanitor(t *testing.T) {
...
+		{
+			TaskID:          "tsk_id2",
+			TaskName:        "tsk_name2",
+			EnvironmentID:   "env_id2",
+			EnvironmentName: "env_name2",
+		},
+	}
+
+	taskProvider.
+		EXPECT().
```
zpatrick: `EXPECT()` is always on the same line as the mock object, e.g. 
```
mockObject.EXPECT().
    Function(args...).
    Return(results...)
```

api/tag/janitor_test.go
zpatrick: why are we doing this call here?

api/tag/janitor_test.go
```diff
@@ -14,12 +16,109 @@ func TestJanitor(t *testing.T) {
...
+		List().
+		Return(taskSummaries, nil)
+
+	actual, _ := taskProvider.List()
+
+	taskProvider.
```
zpatrick: why are you we expecting this call twice?

api/tag/janitor_test.go
```diff
@@ -14,12 +16,109 @@ func TestJanitor(t *testing.T) {
...
+		List().
+		Return(taskSummaries, nil)
 
 	janitor := NewJanitor(tagStore, taskProvider)
 	if err := janitor.Run(); err != nil {
 		t.Fatal(err)
 	}
 
-	// todo: assert only old tags got deleted
+	expected := []models.TaskSummary{
```
zpatrick: This doesn't make any sense - you're only asserting that the mock call on line `96` returned the objects you told it to return on line `91`. There aren't any expected calls anywhere that would verify that delete was called to any leftover tags.

common/config/api.go
```diff
@@ -14,6 +15,11 @@ func APIFlags() []cli.Flag {
 			Value:  9090,
 			EnvVar: ENVVAR_PORT,
 		},
+		cli.DurationFlag{
+			Name:   FLAG_JOB_EXPIRY,
+			Value:  time.Hour * 24,
```
zpatrick: This should have a default variable in `common.go` in this directory.

common/config/api.go
```diff
@@ -202,3 +209,7 @@ func (c *ContextAPIConfig) PrivateSubnets() []string {
 func (c *ContextAPIConfig) LogGroupName() string {
 	return c.C.String(FLAG_AWS_LOG_GROUP_NAME)
 }
+
+func (c *ContextAPIConfig) JobExpiryHour() time.Duration {
```
zpatrick: `JobExpiry()`

common/config/api.go
nehayward: Like this in  common/config/common.go

const (
	ENVVAR_JOB_EXPIRY           = time.Hour * 24

api/job/janitor_test.go
nehayward: I get an import cycle error when I try to use `mock_job.MockJobStore`

api/tag/janitor_test.go
nehayward: This is one of the test I was struggling with as you can see.

api/job/janitor.go
zpatrick: you also need to delete the job's tags here. You'll need to add that as a DI

nehayward: Isn't that handled in the tag janitor? Or should it be.

api/job/janitor_test.go
nehayward: Sub uses time not duration. https://golang.org/pkg/time/#Time.Sub

api/job/janitor.go
nehayward: I think it needs to be >

api/job/janitor.go
zpatrick: nope

api/job/janitor_test.go
zpatrick: huh weird that `Add` and `Sub` wouldn't be the opposite of eachother 🤷‍♂️

api/job/janitor_test.go
```diff
@@ -3,17 +3,48 @@ package job
...
+		store.jobs = append(store.jobs, job)
+	}
 
 	if err := janitor.Run(); err != nil {
 		t.Fatal(err)
 	}
 
-	// todo: assert only old jobs got deleted
+	expected := []*models.Job{
```
zpatrick: using the mock store would allow this part to be removed

api/tag/janitor.go
```diff
@@ -7,7 +7,28 @@ import (
...
+		}
+
+		for _, tag := range tags {
+			if !m[tag.EntityID] {
+				if err := tagStore.Delete(tag.EntityType, tag.EntityID, tag.Key); err != nil {
+					return err
+				}
+			}
+		}
 		return nil
```
zpatrick: newline after closing bracket

api/tag/janitor_test.go
```diff
@@ -14,12 +16,55 @@ func TestJanitor(t *testing.T) {
...
+		List().
+		Return(tasks, nil)
 
 	janitor := NewJanitor(tagStore, taskProvider)
 	if err := janitor.Run(); err != nil {
 		t.Fatal(err)
 	}
 
-	// todo: assert only old tags got deleted
+	actual, err := tagStore.SelectByType("task")
```
zpatrick: using a mock tag store would allow this part to be removed

zpatrick: not sure why you're strafing away from using the existing `assert.Equal(t, expected, tagStore.tags)` pattern?

common/config/api.go
```diff
@@ -14,6 +15,11 @@ func APIFlags() []cli.Flag {
 			Value:  9090,
 			EnvVar: ENVVAR_PORT,
 		},
+		cli.DurationFlag{
+			Name:  FLAG_JOB_EXPIRY,
+			Value: ENVVAR_JOB_EXPIRY,
+			// EnvVar: ENVVAR_JOB_EXPIRY,
```
zpatrick: This needs to be added

common/config/api.go
```diff
@@ -14,6 +15,11 @@ func APIFlags() []cli.Flag {
 			Value:  9090,
 			EnvVar: ENVVAR_PORT,
 		},
+		cli.DurationFlag{
+			Name:  FLAG_JOB_EXPIRY,
+			Value: ENVVAR_JOB_EXPIRY,
```
zpatrick: This is the default, not the envvar

common/config/common.go
```diff
@@ -28,6 +31,7 @@ const (
 )
 
 const (
+	ENVVAR_JOB_EXPIRY           = time.Hour * 24
```
zpatrick: this is not an envvar, this is a default

api/job/janitor_test.go
nehayward: I get an import cycle error when I try to use mock_job.MockJobStore. I wasn't sure how to implement it.

api/tag/janitor_test.go
nehayward: Yup, you're right that is cleaner.

common/config/api.go
nehayward: So it should be like time.Hour * 24 or what should value you be. That's what I had it before.

common/config/api.go
zpatrick: Yup, the example you have looks good. 

Good question. It doesn't make sense for **all** defaults to exist in the `config` package - for example, the default aws iam policies obviously should live in the `provider/aws` package. However, I agree that it would make more sense for things like the default port to live here. To be honest, I figured it wasn't worth putting up the fight when I was reviewing that PR. Welcome to you making that change though.

api/job/janitor_test.go
```diff
@@ -3,17 +3,48 @@ package job
...
+	"github.com/quintilesims/layer0/common/models"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestJanitor(t *testing.T) {
 	store := NewMemoryStore()
 	janitor := NewJanitor(store, time.Hour*24)
 
-	// todo: put some jobs in the store
+	expiry := -24 * time.Hour
```
zpatrick: This doesn't make sense because we wouldn't have a negative expiry. It also makes the logic below, `now.Add(expiry * -2)`, really confusing. IMO, 'keep' should just `expiry - 1` and 'delete' should be `expiry + 1`; that make it very clear we keep things that are just before expiring, and delete things just after expiring.

api/job/janitor_test.go
zpatrick: ok, then this should just be `assert.Equal(t, expected, tagStore.Tags())`

common/config/api.go
zpatrick: no, it should be `Value: DEFAULT_JOB_EXPIRY`, and there should exist 
```
const (
    DEFAULT_JOB_EXPIRY = time.Hour * 24
```

api/job/janitor_test.go
nehayward: I agree the logic doesn’t make sense. I tried to debug the issue, because when I do addition it doesn’t change the value. I tried wrapping it in ‘( )’ but it didn’t change anything. Have you seen this?

common/models/job.go
```diff
@@ -30,3 +30,40 @@ func (j Job) Definition() swagger.Definition {
...
+)
+
+type JobStatus string
+
+func (j JobStatus) String() string {
+	return string(j)
+}
+
+const (
+	Pending    JobStatus = "Pending"
```
zpatrick: Sense this is no longer in the `job` package, these variables need to be qualified, e.g. `PendingJobStatus`

nehayward: Okay. I'll change that. Is this only syntax or does it do anything else?

nehayward: Should the JobType be qualified too?

zpatrick: I think it already is - `CreateEnvironmentJob`, right?

common/config/common.go
```diff
@@ -53,3 +57,8 @@ const (
 	ENVVAR_AWS_DYNAMO_TAG_TABLE = "LAYER0_AWS_DYNAMO_TAG_TABLE"
 	ENVVAR_AWS_LOG_GROUP_NAME   = "LAYER0_AWS_LOG_GROUP_NAME"
 )
+
+const (
+	DEFAULT_JOB_EXPIRY = time.Hour * 24
+	DEFAULT_PORT       = 9090
```
zpatrick: This doesn't appear to be used anywhere?

common/models/job.go
nehayward: Any other changes?

zpatrick:
@nehayward: you'll need to update the codebase to get all unit tests passing again

api/job/dispatcher_test.go
```diff
@@ -5,33 +5,34 @@ import (
 	"testing"
 
 	"github.com/golang/mock/gomock"
+	"github.com/quintilesims/layer0/common/models"
 	"github.com/stretchr/testify/assert"
 )
 
-func TestDispatcherQueuesPendingJobs(t *testing.T) {
+func TestDispatcherQueuesPendingJobStatusJobs(t *testing.T) {
```
zpatrick: need to undo change here

api/job/janitor_test.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+func TestJanitor(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockJobStore := mock_job.NewMockStore(ctrl)
+	janitor := NewJanitor(mockJobStore, time.Hour*24)
+
+	expiry := -24 * time.Hour
```
zpatrick: this needs to be a positive value

common/config/api.go
```diff
@@ -12,9 +12,14 @@ func APIFlags() []cli.Flag {
 		cli.IntFlag{
 			// todo: renamed from 'LAYER0_API_PORT'
 			Name:   FLAG_PORT,
-			Value:  9090,
+			Value:  DEFAULT_PORT,
 			EnvVar: ENVVAR_PORT,
 		},
+		cli.DurationFlag{
```
zpatrick: this is a duplicate of the one below

common/config/api.go
```diff
@@ -24,6 +29,12 @@ func APIFlags() []cli.Flag {
 			Name:   FLAG_INSTANCE,
 			EnvVar: ENVVAR_INSTANCE,
 		},
+		cli.DurationFlag{
+			Name:   FLAG_JOB_EXPIRY,
+			Value:  24 * time.Hour,
+			EnvVar: ENVVAR_JOB_EXPIRY,
+			Usage:  "duration [ns,us (or µs),ms,s,m,h]",
```
zpatrick: [nit] we don't need a usage flags for the api - it isn't customer facing

common/config/api.go
```diff
@@ -24,6 +29,12 @@ func APIFlags() []cli.Flag {
 			Name:   FLAG_INSTANCE,
 			EnvVar: ENVVAR_INSTANCE,
 		},
+		cli.DurationFlag{
+			Name:   FLAG_JOB_EXPIRY,
+			Value:  24 * time.Hour,
```
zpatrick: this should be lower, maybe 1, 4, or 8 hours, and should use the `DEFAULT_JOB_EXPIRY` variable

api/job/janitor_test.go
```diff
@@ -14,19 +14,20 @@ func TestJanitor(t *testing.T) {
 	defer ctrl.Finish()
 
 	mockJobStore := mock_job.NewMockStore(ctrl)
-	janitor := NewJanitor(mockJobStore, time.Hour*24)
-
-	expiry := -24 * time.Hour
+	janitor := NewJanitor(mockJobStore, time.Hour*12)
```
zpatrick: [nit]
```
expiry := time.Hour * 12
janitor := NewJanitor(mockJobStore, expiry)
```

PR #427, opened 2017-11-16
[l0-setup] update linux AMI

diemonster:
Updates the default Linux AMI used for Layer0.

Cannot be properly tested until l0-setup has been refactored. Merging is a "risk" as-is.

### DIFF
--- a/setup/module/api/variables.tf
+++ b/setup/module/api/variables.tf
@@ -21,13 +21,13 @@ variable "tags" {
   default     = {}
 }
 
-# Current AMI: 	amzn-ami-2016.09.g-amazon-ecs-optimized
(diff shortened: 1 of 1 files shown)


zpatrick:
we should probably switch to ami data sources https://www.terraform.io/docs/providers/aws/d/ami.html

jparsons04:
In my testing of the terraform plugins, I have come to the realization that the Windows AMI also should be updated. For instance, the Windows AMI ID specified for us-west-2 currently doesn't exist. So if you were to create a Windows-based ECS cluster, it would fail.

diemonster:
@zpatrick Cool, TIL. Wondering if using some generic filtering to get most recent AMIs (via `most_recent`) would be a good idea or if that's too risky? Possibly worth exposing as a l0-setup param?

PR #428, opened 2017-11-21
423: Implement retry and delay logic against AWS API

### LINKED ISSUE #423: Configure aws-sdk-go error retry logic
The aws-sdk-go has exponential backoff and other retry logic params that are configurable for a given API client. This issue should track what options are relevant for Layer0 and the work done to implement those configurations.

http://docs.aws.amazon.com/general/latest/gr/api-retries.html

jparsons04:
In reference to issue #423. This adds retry and delay configuration options to help with rate limiting issues that have previously been encountered.

~This introduces three new API flags: a min retry, a max retry and a time between requests flag.~

~We are using a [Config.Retryer](https://github.com/quintilesims/layer0/blob/develop/vendor/github.com/aws/aws-sdk-go/aws/config.go#L108) which is a `RequestRetryer` (an interface) which is in turn an alias for a type that implements the [request.Retryer interface](https://github.com/quintilesims/layer0/blob/develop/vendor/github.com/aws/aws-sdk-go/aws/request/retryer.go#L13).~

~This interface has [three methods](https://github.com/aws/aws-sdk-go/blob/master/aws/client/default_retryer.go#L37) and I've opted to implement two of these methods at this time.~

~There might be a way to test this by unit testing the custom methods of the Retryer. Other suggestions for testing this would be welcome.~

This is just adding some custom configuration flags and environment variables now, so this could be tested by setting these configuration values by hand and observing results.

**Checklist**
~- [ ] Unit tests~

closes #423


jparsons04:
One thing I could certainly do in the interim to make this code cleaner would be to properly implement the Retryer interface rather than short circuit it as I've done here so far.

api/main.go
```diff
@@ -55,13 +56,22 @@ func main() {
 		log.SetOutput(logger)
 
 		awsConfig := defaults.Get().Config
+		customRetryer := config.NewRetryer(cfg.MinRetryTime(), cfg.MaxRetryTime())
+		awsConfig.Retryer = customRetryer
```
zpatrick: From our discussion last week: We actually don’t need a custom retryer. The default exp backoff logic is sufficient; we only need to set the MaxRetries field to something like 999 (so it doesn’t use the “service default” - see sdk docs for more details there).

diemonster: Sorry, I missed this conversation. What's wrong with exposing some of the config for the retryer logic? Part of the benefit of doing so is that we could easily tweak defaults while testing.

common/config/retryer.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	retryCount := r.RetryCount
+	if retryCount > 13 {
+		retryCount = 13
+	} else if throttle && retryCount > 8 {
+		retryCount = 8
+	}
+
+	delay := (1 << uint(retryCount)) * (seededRand.Intn(minTime) + minTime)
```
diemonster: one reason I can see for not wanting this code implemented on our side is testing it :trollface:

api/main.go
zpatrick: The backoff retry logic was never an issue, the rate of api calls was. 

In a broader sense, we found the "correct" solution in the 2nd iteration of this problem, but forgot to cleanup the mess left by the first iteration.

common/config/retryer.go
zpatrick: As stated above, this doesn't fix the root of problem we are trying to solve; it should be removed regardless.

jparsons04:
@zpatrick @diemonster I removed the custom retryer stuff as requested and also made a envvar/flag for customizing the max number of retries.

jparsons04:
I removed the requirement for unit tests as we are not implementing a custom retryer. This now just essentially adding a couple of flags/envvars to customize and enforce the time between AWS requests and the max number of retries.

common/config/api.go
```diff
@@ -80,6 +81,16 @@ func APIFlags() []cli.Flag {
 			Name:   FLAG_AWS_LOG_GROUP_NAME,
 			EnvVar: ENVVAR_AWS_LOG_GROUP_NAME,
 		},
+		cli.DurationFlag{
+			Name:   FLAG_AWS_TIME_BETWEEN_REQUESTS,
+			Value:  10 * time.Millisecond,
+			EnvVar: ENVVAR_AWS_TIME_BETWEEN_REQUESTS,
```
diemonster: Can we be more explicit with the time increment we're using here?

diemonster:
@jparsons04 have you done any of the testing by hand that you mentioned in the PR details?

common/config/api.go
jparsons04: You can where you define your environment variables. Because of urfave/cli's [ParseDuration](https://github.com/golang/go/blob/go1.9/src/time/format.go#L1262), you can have a line in your configuration that reads `export LAYER0_AWS_TIME_BETWEEN_REQUESTS="2s"` and this will correctly be interpreted as 2 seconds. I think we want to set a default here because I think we should give ourselves the ability to rate limit across the board if we run into any serious rate limiting issues that accounts for cases where users are not setting their environment variable.

jparsons04:
Yes, I have.

As one example, I set up an environment using terraform and set the environment variable before destroying it, then set to destroy it. I inserted some debug statements when the PushBack method was being invoked:

```
jparsons@seawl000002 ~/dev/go/src/github.com/quintilesims/layer0/api $ echo $LAYER0_AWS_TIME_BETWEEN_REQUESTS
1500ms
```
Then running terraform destroy after starting the API:
```
jparsons@seawl000002 ~/dev/go/src/github.com/quintilesims/layer0/api $ ./l0-api --debug
2017/11/27 12:19:12 [INFO] Listening on port 9090
2017/11/27 12:19:12 [DEBUG] [JobWorker 2]: Start signalled
2017/11/27 12:19:12 [DEBUG] [JobWorker 1]: Start signalled
aws request made, time: 2017-11-27 12:19:17.78285534 -0800 PST m=+5.204497041
aws request made, time: 2017-11-27 12:19:17.801856386 -0800 PST m=+5.223498099
aws request made, time: 2017-11-27 12:19:18.028927258 -0800 PST m=+5.450568996
aws request made, time: 2017-11-27 12:19:18.822308377 -0800 PST m=+6.243950103
aws request made, time: 2017-11-27 12:19:20.09792704 -0800 PST m=+7.519568753
aws request made, time: 2017-11-27 12:19:21.622500288 -0800 PST m=+9.044142009
aws request made, time: 2017-11-27 12:19:24.597819202 -0800 PST m=+12.019460930
```

common/config/api.go
jparsons04: I implemented something in Usage. Here is what it looks like now:

```
$ ./l0-api --help
NAME:
   Layer0 API - A new cli application

USAGE:
   l0-api [global options] command [command options] [arguments...]

VERSION:
   v0.10.0-270-gf0458e1

COMMANDS:
     help, h  Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --port value                       (default: 9090) [$LAYER0_PORT]
   --debug                             [$LAYER0_DEBUG]
   --instance value                    [$LAYER0_PREFIX]
   --account-id value                  [$LAYER0_AWS_ACCOUNT_ID]
   --access-key value                  [$LAYER0_AWS_ACCESS_KEY_ID]
   --secret-key value                  [$LAYER0_AWS_SECRET_ACCESS_KEY]
   --region value                     (default: "us-west-2") [$LAYER0_AWS_REGION]
   --vpc value                         [$LAYER0_AWS_VPC_ID]
   --linux-ami value                   [$LAYER0_AWS_LINUX_SERVICE_AMI]
   --windows-ami value                 [$LAYER0_AWS_WINDOWS_SERVICE_AMI]
   --s3-bucket value                   [$LAYER0_AWS_S3_BUCKET]
   --instance-profile value            [$LAYER0_AWS_ECS_INSTANCE_PROFILE]
   --job-table value                   [$LAYER0_AWS_DYNAMO_JOB_TABLE]
   --tag-table value                   [$LAYER0_AWS_DYNAMO_TAG_TABLE]
   --public-subnets value              [$LAYER0_AWS_PUBLIC_SUBNETS]
   --private-subnets value             [$LAYER0_AWS_PRIVATE_SUBNETS]
   --log-group-name value              [$LAYER0_AWS_LOG_GROUP_NAME]
   --aws-time-between-requests value  duration [ns,us (or µs),ms,s,m,h (default: 10ms) [$LAYER0_AWS_TIME_BETWEEN_REQUESTS]
   --aws-max-retries value            (default: 15) [$LAYER0_AWS_MAX_RETRIES]
   --help, -h                         show help
   --version, -v                      print the version
```

api/main.go
```diff
@@ -58,10 +58,16 @@ func main() {
 		staticCreds := credentials.NewStaticCredentials(cfg.AccessKey(), cfg.SecretKey(), "")
 		awsConfig.WithCredentials(staticCreds)
 		awsConfig.WithRegion(cfg.Region())
-		awsConfig.WithMaxRetries(MAX_AWS_RETRIES)
+		awsConfig.WithMaxRetries(cfg.MaxRetries())
```
zpatrick: does this need a value? why shouldn't it be a  super high `const`  like 999?

jparsons04: The flag [defaults to 15](https://github.com/quintilesims/layer0/blob/423-retry-logic-sdk-jparsons/common/config/api.go#L92), and simultaneously allows the user to define a larger number if so desired.

zpatrick: I worry that's not enough, users would rather see a timeout before a throttled error. I think something like 50 should be the default.

jparsons04: A secondary follow-up question: Is it acceptable that this is a configurable value as I've made it to be?

common/config/api.go
diemonster: ✅

api/main.go
zpatrick: yeah i think it can't hurt to have it configurable.

PR #430, opened 2017-11-28
Setup refactor

zpatrick:
Fix setup package to accommodate the changes in api refactor.  

All cli commands for the setup tool should be tested. 

**Checklist**
- [x] Unit tests


api/controllers/swagger.go
```diff
@@ -30,6 +36,11 @@ func (s *SwaggerController) Routes() []*fireball.Route {
 	return routes
 }
 
+func (s *SwaggerController) Root(c *fireball.Context) (fireball.Response, error){
+	html := "<a href='/api/?url=/swagger.json'>Swagger</a>"
+	return fireball.NewResponse(200, []byte(html), nil), nil
```
nehayward: Nice add. I like it.

common/config/api.go
```diff
@@ -38,7 +38,7 @@ func APIFlags() []cli.Flag {
 		},
 		cli.StringFlag{
 			Name:   FLAG_AWS_REGION,
-			Value:  "us-west-2",
+			Value:  DefaultAWSRegion,
```
nehayward: Should this be all caps or no since it's not an ENV?

setup/instance/inputs.go
```diff
@@ -148,7 +146,7 @@ var Layer0ModuleInputs = []*ModuleInput{
 	{
 		Name:        INPUT_AWS_REGION,
 		Description: INPUT_AWS_REGION_DESCRIPTION,
-		Default:     aws.DEFAULT_AWS_REGION,
+		Default:     "us-west-2",
```
nehayward: Should this use the new default instead?

setup/module/api/variables.tf
```diff
@@ -25,7 +25,7 @@ variable "tags" {
 variable "linux_region_amis" {
   default = {
     us-west-1 = "ami-e5cdf385"
-    us-west-2 = "ami-4f8d912b"
+    us-west-2 = "ami-62d35c02"
```
diemonster: pull in the latest version of `api-refactor`

common/config/api.go
zpatrick: We've being making all of our `const` variables in all caps because that's what they do in python. Turns out, that's not what they do in Go; they just use upper camel case. There's something to be said for fixing all of these at once though...

zpatrick:
@diemonster I think we are saving all docs for a standalone pass

setup/instance/inputs.go
```diff
@@ -2,9 +2,7 @@ package instance
 
 import (
 	"fmt"
-
-	"github.com/Sirupsen/logrus"
-	"github.com/quintilesims/layer0/setup/aws"
+	"log"
 )
```
jparsons04: The package `github.com/quintilesims/layer0/common/config` needs to be here, otherwise the code will not compile. A reference is made to config.DefaultAWSRegion on line 149, so this package has to be imported.

jparsons04: If this is in fact intended behavior, then I would make two separate statements.

1) The help usage isn't clear enough to reflect that the flags are meant as options to exclude an instance type. It's not apparent that `--local=false` is an option looking at the usage help

2) I would argue that the UX for specifying a flag in CLI to exclude something from a listing rather than exclusively listing something is really awkward and non-intuitive.

jparsons04:
If the `BoolTFlag` and listing stuff is in fact intended behavior, then I would make two separate statements.

1) The help usage isn't clear enough to reflect that the flags are meant as options to exclude an instance type. It's not apparent that `--local=false` is an option looking at the usage help.

2) I would argue that the UX for specifying a flag in CLI to exclude something from a listing rather than exclusively listing something is really awkward and non-intuitive.

api/controllers/swagger.go
jparsons04: Rather than rendering html with a link to the base swagger definitions, could we do some sort of URL rewriting or permanent redirection if the user hits the root of the API?

zpatrick: I'm using this as a healthcheck, so i want to return a 200, not a 300

setup/command/list.go
```diff
@@ -18,23 +18,23 @@ func (f *CommandFactory) List() cli.Command {
 		Name:  "list",
 		Usage: "list local and/or remote Layer0 instances",
 		Flags: append(awsFlags,
-			cli.BoolTFlag{
-				Name:  "l, local",
-				Usage: "show local Layer0 instances, denoted by 'l' (default: true)",
+			cli.BoolFlag{
```
jparsons04: There is now an inconsistency with these changes, as you've changed this to `BoolFlag` but then didn't change the flag on [line 25](https://github.com/quintilesims/layer0/pull/430/commits/510b2456bb73f813a6dc8980087140979be22aac#diff-1f61dfa67bf4a780befc9cfc3c65388aR25) to be consistent.

Also, this is still a stylistic thing, but this still feels really awkward to me. Compare what we're doing here to what `git` does when you list branches.

`l0-setup` would show local and remote l0 instances
`l0-setup --no-local` shows just remote
`l0-setup --no-remote shows just local`

`git branch` shows you your local branches
`git branch -r` shows you just remote branches
`git branch -a` shows you local and remote branches

This git model feels much cleaner and more intuitive to me. Is there a specific reason we're designing list to behave this way?

common/config/api.go
```diff
@@ -88,9 +88,12 @@ func APIFlags() []cli.Flag {
 		},
 		cli.DurationFlag{
 			Name:   FLAG_AWS_TIME_BETWEEN_REQUESTS,
-			Value:  10 * time.Millisecond,
+			Value:  DefaultTimeBetweenRequests,
 			EnvVar: ENVVAR_AWS_TIME_BETWEEN_REQUESTS,
-			Usage:  "duration [ns,us (or µs),ms,s,m,h]",
```
jparsons04: Without the Usage help, how will a user know how to configure or specify this flag? Additionally, this usage statement follows a pattern we've used [elsewhere](https://github.com/quintilesims/layer0/blob/api-refactor/common/config/cli.go#L24)

setup/instance/apply.go
```diff
@@ -37,17 +36,17 @@ func (l *LocalInstance) waitForHealthyAPI(endpoint string, timeout time.Duration
 	client := &http.Client{Transport: tr}
 
 	for start := time.Now(); time.Since(start) < timeout; time.Sleep(time.Second * 15) {
-		logrus.Infof("Waiting for API Service to be healthy... (%s)", time.Since(start).String())
+		log.Printf("[INFP] Waiting for API Service to be healthy... (%s)", time.Since(start).String())
```
jparsons04: [nit] Typo, should be `[INFO]`

setup/command/list.go
zpatrick: we don't have to deal with conflicting flags in my pattern

common/config/api.go
zpatrick: this isn't client facing

PR #435, opened 2017-12-01
WIP: Env scaler 

nehayward:
Adds scaler functionality to env.

**Checklist**
- [ ] Unit tests

closes #


api/provider/aws/common.go
```diff
@@ -243,3 +243,47 @@ func listClusterTaskARNs(ecsapi ecsiface.ECSAPI, clusterName, startedBy, status
 
 	return taskARNs, nil
 }
+
+func listContainerInstances(ecsapi ecsiface.ECSAPI, clusterName string) ([]string, error) {
```
zpatrick: why is this is `common.go` - won't this only be used by the scaler?

api/provider/aws/common.go
```diff
@@ -243,3 +243,47 @@ func listClusterTaskARNs(ecsapi ecsiface.ECSAPI, clusterName, startedBy, status
...
+	containerInstanceARNs := []string{}
+	fn := func(output *ecs.ListContainerInstancesOutput, lastPage bool) bool {
+		for _, c := range output.ContainerInstanceArns {
```
zpatrick: Do we need to filter by the state of the container instances? For example, will this list container instances that aren't actively running?

api/provider/aws/common.go
```diff
@@ -243,3 +243,47 @@ func listClusterTaskARNs(ecsapi ecsiface.ECSAPI, clusterName, startedBy, status
...
+	input.SetCluster(clusterName)
+
+	if err := ecsapi.ListContainerInstancesPages(input, fn); err != nil {
+		return nil, err
+	}
+
+	return containerInstanceARNs, nil
+}
+
+func listClusterServiceNames(ecsapi ecsiface.ECSAPI, clusterNames []string) ([]string, error) {
```
zpatrick: don't we already have a helper that does this?

api/provider/aws/common.go
nehayward: @sesh-kebab

api/provider/aws/environment_scaler.go
```diff
@@ -1,14 +1,629 @@
...
+				}
+			}
+		}
+
+		inUse := aws.Int64Value(instance.PendingTasksCount)+aws.Int64Value(instance.RunningTasksCount) > 0
+		r := &models.ResourceProvider{
+			ID:              aws.StringValue(instance.Ec2InstanceId),
+			InUse:           inUse,
+			UsedPorts:       usedPorts,
+			AvailableMemory: fmt.Sprintf("%v", availableMemory), //todo: this is definitely not correct
```
jparsons04: Assuming that the `AvailableMemory` field here accepts a string that is a numeric value and the unit (for instance, 10MB), could you not leverage the [Format()](https://github.com/zpatrick/go-bytesize/blob/master/bytesize.go#L112) method in go-bytesize?

api/provider/aws/environment_scaler.go
```diff
@@ -1,14 +1,629 @@
...
+		containerInstanceARNs = append(containerInstanceARNs, output.ContainerInstanceArns...)
+
+		return !lastPage
+	}
+
+	if err := e.Client.ECS.ListContainerInstancesPages(listContainerInstancesInput, listContainerInstancesPagesFN); err != nil {
+		return nil, err
+	}
+
+	describeContainerInstancesInput := &ecs.DescribeContainerInstancesInput{}
```
sesh-kebab: Might as well check for a 0 length of `containerInstanceARNs` returned. If 0 we don't need to proceed any further. I.e., 

```go
result := []*models.ResourceProvider{}
if len(containerInstanceARNs) == 0 {
	return result, nil
}
```

api/provider/aws/environment_scaler.go
```diff
@@ -1,14 +1,629 @@
...
-	return fmt.Errorf("EnvironmentScaler not implemented")
+	// GET RESOURCE PROVIDERS
+	clusterName := addLayer0Prefix(e.Config.Instance(), environmentID)
+
+	resourceProviders, err := e.getResourceProviders(clusterName)
+	if err != nil {
+		return err
+	}
+
+	log.Printf("[DEBUG] resourceProviders for env '%s': %#v", environmentID, resourceProviders)
```
jparsons04: [nit] Could we print this only if some sort of debug flag is set in configuration?

api/provider/aws/environment_scaler.go
```diff
@@ -1,14 +1,629 @@
...
+
+	// GET RESOURCE CONSUMERS
+	resourceConsumers, err := e.getResourceConsumers(clusterName)
+	if err != nil {
+		return err
+	}
+
+	log.Printf("[DEBUG] resourceConsumers for env '%s': %#v", environmentID, resourceConsumers)
```
jparsons04: [nit] Same question about debug flags

api/provider/aws/environment_scaler.go
```diff
@@ -1,14 +1,629 @@
...
+
+	info, err := e.scale(clusterName, resourceProviders, resourceConsumers)
+	if err != nil {
+		return err
+	}
+	fmt.Println(info)
```
jparsons04: Do we intend to output this model to the terminal long term?

api/provider/aws/environment_scaler.go
jparsons04: Disregard. I can see in several other areas within this package that DEBUG statements are freely printed. It must just be a matter of log level specification for the event happening. Seeing this, I don't think the debug flag question is worth looking into at this point.

api/provider/aws/environment_scaler.go
jparsons04: Disregard for the same reason as stated above.

api/provider/aws/environment_scaler.go
zpatrick: statements that start with `[DEBUG]` only print if the debug log level is set

api/provider/aws/environment_scaler.go
```diff
@@ -1,14 +1,629 @@
...
+	// check if we need to scale up
+	for _, consumer := range consumers {
+		hasRoom := false
+
+		// first, sort by memory so we pack tasks by memory as tightly as possible
+		sortProvidersByMemory(providers)
+
+		// next, place any unused providers in the back of the list
+		// that way, we can can delete them if we avoid placing any tasks in them
+		sortProvidersByUsage(providers)
```
jparsons04: I may be misunderstanding how the sort package works with this, so please correct me if I'm wrong. If you sort the same object (the length and members are unchanged after the first sort) two consecutive times, mathematically speaking, wouldn't the second sort just simply override what the first sort did?  These functions don't appear to manipulate (add or remove) any items into the slice that I can see, just simply sorting them. Assuming the length and members of `providers` is unchanged after the first sort, I don't think the second sort operation cares what the first sort did.

jparsons04: I'm implying here that only `sortProvidersByUsage` is required.

api/provider/aws/environment_scaler.go
```diff
@@ -1,14 +1,629 @@
...
+	default:
+		log.Printf("Environment %s is at desired scale of %d. No scaling action required.", clusterName, scale)
+		return currentCapacity, nil
+	}
+}
+
+func (e *EnvironmentScaler) scaleUp(ecsEnvironmentID string, scale int, asg *autoscaling.Group) (int, error) {
+	maxCapacity := int(aws.Int64Value(asg.MaxSize))
+	if scale > maxCapacity {
+		input := &autoscaling.UpdateAutoScalingGroupInput{}
```
jparsons04: [nit] For consistency, since you have `log.Printf()` statements for `scaleDown` when scale < minCapacity, I think similar logging statements should be present in this method when scale > maxCapacity.

api/provider/aws/test_aws/environment_scaler_test.go
```diff
@@ -0,0 +1,9 @@
+package test_aws
+
+import (
+	"testing"
+)
+
+func TestEnvironmentScale(t *testing.T) {
```
zpatrick: You should add in the old tests back in verbatim - they are really extensive and do a good job of catching corner cases

api/provider/aws/environment_scaler.go
zpatrick: @jparsons04 `sortProvidersByMemory` uses an integer comparison, `sortProvidersByUsage` just does a boolean comparison. Doing these sorts allows unused providers to bubble to the end of the list - e.g. it's the last candidate we should try to place a resource in. 

This is important because if we have unused providers at the end of this process, we know we can scale down.

api/provider/aws/common.go
sesh-kebab: Changes to common can be removed. It doesn't look like the latest set of changes that @nehayward and @tlake worked on use these methods.

api/provider/aws/common.go
sesh-kebab: Agreed - although this method isn't being used any longer, the same request being made from `environment_scaler.go` could would want to specify the `Status` property of the `ListContainerInstancesInput` (where status == ACTIVE)

api/provider/aws/common.go
sesh-kebab: Not that i'm aware of. That was I think one of the biggest issues with the existing code where it looked like we were getting services from all environments and then filter for a particular environment. I'm sure this would have been a major if not the biggest contributor to the rate limits we were seeing.

sesh-kebab: Also same thing applies - this code is no longer used.

api/provider/aws/environment_scaler.go
```diff
@@ -1,14 +1,629 @@
...
+	"github.com/quintilesims/layer0/common/errors"
+	"github.com/quintilesims/layer0/common/models"
+)
 
 type EnvironmentScaler struct {
+	Client              *awsc.Client
+	EnvironmentProvider provider.EnvironmentProvider
+	ServiceProvider     provider.ServiceProvider
+	TaskProvider        provider.TaskProvider
+	JobStore            job.Store
```
sesh-kebab: We shouldn't need the job store.

api/provider/aws/environment_scaler.go
```diff
@@ -1,14 +1,629 @@
...
+	Config              config.APIConfig
+	deployCache         map[string][]models.ResourceConsumer
```
sesh-kebab: Not thread safe (or should that be go routine safe). Not that scale can be called concurrently via the JobRunner and the Scaler timer. Maybe a RWMutex or a concurrent collection?

api/provider/aws/environment_scaler.go
```diff
@@ -1,14 +1,629 @@
...
+			AvailableMemory: provider.AvailableMemory,
+		}
+		providerModels[i] = newProvider
+	}
+
+	return providerModels
+}
+
+func (e *EnvironmentScaler) getResourceProviders(clusterName string) ([]*models.ResourceProvider, error) {
+	listContainerInstancesInput := &ecs.ListContainerInstancesInput{}
```
sesh-kebab: [nit] I know we have a guideline of not calling helper methods from other helper methods but it sure would be cleaner in my opinion to wrap the aws calls in their own method. So getResourceProviders would look something like

```go
func (e *EnvironmentScaler) getResourceProviders(clusterName string) ([]*models.ResourceProvider, error) {
   clusterInstances, err := getContainerInstances(clusterName)
   if err != nil ...

```

api/provider/aws/environment_scaler.go
```diff
@@ -1,14 +1,629 @@
...
+		return nil, err
+	}
+
+	result := []*models.ResourceProvider{}
+	if len(output.ContainerInstances) == 0 {
+		return result, nil
+	}
+
+	for _, instance := range output.ContainerInstances {
+		if !aws.BoolValue(instance.AgentConnected) {
```
sesh-kebab: As discussed in slack - there is an incorrect assumption being made here. Either remove this (as we don't handle this case when container instance count == min cluster count. Or would need to check the connection state for a duration. I.e., if disconnected for more than 3 minutes then consider it permanently disconnected.

In practice - checking for a certain duration will add considerable complexity which I don't think is warranted given:

1. we don't do anything even if the agent is disconnected when current cluster count == min cluster count
2. how often does this situation actually happen?

api/provider/aws/environment_scaler.go
```diff
@@ -1,14 +1,629 @@
...
+			continue
+		}
+
+		if aws.StringValue(instance.Status) != "ACTIVE" {
```
sesh-kebab: Can be added to the request rather than filtering for it here.

api/provider/aws/environment_scaler.go
```diff
@@ -1,14 +1,629 @@
...
+		return !lastPage
+	}
+
+	if err := e.Client.ECS.ListServicesPages(listServicesInput, listServicesPagesFN); err != nil {
+		return nil, err
+	}
+
+	services := []*ecs.Service{}
+	if len(serviceARNs) > 0 {
+		// The SDK states that you can specify up to 10 services in one DescribeServices operation:
```
sesh-kebab: 👍

api/provider/aws/environment_scaler.go
```diff
@@ -1,14 +1,629 @@
...
+			if err != nil {
+				return nil, err
+			}
+
+			resourceConsumers = append(resourceConsumers, c...)
+		}
+	}
+
+	// GET PENDING TASK RESOURCE CONSUMERS IN ECS
+	taskSummaries, err := e.TaskProvider.List()
```
sesh-kebab: ❗️ This is super inefficient. Services had a similar pattern (in develop) where all entities are querie, cluster by cluster for all the clusters. Then you are throwing away most of the results by filtering for a single environment cluster.

Lets only query the tasks from the environment that we're currently scaling.

api/provider/aws/environment_scaler.go
```diff
@@ -1,14 +1,629 @@
...
+			if err != nil {
+				return nil, err
+			}
+
+			resourceConsumers = append(resourceConsumers, taskResourceConsumers...)
+		}
+	}
+
+	// GET PENDING TASK RESOURCE CONSUMERS IN JOBS
+	jobs, err := e.JobStore.SelectAll()
```
sesh-kebab: As discussed in slack - we don't need to consider jobs as resource consumers any longer as we execute the jobs via the JobRunner and no ecs tasks.

api/provider/aws/environment_scaler.go
```diff
@@ -1,14 +1,629 @@
...
+		return nil, err
+	}
+
+	var instanceType string
+	if len(config.LaunchConfigurations) > 0 {
+		instanceType = aws.StringValue(config.LaunchConfigurations[0].InstanceType)
+	}
+
+	// these ports are automatically used by the ecs agent
+	defaultPorts := []int{
```
sesh-kebab: Can we move this to a const at the top of the file.

api/provider/aws/environment_scaler.go
```diff
@@ -1,14 +1,629 @@
...
+				}
+			}
+		}
+	}
+
+	return resourceConsumers, nil
+}
+
+func (e *EnvironmentScaler) calculateNewProvider(clusterName string) (*models.ResourceProvider, error) {
+	input := &autoscaling.CreateAutoScalingGroupInput{}
```
sesh-kebab: Why are we creating an AutoScalingGroup here? The whole purpose seems to be to get the InstanceType from a launch config. Can we not get this from environment.size?

1. Create ASG
2. Describe the launch config for the newly created group

api/provider/aws/environment_scaler.go
```diff
@@ -1,14 +1,629 @@
...
+	resource.ID = "<new instance>"
+	resource.InUse = false
+	resource.UsedPorts = defaultPorts
+	resource.AvailableMemory = instanceType
+	return resource, nil
+}
+
+func (e *EnvironmentScaler) getContainerResourceFromDeploy(deployID string) ([]models.ResourceConsumer, error) {
+	// use some kind of deploy cache
+	if consumers, ok := e.deployCache[deployID]; ok {
```
sesh-kebab: See previous comment about thread-safety. The easiest approach I suppose could be to have two instances of scaler (one for scaling and one for the job runner).

api/provider/aws/environment_scaler.go
jparsons04: Ok, then that makes sense if the sorts are operating on numeric then boolean comparisons.

api/provider/aws/environment_scaler.go
nehayward: Good catch, It looks like we could.

api/provider/aws/test_aws/environment_scaler_test.go
nehayward: Thanks, I was wondering about that. I'll start working on that too.

api/provider/aws/environment_scaler.go
nehayward: Would the aws api be used directly to filter input for single env

api/provider/aws/environment_scaler.go
nehayward: We could remove #L378-399 and then just get available memory from env size, using tag store right?

api/provider/aws/environment_scaler.go
sesh-kebab: Yup exactly that. If we look in the implementation for `e.TaskProvivider.List()`, it does an equivalent of an `environment.List()` and then calls a helper method called `listClusterTaskARNs` for each cluster/env. I think either exposing the listClusterTaskARNs to the provider somehow or duplicate the code in the scaler would be good approach (probably the latter might be simple to start with).

api/provider/aws/environment_scaler.go
nehayward: Hmm maybe use this package? https://github.com/orcaman/concurrent-map

zpatrick: or possible [go-cache](https://github.com/zpatrick/go-cache) :trollface:

api/provider/aws/environment_scaler.go
```diff
@@ -358,6 +362,33 @@ func (e *EnvironmentScaler) getResourceConsumers(clusterName string) ([]models.R
...
 	}
 
+	// GET PENDING TASK RESOURCE CONSUMERS IN JOBS
+	jobs, err := e.JobStore.SelectAll()
+	if err != nil {
+		return nil, err
+	}
+
+	for _, job := range jobs {
+		if job.Type == models.CreateDeployJob {
```
zpatrick: `CreateTaskJob`

api/provider/aws/environment_scaler.go
zpatrick: Not all jobs - but the `CreateTaskJobs` that are `InProgress` do have pending resources

api/provider/aws/environment_scaler.go
```diff
@@ -1,14 +1,656 @@
...
+
+	_, err = e.scale(clusterName, resourceProviders, resourceConsumers)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (e *EnvironmentScaler) scale(clusterName string, providers []*models.Resource, consumers []*models.Resource) (*models.ScalerRunInfo, error) {
```
diemonster: It would be nice to reduce the overall spaghetti code and/or improve the naming of some of these functions. I still really like the pattern we landed on in the `provider/aws` layer where we used the main exported function (in this case, `Scale`) as a 'master function' for calling all the related helpers. We could do something similar here and I think it would greatly aid clarity.

api/provider/aws/environment_scaler.go
```diff
@@ -1,14 +1,656 @@
...
+		ScaleBeforeRun:          scaleBeforeRun,
+		DesiredScaleAfterRun:    desiredScale,
+		ActualScaleAfterRun:     actualScale,
+		UnusedResourceProviders: len(unusedProviders),
+	}
+
+	return info, errors.MultiError(errs)
+}
+
+func resourceModels(resources []*models.Resource) []models.Resource {
```
diemonster: This function needs a better name

api/provider/aws/environment_scaler.go
```diff
@@ -1,14 +1,656 @@
...
+			CPU:    availableCPU.Megabytes(),
+		}
+
+		result = append(result, r)
+	}
+
+	return result, nil
+}
+
+func (e *EnvironmentScaler) getResourceConsumers(clusterName string) ([]*models.Resource, error) {
```
diemonster: This function needs to be broken up

api/provider/aws/test_aws/environment_scaler_test.go
```diff
@@ -0,0 +1,10 @@
+package test_aws
+
+import (
+	"testing"
+)
+
+// Add Old Test Verbatim from Dev PR
```
diemonster: Thinking those tests will need to be totally rewritten, I don't know how worthwhile they were to begin with.

api/provider/aws/environment_scaler.go
```diff
@@ -1,14 +1,656 @@
...
+	}
+
+	if cpuInstanceSize > memoryInstanceSize {
+		return CPU
+	}
+
+	return Memory
+}
+
+func hasResourcesFor(consumer models.Resource, provider models.Resource) bool {
```
diemonster: needs a better function name

api/provider/aws/environment_scaler.go
```diff
@@ -1,14 +1,656 @@
...
+			if wanted == used {
+				return false
+			}
+		}
+	}
+
+	return consumer.Memory <= provider.Memory
+}
+
+func subtractResourcesFor(consumer *models.Resource, provider *models.Resource) error {
```
diemonster: needs a better function name

api/provider/aws/environment_scaler.go
```diff
@@ -1,14 +1,656 @@
...
+
+	provider.Ports = append(provider.Ports, consumer.Ports...)
+	provider.Memory -= consumer.Memory
+	provider.CPU -= consumer.CPU
+	provider.InUse = true
+
+	return nil
+}
+
+func sortByMemory(resources ...[]*models.Resource) {
```
diemonster: More naming gripes. With all of these sort functions: sort how?

api/provider/aws/test_aws/environment_scaler_test.go
nehayward: I’ve already started on testing and I’m using all the mockaws calls. I’ll put something up today to view. However I’m waiting on tlake to finish scaler so it’ll have to be adjusted. Also breaking up the code will help.

api/provider/aws/environment_scaler.go
nehayward: I agree completely.

nehayward: I agree completely.

api/provider/aws/test_aws/environment_scaler_test.go
```diff
@@ -0,0 +1,351 @@
...
+	ResourceConsumers []models.Resource
+}
+
+// Add Old Test Verbatim from Dev PR
+
+func (e *EnvironmentScalerUnitTest) Run(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockAWS := awsc.NewMockClient(ctrl)
```
zpatrick: this all needs to be split out - we should be testing the scaling algorithm as a single unit

api/provider/aws/test_aws/environment_scaler_test.go
```diff
@@ -0,0 +1,351 @@
...
+	// 	Return(nil)
+
+	environmentScaler := provider.NewEnvironmentScaler(mockAWS.Client(), mockEnvironmentProvider, mockServiceProvider, mockTaskProvider, mockJobStore, mockConfig)
+
+	if err := environmentScaler.Scale("env_id"); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestResourceManagerScaleUp_noProviders(t *testing.T) {
```
zpatrick: Seems like these should jsut be table tests

api/provider/aws/test_aws/environment_scaler_test.go
```diff
@@ -0,0 +1,351 @@
...
+				InUse:  true}},
+		ResourceConsumers: []models.Resource{
+			{Ports: []int{80}},
+		},
+	}
+
+	test.Run(t)
+}
+
+func TestResourceManagerNoScale_noResourceConsumers(t *testing.T) {
```
zpatrick: Seems like we are missing a lot of test cases

common/models/resource.go
```diff
@@ -0,0 +1,9 @@
+package models
+
+type Resource struct {
```
zpatrick: Not sure why this is a model - we don't use these objects to display anything to users

common/models/scaler_run_info.go
```diff
@@ -1,11 +1,11 @@
 package models
 
 type ScalerRunInfo struct {
```
zpatrick: [nit] I thought we were removing this?

common/models/resource.go
diemonster: Good point. Maybe it's a holdover from when we talked (?) about having the scaler be a standalone thing? Regardless, I agree, these structs should be moved locally.

api/provider/aws/environment_scaler.go
zpatrick: these functions already exist on the resource consumer - why are they duplicated here?

api/provider/aws/environment_scaler.go
zpatrick: ditto

api/provider/aws/environment_scaler.go
zpatrick: why does this function exist?

api/provider/aws/test_aws/environment_scaler_test.go
zpatrick: I totally disagree with @bivers. The old tests only testing the scaling algorithm (these need to be properly split up to achieve that), and had a ton of test cases. It's very difficult to get all of the old test cases to pass, because they were specifically designed to trip up a non-optimal algorithm. They should be added verbatim, or one needs to make a strong case as to why any should be left out.

common/models/scaler_run_info.go
tlake: Ooh, I didn't catch this struct. I removed the usage of the struct in `environment_scaler.go` over in #440 PR, but not the struct. I'll fix that in #440.

common/models/resource.go
tlake: IIRC (and @nehayward please correct me if I'm wrong), but this was part of an experiment to see if it were absolutely necessary for `ResourceProvider` and `ResourceConsumer` to be two different things. I think the conclusion was that yes, they do need to be separate, and so this would want to go away.

api/provider/aws/environment_scaler.go
tlake: FWIW, I didn't see these sorts of functions ever defined as methods on a `ResourceProvider` or `ResourceConsumer`. I only ever saw these functions as standalone helper functions that took a provider and a consumer, and it was a little opaque in regards to what was happening without reading through the function itself.

The changes in #440 have refactored `HasResourcesFor` and `SubtractResourcesFor` into methods on `ResourceProvider`.

api/provider/aws/environment_scaler.go
tlake: This will be broken up into three functions as a result of #440.

api/provider/aws/environment_scaler.go
tlake: We don't know - this was something we found in `develop` and we weren't sure what the heck the point of it was.

I've removed it over in #440.

✅

api/provider/aws/environment_scaler.go
tlake: I think you'll find some relief over in #440. :D

api/provider/aws/environment_scaler.go
tlake: ✅ (#440)

common/models/resource.go
tlake: Now I understand what you all meant from a comment over in #440 - I'll make this change over there.

✅

nehayward: Yup it was. @tlake.

PR #436, opened 2017-12-01
add auth decorator. fix debug flag in module dockerrun

zpatrick:
Adds auth decorator to api. It also fixes some other small issues:
* Debug environment variable in setup module 
* Fix job status references in terraform unit tests

Try running the cli against your local api with valid/invalid auth token


jparsons04:
:white_check_mark: 

Manually tested valid and invalid base64 encoded `--token` values, got expected results. Provided valid base64 but invalid user:pass format and got the expected result. Provided invalid credentials and also got the expected result.

Also manually tested through Swagger and was correctly challenged. Was able to auth successfully.

common/config/api.go
```diff
@@ -106,6 +110,8 @@ func APIFlags() []cli.Flag {
 
 type APIConfig interface {
 	Port() int
+	AuthToken() string
```
sesh-kebab: [nit] doesn't seem necessary to expose `AuthToken` via the interface as it isn't being consumed anywhere.

sesh-kebab:
> Debug environment variable in setup module

I can't see a change related to this. Can you please point me to it?

zpatrick:
@sesh-kebab may have been in a different pr - was this line: https://github.com/quintilesims/layer0/pull/436/files#diff-1eb4fffac76b6390965be04a7bd8b510R39

PR #437, opened 2017-12-05
Cleanup of system and stress test packages

jparsons04:
There are a number of cleanup tasks this PR does:

* Removes all references and usage of the `logrus` package
* Reintroduces `WaitFor()` from the `develop` branch's `testutils` package. 
* Prepends all log output lines with `[DEBUG]` or `[ERROR]` where applicable
* Removes references to the old `logutils` packages and instead uses the standard `log` package
* Cleans up the layer0 test client to more closely align with layer0 client methods
* Pulls in fixes from the `zpatrick/fireball` package and the `quintilesims/tftest` packages

A future PR will be required to fix the functionality of system and stress tests themselves. As of right now, other work in progress needs to be completed before the tests can be fixed. This is primarily to clean up compilation issues and general tidiness in the packages. The tests will run but fail.

`go test ./... -v -dry` should execute without any compilation issues. Test failures are expected.

**Checklist**
~- [ ] Unit test~


tests/stress/main_test.go
```diff
@@ -3,22 +3,23 @@ package system
...
+	"github.com/quintilesims/layer0/common/logging"
 )
 
 var (
-	dry   = flag.Bool("dry", false, "Perform a dry run - don't execute terraform 'apply' commands")
-	debug = flag.Bool("debug", false, "Print debug statements")
-	log   = logutils.NewStandardLogger("Test")
+	dry    = flag.Bool("dry", false, "Perform a dry run - don't execute terraform 'apply' commands")
+	debug  = flag.Bool("debug", false, "Print debug statements")
+	logger = logging.NewLogWriter(*debug)
```
zpatrick: You don't need this variable (also, fwiw, if we did, you would need to call `flag.Parse()` first).

After `flag.Parse()`, you need to do the following:
```
logger := logging.NewLogWriter(debug)
log.SetOutput(logger)
```

and just logging during the tests should be done with `log.Printf("[DEBUG] ...`

tests/stress/main_test.go
```diff
@@ -46,7 +39,7 @@ func setup() {
 func teardown() {
 	if !*dry {
 		if err := filepath.Walk("cases", deleteStateFiles); err != nil {
-			fmt.Println("Error occurred during teardown: ", err)
+			log.Fatalf("[ERROR] Error occurred during teardown: ", err)
```
zpatrick: you can remove `os.Exit` if you use `log.Fatal`

tests/clients/layer0_test_client.go
```diff
@@ -60,43 +81,77 @@ func (l *Layer0TestClient) CreateLoadBalancer(name, environmentID string) *model
...
 
-func (l *Layer0TestClient) CreateService(name, environmentID, deployID, loadBalancerID string) *models.Service {
-	service, err := l.Client.CreateService(name, environmentID, deployID, loadBalancerID)
+func (l *Layer0TestClient) CreateService(name, environmentID, deployID, loadBalancerID string) string {
+	req := models.CreateServiceRequest{
+		DeployID:       deployID,
+		EnvironmentID:  environmentID,
+		LoadBalancerID: loadBalancerID,
+		ServiceName:    name,
+		Scale:          1,
```
diemonster: [nit] does this have to be set?

tests/clients/sts_test_client.go
```diff
@@ -28,15 +27,14 @@ func NewSTSTestClient(t *testing.T, url string) *STSTestClient {
...
 		if _, err := s.Client.GetHealth(); err != nil {
-			logrus.Debug(err)
-			return false
+			log.Printf(err.Error())
 		}
+	}
 
-		return true
-	})
+	log.Fatalf("[ERROR] Timeout reached after %v", timeout)
```
diemonster: This is a case where I can kinda see Dave Cheney's point re: not logging `[ERROR]`s, or at least labeling them as such. Reaching a timeout is an event that is handled by us (to some degree) and results in that test not passing, which is not necessarily unexpected behavior.

tests/stress/main_test.go
```diff
@@ -27,27 +26,19 @@ func TestMain(m *testing.M) {
...
-
-	logutils.SetGlobalLogger(log)
-
+	logger := logging.NewLogWriter(*debug)
+	log.SetOutput(logger)
 	if !*dry {
 		if err := filepath.Walk("cases", deleteStateFiles); err != nil {
-			fmt.Println("Error occurred during setup: ", err)
-			os.Exit(1)
+			log.Fatalf("[ERROR] Error occurred during setup: ", err)
```
diemonster: Continuing on from the Dave Cheney mindset, why have this go to logs?

tests/clients/layer0_test_client.go
jparsons04: It looks like the answer is no given that the `aws` package will set this value to 1 if no value is specified in the request.

tests/clients/sts_test_client.go
jparsons04: Let's say I remove this line and the timeout condition happens. I'm not sure that the tests that call `sts.WaitForHealthy` would actually show failure or fail. It seems as though I should probably treat this as an informational level message like `log.Printf("[DEBUG] Timeout reached after %v", timeout)`

tests/stress/main_test.go
jparsons04: I could instead do something like `log.Printf(err.Error())`?

tests/clients/sts_test_client.go
diemonster: yep that's what I was thinking

tests/stress/main_test.go
diemonster: Or just do what the code did previously

jparsons04: :white_check_mark:

tests/clients/sts_test_client.go
jparsons04: :white_check_mark:

tests/stress/taskStress_test.go
```diff
@@ -35,38 +34,40 @@ func TestTaskStress(t *testing.T) {
...
 		if taskSummary.EnvironmentID == environmentID {
-			task := s.Layer0.GetTask(taskSummary.TaskID)
-			detail := task.Copies[0].Details[0]
+			task := s.Layer0.ReadTask(taskSummary.TaskID)
+			container := task.Containers[0]
 
-			if detail.ExitCode != 0 {
-				t.Fatalf("Task %s has unexpected exit code: %#v", task.TaskID, detail)
+			if container.ExitCode != 0 {
+				t.Fatalf("[ERROR] Task %s has unexpected exit code: %#v", task.TaskID, container)
```
zpatrick: only log messages have log levels

tests/system/datasources_test.go
```diff
@@ -58,5 +60,5 @@ func TestDataSources(t *testing.T) {
 	checkOutput("service_environment_name")
 	checkOutput("service_scale")
 
-	log.Debugf("L0 Terraform Provider Data sources Tests completed.")
+	log.Printf("[DEBUG] L0 Terraform Provider Data sources Tests completed.")
```
zpatrick: [nit] this log isn't necessary

tests/system/deadServiceRecreated_test.go
```diff
@@ -25,15 +25,26 @@ func TestDeadServiceRecreated(t *testing.T) {
...
-	})
-
-	testutils.WaitFor(t, time.Second*10, time.Minute*2, func() bool {
-		log.Debugf("Waiting for service to recreate")
-		service := s.Layer0.GetService(serviceID)
-		return service.RunningCount == 1
-	})
+	log.Printf("[DEBUG] Waiting for service to die")
+	service := s.Layer0.ReadService(serviceID)
+	for start := time.Now(); time.Since(start) < time.Minute*2; time.Sleep(time.Second * 10) {
```
zpatrick: seems like we should have moved that helper function to this package

tests/system/deadServiceRecreated_test.go
```diff
@@ -25,15 +25,26 @@ func TestDeadServiceRecreated(t *testing.T) {
...
+		if service.RunningCount == 0 {
+			continue
+		}
+	}
+
+	if service.RunningCount != 0 {
+		t.Fatalf("[ERROR] Timeout reached after %v", time.Minute*2)
```
zpatrick: only logs have levels

tests/system/deadServiceRecreated_test.go
```diff
@@ -25,15 +25,26 @@ func TestDeadServiceRecreated(t *testing.T) {
...
+	}
+
+	log.Printf("[DEBUG] Waiting for service to recreate")
+	for start := time.Now(); time.Since(start) < time.Minute*2; time.Sleep(time.Second * 10) {
+		if service.RunningCount == 1 {
```
zpatrick: shouldn't this be `!= 1`?

tests/system/deadServiceRecreated_test.go
```diff
@@ -25,15 +25,26 @@ func TestDeadServiceRecreated(t *testing.T) {
...
-
-	testutils.WaitFor(t, time.Second*10, time.Minute*2, func() bool {
-		log.Debugf("Waiting for service to recreate")
-		service := s.Layer0.GetService(serviceID)
-		return service.RunningCount == 1
-	})
+	log.Printf("[DEBUG] Waiting for service to die")
+	service := s.Layer0.ReadService(serviceID)
+	for start := time.Now(); time.Since(start) < time.Minute*2; time.Sleep(time.Second * 10) {
+		if service.RunningCount == 0 {
```
zpatrick: shouldn't this be `!= 0`?

zpatrick: you need to re-read the service in this loop

tests/system/deadServiceRecreated_test.go
zpatrick: you need to re-read the servie in this loop

tests/system/environmentLink_test.go
```diff
@@ -20,7 +20,6 @@ func TestEnvironmentLink(t *testing.T) {
 	defer s.Terraform.Destroy()
 
 	publicEnvironmentID := s.Terraform.Output("public_environment_id")
-	privateEnvironmentID := s.Terraform.Output("private_environment_id")
```
zpatrick: why is this gone?

tests/system/environmentLink_test.go
```diff
@@ -29,38 +28,31 @@ func TestEnvironmentLink(t *testing.T) {
...
 	// the private service returns "Hello, World!" from its root path
-	testutils.WaitFor(t, time.Second*10, time.Minute*5, func() bool {
-		log.Printf("Running curl while link exists")
+	log.Printf("[DEBUG] Running curl while link exists")
+	for start := time.Now(); time.Since(start) < time.Minute*5; time.Sleep(time.Second * 10) {
 		output, err := publicService.RunCommand("curl", "-m", "10", "-s", privateServiceURL)
 		if err != nil {
-			log.Printf("Error running curl: %v", err)
-			return false
+			t.Fatalf("[ERROR] Error running curl: %v", err)
```
zpatrick: this shouldn't be fatal, we should log and retry

tests/system/environmentLink_test.go
```diff
@@ -29,38 +28,31 @@ func TestEnvironmentLink(t *testing.T) {
...
 		}
 
 		if expected := "Hello, World!"; output != expected {
-			log.Printf("Output from curl was '%s', expected '%s'", output, expected)
-			return false
+			t.Fatalf("[ERROR] Output from curl was '%s', expected '%s'", output, expected)
```
zpatrick: only logs have levels

tests/system/environmentLink_test.go
zpatrick: only logs have level s

tests/system/environmentLink_test.go
```diff
@@ -29,38 +28,31 @@ func TestEnvironmentLink(t *testing.T) {
...
-
-	testutils.WaitFor(t, time.Second*10, time.Minute*2, func() bool {
-		log.Println("Running curl without link")
+	log.Printf("[DEBUG] Running curl without link")
+	for start := time.Now(); time.Since(start) < time.Minute*2; time.Sleep(time.Second * 10) {
 		output, err := publicService.RunCommand("curl", "-m", "10", "-s", privateServiceURL)
 		if err != nil {
-			log.Printf("Error running curl: %v", err)
-			return false
+			t.Fatalf("[ERROR] Error running curl: %v", err)
```
zpatrick: log + retry

tests/system/environmentLink_test.go
```diff
@@ -29,38 +28,31 @@ func TestEnvironmentLink(t *testing.T) {
...
 		}
 
 		if output != "" {
-			log.Printf("Output from curl was '%s', expected no output", output)
-			return false
+			t.Fatalf("[ERROR] Output from curl was '%s', expected no output", output)
```
zpatrick: only logs have levels

tests/system/import_test.go
```diff
@@ -16,14 +17,21 @@ func TestImport(t *testing.T) {
 
 	data, err := ioutil.ReadFile("cases/modules/sts/Dockerrun.aws.json")
 	if err != nil {
-		t.Fatalf("Failed to read dockerrun: %v", err)
+		t.Fatalf("[ERROR] Failed to read dockerrun: %v", err)
```
zpatrick: ditto

tests/system/main_test.go
```diff
@@ -27,27 +26,19 @@ func TestMain(m *testing.M) {
...
-
-	logutils.SetGlobalLogger(log)
-
+	logger := logging.NewLogWriter(*debug)
+	log.SetOutput(logger)
 	if !*dry {
 		if err := filepath.Walk("cases", deleteStateFiles); err != nil {
-			fmt.Println("Error occurred during setup: ", err)
-			os.Exit(1)
+			log.Println("[ERROR] Error occurred during setup: ", err)
```
zpatrick: log.Fatal or keep ox.Exit

tests/system/main_test.go
```diff
@@ -27,27 +26,19 @@ func TestMain(m *testing.M) {
...
 		}
 	}
 }
 
 func teardown() {
 	if !*dry {
 		if err := filepath.Walk("cases", deleteStateFiles); err != nil {
-			fmt.Println("Error occurred during teardown: ", err)
-			os.Exit(1)
+			log.Println("[ERROR] Error occurred during teardown: ", err)
```
zpatrick: ditto

tests/system/serviceScale_test.go
```diff
@@ -18,18 +17,30 @@ func TestServiceScale(t *testing.T) {
 	defer s.Terraform.Destroy()
 
 	serviceID := s.Terraform.Output("service_id")
+	deployID := s.Terraform.Output("deploy_id")
+
+	s.Layer0.UpdateService(serviceID, deployID, 3)
+	log.Printf("[DEBUG] Waiting for service to scale up")
+	service := s.Layer0.ReadService(serviceID)
+	for start := time.Now(); time.Since(start) < time.Minute*5; time.Sleep(time.Second * 10) {
```
zpatrick: It seems like there is logical inverse bug introduced everytime `logutils.WaitFor` was replaced. Let's just move that helper function to this package and not change the internal logic. I'm going to stop commenting on them for the rest of the pr to save time.

tests/system/serviceScale_test.go
```diff
@@ -18,18 +17,30 @@ func TestServiceScale(t *testing.T) {
...
-	})
-
-	s.Layer0.ScaleService(serviceID, 1)
-	testutils.WaitFor(t, time.Second*10, time.Minute*5, func() bool {
-		log.Debugf("Waiting for service to scale down")
-		service := s.Layer0.GetService(serviceID)
-		return service.RunningCount == 1
-	})
+	if service.RunningCount != 1 {
+		t.Fatalf("[ERROR] Timeout reached after %v", time.Minute*5)
```
zpatrick: ditto

tests/system/windowsService_test.go
```diff
@@ -25,26 +24,28 @@ func TestWindowsService(t *testing.T) {
...
+		t.Fatalf("[ERROR] Timeout reached after %v", time.Minute*45)
+	}
+
+	log.Printf("[DEBUG] Waiting for service to be healthy")
+	for start := time.Now(); time.Since(start) < time.Minute*10; time.Sleep(time.Second * 30) {
 		resp, err := http.Get(serviceURL)
 		if err != nil {
-			log.Debugf("There was an error checking the window service's url: %v", err)
-			return false
+			t.Fatalf("[ERROR] There was an error checking the Windows service's URL: %v", err)
```
zpatrick: ditto

tests/system/windowsService_test.go
```diff
@@ -25,26 +24,28 @@ func TestWindowsService(t *testing.T) {
...
 		}
 
 		defer resp.Body.Close()
 		if resp.StatusCode < 200 || resp.StatusCode > 299 {
-			log.Debugf("Windows service returned non-200 status: %d", resp.StatusCode)
-			return false
+			t.Fatalf("[ERROR] Windows service returned non-200 status: %d", resp.StatusCode)
```
zpatrick: ditto

tests/clients/sts_test_client.go
```diff
@@ -28,15 +27,14 @@ func NewSTSTestClient(t *testing.T, url string) *STSTestClient {
...
 		if _, err := s.Client.GetHealth(); err != nil {
-			logrus.Debug(err)
-			return false
+			log.Println(err.Error())
 		}
+	}
 
-		return true
-	})
+	log.Printf("[DEBUG] Timeout reached after %v", timeout)
```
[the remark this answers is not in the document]
jparsons04: The `WaitForHealthy` helper function is not passed `T`, so it has no context for testing in this specific case.

tests/system/environmentLink_test.go
jparsons04: Because of how links work now (that they aren't bi-directional), I had attempted to modify main.tf for this system test case to create one link from public to private. What ended up happening was I only needed to inspect the public environment's links, thus this variable was not needed. The terraform files are now wrong after my reversion because you have an environment link block that looks like this:

```
resource "layer0_environment_link" "public_private" {
  source = "${layer0_environment.el_public.id}"
  dest = "${layer0_environment.sl_private.id}"
}
```

But this structure is wrong.

I'm just going to reverse this removal for now but fixing this will require fixing the terraform files more fully.

tests/clients/sts_test_client.go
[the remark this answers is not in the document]
jparsons04: Oh, unless I use `s.T.Fatal()` perhaps?

tests/clients/sts_test_client.go
```diff
@@ -28,15 +27,14 @@ func NewSTSTestClient(t *testing.T, url string) *STSTestClient {
...
 
 func (s *STSTestClient) WaitForHealthy(timeout time.Duration) {
-	testutils.WaitFor(s.T, time.Second*10, timeout, func() bool {
-		logrus.Debugf("Waiting for sts service to be healthy")
+	log.Printf("[DEBUG] Waiting for sts service to be healthy")
+	for start := time.Now(); time.Since(start) < timeout; time.Sleep(time.Second * 10) {
 		if _, err := s.Client.GetHealth(); err != nil {
-			logrus.Debug(err)
-			return false
+			log.Println(err.Error())
```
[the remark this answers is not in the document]
jparsons04: Would this also need to use `s.T.Fatal()` if we're unable to check the health?

tests/stress/taskStress_test.go
```diff
@@ -35,38 +34,40 @@ func TestTaskStress(t *testing.T) {
 	}
 
 	for taskName, copies := range taskNameCopies {
-		go func(taskName string, copies int) {
-			log.Debugf("Creating task %s (copies: %d)", taskName, copies)
-			s.Layer0.CreateTask(taskName, environmentID, deployID, copies, nil)
-		}(taskName, copies)
+		go func(taskName string) {
+			log.Printf("[DEBUG] Creating task %s (copies: %d)", taskName, copies)
```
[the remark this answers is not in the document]
jparsons04: The problem though is that `CreateTaskRequest` doesn't accept Copies as a field anymore:
See here: https://github.com/quintilesims/layer0/blob/system-test-fixes-jparsons/tests/clients/layer0_test_client.go#L26
And here: https://github.com/quintilesims/layer0/blob/system-test-fixes-jparsons/common/models/create_task_request.go#L9

So, it doesn't seem like there is any current notion of `copies` that I could pass in here.

tests/system/environmentLink_test.go
jparsons04: Actually, I see more clearly why this is gone now. We need to call `s.Layer0.UpdateEnvironmentLink(publicEnvironmentID, links)` whereas before this was `s.Layer0.DeleteLink(publicEnvironmentID, privateEnvironmentID)`

So unless I change how this test is written to make two unidirectional links, this variable needs to stay out.

tests/system/import_test.go
```diff
@@ -19,16 +22,57 @@ func TestImport(t *testing.T) {
...
 
-	log.Debugf("Creating test resources")
-	environment := s.Layer0.CreateEnvironment("import")
-	loadBalancer := s.Layer0.CreateLoadBalancer("sts", environment.EnvironmentID)
-	deploy := s.Layer0.CreateDeploy("sts", data)
-	service := s.Layer0.CreateService("sts", environment.EnvironmentID, deploy.DeployID, loadBalancer.LoadBalancerID)
+	log.Printf("[DEBUG] Creating test resources")
+	createEnvironmentReq := models.CreateEnvironmentRequest{
+		EnvironmentName:  "import",
+		InstanceSize:     "m3.medium",
```
diemonster: [nit] I've been noticing that `t2.micro` instances start up faster, and they reduce the spend slightly. Seems more appropriate for using with tests, at least.

tests/system/main_test.go
```diff
@@ -46,7 +39,7 @@ func setup() {
 func teardown() {
 	if !*dry {
 		if err := filepath.Walk("cases", deleteStateFiles); err != nil {
-			fmt.Println("Error occurred during teardown: ", err)
+			log.Println("[ERROR] Error occurred during teardown: ", err)
```
diemonster: Where did we end up landing re: using the standard logger for `[ERROR]`? Still curious why we would log this versus just print to stdout. I'm thinking mainly of how this would best work in a CI pipeline some day (for example) and what would be most appropriate.

jparsons04: For messages involving the `T` object, I don't give it a log level. For things printed via `fmt` I use log levels. I'm not using `log.Errorf` for the reasons we discussed previously about "handling" error messages, opting to just print that an error has occurred and have something else actually do the graceful error handling.

tests/system/import_test.go
jparsons04: :white_check_mark:

tests/system/import_test.go
```diff
@@ -12,23 +15,71 @@ func TestImport(t *testing.T) {
 	t.Parallel()
 
 	s := NewSystemTest(t, "cases/import", nil)
+
+	// Don't actually run this test if dryrun is specified
+	// as it will first create resources then test imports
+	if s.Terraform.DryRun() {
```
zpatrick: not sure why this was added when that is already accessible via the `dry   = flag.Bool(...)`. 
Also, test should be skipped using the `t.Skip` function, e.g.:
```
if *dry {
    t.Skip("Test cannot execute during a dry run")
}
```

tests/clients/layer0_test_client.go
```diff
@@ -22,81 +25,114 @@ func NewLayer0TestClient(t *testing.T, endpoint, token string) *Layer0TestClient
 	}
 }
 
-func (l *Layer0TestClient) CreateTask(taskName, environmentID, deployID string, copies int, overrides []models.ContainerOverride) string {
-	jobID, err := l.Client.CreateTask(taskName, environmentID, deployID, copies, overrides)
+func (l *Layer0TestClient) CreateTask(req models.CreateTaskRequest) string {
+	jobID, err := l.Client.CreateTask(req)
```
zpatrick: [nit] this could be more DRY with a helper:
```go
func (l *Layer0TestClient) CreateTask(req models.CreateTaskRequest) string {
	return l.jobHelper(func() (string, error) {
		return l.Client.CreateTask(req)
	})
}

func (l *Layer0TestClient) jobHelper(fn func() (string, error)) string {
	jobID, err := fn()
	if err != nil {
		t.Fatal(err)
	}

	return job.Result
}
```

tests/clients/layer0_test_client.go
```diff
@@ -113,23 +149,20 @@ func (l *Layer0TestClient) ListTasks() []*models.TaskSummary {
 	return tasks
 }
 
-func (l *Layer0TestClient) ScaleService(id string, scale int) *models.Service {
-	service, err := l.Client.ScaleService(id, scale)
+func (l *Layer0TestClient) UpdateEnvironment(environmentID string, req models.UpdateEnvironmentRequest) string {
+	jobID, err := l.Client.UpdateEnvironment(environmentID, req)
```
zpatrick: Updates should probably behave like Creates

tests/clients/sts_test_client.go
```diff
@@ -29,14 +29,15 @@ func NewSTSTestClient(t *testing.T, url string) *STSTestClient {
...
 		if _, err := s.Client.GetHealth(); err != nil {
-			logrus.Debug(err)
-			return false
+			s.T.Fatalf("Error checking health: %v", err)
 		}
 
 		return true
 	})
+
+	s.T.Fatalf("Timeout reached after %v", timeout)
```
zpatrick: isn't this handled by testutils.WaitFor?

tests/clients/sts_test_client.go
```diff
@@ -29,14 +29,15 @@ func NewSTSTestClient(t *testing.T, url string) *STSTestClient {
 
 func (s *STSTestClient) WaitForHealthy(timeout time.Duration) {
 	testutils.WaitFor(s.T, time.Second*10, timeout, func() bool {
-		logrus.Debugf("Waiting for sts service to be healthy")
+		log.Printf("[DEBUG] Waiting for sts service to be healthy")
 		if _, err := s.Client.GetHealth(); err != nil {
-			logrus.Debug(err)
-			return false
+			s.T.Fatalf("Error checking health: %v", err)
```
zpatrick: You shouldn't fatal or error here, only log and return false

tests/stress/taskStress_test.go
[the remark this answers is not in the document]
zpatrick: `copies: %d` should be taken out of that log line

tests/system/environmentLink_test.go
zpatrick: sounds good, expect `s.Layer0.UpdateEnvironmentLink(publicEnvironmentID, links)` is not a thing that exists nor should exist. This is handled by the `UpateEnvironment` api call.

tests/system/environmentLink_test.go
```diff
@@ -30,34 +31,38 @@ func TestEnvironmentLink(t *testing.T) {
 	// curl the private service in the private environment from the public service in the public environment
 	// the private service returns "Hello, World!" from its root path
 	testutils.WaitFor(t, time.Second*10, time.Minute*5, func() bool {
-		log.Printf("Running curl while link exists")
+		log.Printf("[DEBUG] Running curl while link exists")
 		output, err := publicService.RunCommand("curl", "-m", "10", "-s", privateServiceURL)
 		if err != nil {
-			log.Printf("Error running curl: %v", err)
+			t.Fatalf("Error running curl: %v", err)
```
zpatrick: log, don't fatal here

tests/system/environmentLink_test.go
```diff
@@ -30,34 +31,38 @@ func TestEnvironmentLink(t *testing.T) {
...
 			return false
 		}
 
 		if expected := "Hello, World!"; output != expected {
-			log.Printf("Output from curl was '%s', expected '%s'", output, expected)
+			t.Fatalf("Output from curl was '%s', expected '%s'", output, expected)
```
zpatrick: ditto

tests/system/environmentLink_test.go
```diff
@@ -30,34 +31,38 @@ func TestEnvironmentLink(t *testing.T) {
...
+
+	s.Layer0.UpdateEnvironment(publicEnvironmentID, req)
 
 	testutils.WaitFor(t, time.Second*10, time.Minute*2, func() bool {
-		log.Println("Running curl without link")
+		log.Printf("[DEBUG] Running curl without link")
 		output, err := publicService.RunCommand("curl", "-m", "10", "-s", privateServiceURL)
 		if err != nil {
-			log.Printf("Error running curl: %v", err)
+			t.Fatalf("Error running curl: %v", err)
```
zpatrick: ditto

tests/system/environmentLink_test.go
```diff
@@ -30,34 +31,38 @@ func TestEnvironmentLink(t *testing.T) {
...
 			return false
 		}
 
 		if output != "" {
-			log.Printf("Output from curl was '%s', expected no output", output)
+			t.Fatalf("Output from curl was '%s', expected no output", output)
```
zpatrick: ditto

tests/system/import_test.go
```diff
@@ -12,23 +15,71 @@ func TestImport(t *testing.T) {
...
-	log.Debugf("Creating test resources")
-	environment := s.Layer0.CreateEnvironment("import")
-	loadBalancer := s.Layer0.CreateLoadBalancer("sts", environment.EnvironmentID)
-	deploy := s.Layer0.CreateDeploy("sts", data)
-	service := s.Layer0.CreateService("sts", environment.EnvironmentID, deploy.DeployID, loadBalancer.LoadBalancerID)
+	log.Printf("[DEBUG] Creating test resources")
+	createEnvironmentReq := models.CreateEnvironmentRequest{
+		EnvironmentName:  "import",
+		InstanceSize:     "t2.micro",
+		UserDataTemplate: nil,
```
zpatrick: Don't set fields if they are 0-values

tests/system/import_test.go
```diff
@@ -12,23 +15,71 @@ func TestImport(t *testing.T) {
...
+		UnhealthyThreshold: 2,
+	}
+
+	ports := []models.Port{{HostPort: 80, ContainerPort: 80, Protocol: "http"}}
+
+	createLoadBalancerReq := models.CreateLoadBalancerRequest{
+		LoadBalancerName: "sts",
+		EnvironmentID:    environmentID,
+		IsPublic:         true,
+		Ports:            ports,
```
zpatrick: why not `DefaultLoadBalancerPort` and `DefaultLoadBalancerHealthCheck`?

tests/system/import_test.go
```diff
@@ -12,23 +15,71 @@ func TestImport(t *testing.T) {
...
+		HealthCheck:      hc,
+	}
+
+	loadBalancerID := s.Layer0.CreateLoadBalancer(createLoadBalancerReq)
+
+	createDeployReq := models.CreateDeployRequest{
+		DeployName: "sts",
+		DeployFile: data,
```
zpatrick: [nit] it's odd the data variable is so far away from its actual usage here.

tests/system/serviceScale_test.go
```diff
@@ -18,18 +20,32 @@ func TestServiceScale(t *testing.T) {
 	defer s.Terraform.Destroy()
 
 	serviceID := s.Terraform.Output("service_id")
+	deployID := s.Terraform.Output("deploy_id")
+	scale := 3
 
-	s.Layer0.ScaleService(serviceID, 3)
+	req := models.UpdateServiceRequest{
+		DeployID: &deployID,
```
zpatrick: this is testing a new deployment and we only want to test scaling here; this  should be removed.

tests/system/serviceScale_test.go
```diff
@@ -18,18 +20,32 @@ func TestServiceScale(t *testing.T) {
...
+		log.Printf("[DEBUG] Waiting for service to scale up")
+		service := s.Layer0.ReadService(serviceID)
 		return service.RunningCount == 3
 	})
 
-	s.Layer0.ScaleService(serviceID, 1)
+	scale = 1
+
+	req = models.UpdateServiceRequest{
+		DeployID: &deployID,
```
zpatrick: ditto

tests/system/windowsService_test.go
```diff
@@ -26,22 +27,22 @@ func TestWindowsService(t *testing.T) {
...
 		return service.RunningCount == 1
 	})
 
 	testutils.WaitFor(t, time.Second*30, time.Minute*10, func() bool {
-		log.Debugf("Waiting for service to be healthy")
+		log.Printf("[DEBUG] Waiting for service to be healthy")
 		resp, err := http.Get(serviceURL)
 		if err != nil {
-			log.Debugf("There was an error checking the window service's url: %v", err)
+			t.Fatalf("There was an error checking the Windows service's URL: %v", err)
```
zpatrick: log, dont fatal or error

tests/system/windowsService_test.go
```diff
@@ -26,22 +27,22 @@ func TestWindowsService(t *testing.T) {
...
 			return false
 		}
 
 		defer resp.Body.Close()
 		if resp.StatusCode < 200 || resp.StatusCode > 299 {
-			log.Debugf("Windows service returned non-200 status: %d", resp.StatusCode)
+			t.Fatalf("Windows service returned non-200 status: %d", resp.StatusCode)
```
zpatrick: ditto

tests/stress/taskStress_test.go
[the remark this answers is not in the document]
jparsons04: Ah, yes, I will fix that.

tests/system/environmentLink_test.go
jparsons04: Yes, I intended to say `UpdateEnvironment` not `UpdateEnvironmentLink`

jparsons04: The code already reflects the usage of `UpdateEnvironment`

tests/system/import_test.go
jparsons04: Because we don't pass in the cli flag or context into each system test, I can't check on the CLI flag content within the test. So I can't directly test for the contents of `dry` or `flag`. This test is unique specifically because it creates resources first and then uses the `*SystemTest`'s tests.

jparsons04: The only way I've been able to see to infer that the dry flag has been specified is through `*SystemTest.Terraform.DryRun()`

tests/clients/layer0_test_client.go
jparsons04: The only similarity that I can see from the caller's perspective is that the function returns a `jobID`. Is there another similarity I'm overlooking?

tests/system/import_test.go
zpatrick: the dry flag is a global variable, so you can check on it form each system test.

tests/clients/layer0_test_client.go
zpatrick: it should probably wait for the job to complete like in create

tests/clients/layer0_test_client.go
```diff
@@ -2,11 +2,14 @@ package clients
...
 import (
 	"testing"
+	"time"
 
-	"github.com/quintilesims/layer0/cli/client"
+	"github.com/quintilesims/layer0/client"
 	"github.com/quintilesims/layer0/common/models"
 )
 
+var defaultTimeout = time.Minute * 5
```
zpatrick: This already exists in the `config` package. Also, fwiw, 5 minutes isn't long enough.

tests/stress/taskStress_test.go
```diff
@@ -34,39 +36,45 @@ func TestTaskStress(t *testing.T) {
 		"TaskJ": 1,
 	}
 
-	for taskName, copies := range taskNameCopies {
-		go func(taskName string, copies int) {
-			log.Debugf("Creating task %s (copies: %d)", taskName, copies)
-			s.Layer0.CreateTask(taskName, environmentID, deployID, copies, nil)
-		}(taskName, copies)
+	for taskName := range taskNameCopies {
```
zpatrick: `taskNameCopies` doesn't make sense anymore, this should be re-created to have the same behavior as it did before (100 tasks I think?)

tests/stress/taskStress_test.go
```diff
@@ -34,39 +36,45 @@ func TestTaskStress(t *testing.T) {
...
+		go func(taskName string) {
+			log.Printf("[DEBUG] Creating task %s", taskName)
+			req := models.CreateTaskRequest{
+				ContainerOverrides: nil,
```
zpatrick: [nit] don't declare 0-value fields

tests/system/environmentLink_test.go
```diff
@@ -16,11 +18,11 @@ func TestEnvironmentLink(t *testing.T) {
 	t.Parallel()
 
 	s := NewSystemTest(t, "cases/environment_link", nil)
+	s.Terraform.Init()
```
zpatrick: This should probably happen as part of `NewSystemTest` so callers don't accidentally forget.

tests/system/windowsService_test.go
```diff
@@ -19,29 +20,30 @@ func TestWindowsService(t *testing.T) {
 	t.Parallel()
 
 	s := NewSystemTest(t, "cases/windows_service", nil)
+	s.Terraform.Init()
```
zpatrick: [nit] ditto

tests/system/import_test.go
jparsons04: If I'm importing the flag package here, could I do something like:
```
     // Don't actually run this test if dryrun is specified
     // as it will first create resources then test imports
     if flag.Lookup("dry") {
         t.Skipf("Test cannot execute during a dry run")
     }
```

tests/stress/taskStress_test.go
jparsons04: Yes, but that will be part of a future PR that will be responsible for fixing the system tests.

zpatrick: 👍 sounds good

tests/clients/layer0_test_client.go
```diff
@@ -31,7 +29,7 @@ func (l *Layer0TestClient) jobHelper(fn func() (string, error)) string {
 		l.T.Fatal(err)
 	}
 
-	job, err := client.WaitForJob(l.Client, jobID, defaultTimeout)
+	job, err := client.WaitForJob(l.Client, jobID, config.DEFAULT_JOB_EXPIRY)
```
zpatrick: `config.DefaultTimeout`

tests/system/import_test.go
```diff
@@ -19,7 +20,7 @@ func TestImport(t *testing.T) {
 
 	// Don't actually run this test if dryrun is specified
 	// as it will first create resources then test imports
-	if s.Terraform.DryRun() {
+	if dry := flag.Lookup("dry"); dry != nil {
```
zpatrick: `if *dry { ... }` - you have access to package-wide variables

PR #440, opened 2017-12-08
Environment Scaler Refactor

tlake:
- Scaler no longer stops counting a provider as a provider if the agent is disconnected.
- Scaler selects the most efficient method of resource consumer allocation by comparing sort-by-memory and sort-by-CPU.
- Scaler aborts scaling actions if they would increase/decrease the number of instances above/below `max_count`/`min_count`.
- Removes/reduces spaghettification.
    - Logic of scaler is more cleanly laid out in the "master" `Scale` function.
    - Less nesting of functions.
- Updates `models.ResourceProvider` and `models.ResourceConsumer`
- Improves logging
- Removes several functions that did not seem to be used or useful.


common/models/resource_consumer.go
```diff
@@ -0,0 +1,10 @@
+package models
+
+import bytesize "github.com/zpatrick/go-bytesize"
+
+type ResourceConsumer struct {
```
diemonster: As stated on another PR (h/t @zpatrick) , these models are not used as part of API requests/responses, so it seems like having them in `models` is not the right place.

tlake: Oh, now I understand what you were getting at in that other comment @bivers. Is the idea, then, to just fold `ResourceConsumer` and `ResourceProvider` into `api/provider/aws/environment_scaler.go`?

zpatrick: It could be, but to me it seems that those things are entirely independent of the `provider/aws` package. Doing a scaling algorithmn based entirely off of `ResourceProvider` and `ResourceConsumer` seems to be a discrete area of code that could be moved/tested independently.

tlake:
Closing this PR, as it looks like the gitflow is moving over towards #441.

PR #445, opened 2017-12-13
Update Smoketests

zpatrick:
Updates our smoke tests to use the updated CLI flags and commands. 
I also ran `dep ensure` because we've merged some new dependencies in `api-refactor` without vendoring them. 

Run the smoke tests, manually scale up the environment's ASG for service and task tests. 
The windows portion of environment tests are expected to fail because the AMI is out of date.


tests/smoke/load_balancer.bats
```diff
@@ -70,14 +70,6 @@ certificate_name="l0-$LAYER0_INSTANCE-api"
     l0 loadbalancer create --port 80:80/http test loadbalancer4
 }
 
-@test "deploy create guestbook" {
```
diemonster: Did you remove these because they aren't loadbalancer-specific and/or there is overlap with `service` smoketests?

zpatrick:
@sesh-kebab: looks like a lot of the changes were pulled into [274](https://github.com/quintilesims/layer0/commit/78691264a37826fcce18f6c2dd9a0b49f9f02b37#diff-32c02121f94499c009b23d5237f101ca)

tests/smoke/load_balancer.bats
zpatrick: Both. This also allows the loadbalancer tests to run w/o the need for ec2 instances, taking the runtime from ~5 minutes to < 1.

zpatrick:
@diemonster 
> I also ran dep ensure because we've merged some new dependencies in api-refactor without vendoring them.

jparsons04:
I'm noticing on commit hash `267af807f4af97a6eb1674c7c7b6097aeb2762e2` that if I run `dep ensure` from the project base, a number of files are modified for me. Can you confirm this?

zpatrick:
@jparsons04 those changes are due to formatting. If you `dep ensure` and then format, there are no changes.

PR #446, opened 2017-12-14
Create API tags on startup

zpatrick:
Adds an `AdminProvider` object to the `provider/aws` package. 
This object has an `Init()` function that will add all of the necessary tags for the API entities. 

* Unit tests
* Manually remove all api-related tags from the Dynamo table, run the api locally, and the tags should get re-populated. All `l0 entity get` commands should succeed for `api` entities. 

**Checklist**
- [x] Unit tests

### DIFF
--- a/api/provider/aws/admin.go
+++ b/api/provider/aws/admin.go
@@ -1,0 +1,61 @@
+package aws
+
+import (
+	"log"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/quintilesims/layer0/api/tag"
+	awsc "github.com/quintilesims/layer0/common/aws"
+	"github.com/quintilesims/layer0/common/config"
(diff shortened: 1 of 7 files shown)


api/provider/aws/admin.go
```diff
@@ -0,0 +1,61 @@
...
+	if err != nil {
+		return err
+	}
+
+	// arn format: "arn:aws:ecs:region:123:task-definition/l0-instance-api:1"
+	taskDefinitionARN := aws.StringValue(service.TaskDefinition)
+	split := strings.Split(taskDefinitionARN, ":")
+	taskDefinitionRevision := split[len(split)-1]
+
+	tags := []models.Tag{
```
diemonster: How did we add these tags previously? I don't recall this existing in `admin`...

zpatrick: `api/logic/admin_logic`

PR #447, opened 2017-12-14
424: Modify L0 API ECS instance user data to send logs to CloudWatch

### LINKED ISSUE #424: Enable Docker debug output for ECS container instances and the Docker daemon
### Expected behavior

We do get some information from ECS container instances when we need to troubleshoot issues with our containers, but it could potentially be improved per AWS guidelines.

http://docs.aws.amazon.com/AmazonECS/latest/developerguide/docker-debug-mode.html

The above document describes how debug mode can be enabled for the Docker daemon which will output more verbose log messages from the daemon.

### Actual behavior

Some log information is sent to CloudWatch but it can be challenging at times to pinpoint the root cause of issues we have been experiencing with ECS containers.

jparsons04:
The initial user data script that is executed at l0 api instance launch time will now stream 6 logs into their own log streams for the api instance's CloudWatch Log Group:
* /var/log/docker
* /var/log/ecs/cloudwatch-logs-start.log
* /var/log/ecs/ecs-agent.log
* /var/log/ecs/ecs-init.log
* /var/log/messages
* /var/log/user-data.log

Create a new layer0 api instance and observe the new log streams created for your l0 instance.

**Note:** You need to have the fixes from issue #455 and #456 applied to run this successfully.

~**Checklist**~
~- [ ] Unit tests~

closes #424


zpatrick:
@jparsons04: Have you verified this works in AWS? If so, :shipit:

jparsons04:
It's actually not enough to simply add these lines to the user data bash script. We also have to do a number of additional things to pipe the container's system logs to CloudWatch. This article has the basic idea, but as it stands, the changes I've made don't actually do anything and require additional work.

http://docs.aws.amazon.com/AmazonECS/latest/developerguide/using_cloudwatch_logs.html
See the section **Configuring CloudWatch Logs at Launch with User Data**

api/provider/aws/defaults.go
```diff
@@ -20,6 +21,7 @@ Write-Host Cluster name set as: $clusterName -foreground green
 
 [Environment]::SetEnvironmentVariable("ECS_CLUSTER", $clusterName, "Machine")
 [Environment]::SetEnvironmentVariable("ECS_ENABLE_TASK_IAM_ROLE", "false", "Machine")
+[Environment]::SetEnvironmentVariable("ECS_LOGLEVEL", "debug", "Machine")
```
zpatrick: is this a typo? I would assume `ECS_LOG_LEVEL`

setup/module/api/user_data.sh
```diff
@@ -1,9 +1,98 @@
+Content-Type: multipart/mixed; boundary="==BOUNDARY=="
+MIME-Version: 1.0
+
+--==BOUNDARY==
 #!/bin/bash
+exec > >(tee /var/log/user-data.log|logger -t user-data -s 2>/dev/console) 2>&1
 echo ECS_CLUSTER=${cluster_id} >> /etc/ecs/ecs.config
 echo ECS_ENGINE_AUTH_TYPE=dockercfg >> /etc/ecs/ecs.config
-yum install -y aws-cli awslogs
+echo ECS_LOGLEVEL=debug >> /etc/ecs/ecs.config
```
zpatrick: 🤦‍♂️  windows...

api/provider/aws/defaults.go
jparsons04: No. Per the [docs](https://github.com/aws/amazon-ecs-agent/blob/master/README.md#environment-variables) under the Environment Variables section, it is `ECS_LOGLEVEL`

setup/module/api/user_data.sh
jparsons04: This is for the L0 API module specifically which is always going to be a Linux ECS container. I think I want to do something separate for Windows and Linux environments that are created.

sesh-kebab:
As long as DEBUG is a good default among
`crit | error | warn | info | debug` then this is good 👍

setup/module/api/user_data.sh
zpatrick: 🤦‍♂️ linux...

jparsons04:
After discussion, I think I'm going to turn down the logging levels to INFO for both the docker daemon and the ECS agent.

zpatrick:
closing to reduce clutter

Issue #444, opened 2017-12-13
failure to push remote tf state when not using default region


davemeyer:
### Expected behavior
tf state should be pushed to s3 bucket in the region specified in l0-setup init

### Actual behavior
the state is pushed to a nonexistent bucket in the default refion (us-west-2), so no remote state is ever saved.

[INFO] Pushing /Users/dmeyer/.layer0/hexeast1/dockercfg.json to s3://layer0-hexeast1-236817160978/terraform/dockercfg.json
BucketRegionError: incorrect region, the bucket is not in 'us-west-2' region

v0.10.3

jparsons04:
Do you happen to have the environment variable `LAYER0_AWS_REGION` set in your environment? I ask because `l0-setup push` invokes `newAWSProviderHelper` which in turn checks for the cli flag `aws-region`.

```12:52:08 jparsons@seawl000002 {(v0.10.3)} ~/dev/go/src/github.com/quintilesims/layer0/setup $ ./l0-setup push --help
NAME:
   l0-setup push - push a Layer0 instance configuration to S3

USAGE:
   l0-setup push [command options] NAME

OPTIONS:
   --aws-access-key value  access key portion of an AWS key [$LAYER0_AWS_ACCESS_KEY_ID]
   --aws-secret-key value  secret key portion on an AWS key [$LAYER0_AWS_SECRET_ACCESS_KEY]
   --aws-region value      AWS region [$LAYER0_AWS_REGION]
```
If you aren't passing in `--aws-region` as a flag for the command `l0-setup push` and you don't have the environment variable `LAYER0_AWS_REGION` set, then it will default to us-west-2:
https://github.com/quintilesims/layer0/blob/v0.10.3/setup/command/aws.go#L58

jparsons04:
The PR that I just merged into `develop` would require `LAYER0_AWS_REGION` to be set in the environment. If you don't have this and run `l0-setup` then you need to pass in the `--aws-region` flag. There is no default region any longer, and it needs to be explicitly provided.

In my testing, I used both the environment variable and cli flag, and created and applied a new layer0 instance in us-east-1 and was able to get the terraform state pushed to the S3 bucket housed in us-east-1 without issue.

jparsons04:
Just want to leave a comment with a summary of our upcoming changes:

1. The l0-setup cli flag `--aws-region` has been removed as an option.
[...]

PR #448, opened 2017-12-14
Enable aws_ami data source for l0-setup

diemonster:
Enables l0-setup to use terraform's[ aws_ami data source](https://www.terraform.io/docs/providers/aws/d/ami.html#description).

Test by running `l0-setup init` and `apply`. The correct AMI for your region should be interpolated. 

Testing Caveat: you'll need to select the correct branch of the layer0 terraform module when running l0-setup `terraform_ami_update`. Unfortunately, selecting this branch as the module source when prompted by `l0-setup` seems to be broken. I was able to get around this by having `terraform get` pull down `master`, and then I manually switched branches in `~/.layer0/.terraform/...`.

**Checklist**
- [x] Manual testing


setup/module/api/core.tf
```diff
@@ -104,3 +104,31 @@ resource "aws_iam_group_policy" "mod" {
   group  = "${aws_iam_group.mod.id}"
   policy = "${element(data.template_file.group_policy.*.rendered, count.index)}"
 }
+
+data "aws_ami" "linux" {
+  most_recent = true
+  executable_users = ["self"]
```
zpatrick: this this necessary? https://www.terraform.io/docs/providers/aws/d/ami.html

setup/module/api/core.tf
```diff
@@ -104,3 +104,31 @@ resource "aws_iam_group_policy" "mod" {
...
+
+  filter {
+    name = "name"
+    values = ["amzn-ami-2017.09.d-amazon-ecs-optimized"]
+  }
+}
+
+data "aws_ami" "windows" {
+  most_recent = true
+  executable_users = ["self"]
```
zpatrick: ditto?

setup/module/api/core.tf
diemonster: I don't think it's necessary, but as the docs say:
"Limit search to users with explicit launch permission on the image."

It's an untested use case, but in the event that an image got published that we didn't have launch permissions on, this _could_ come in handy.

zpatrick: Ah I see. I'm not 100% what "explicit launch permission" entails, but I'm worried about the inverse problem: no one has "explicit launch permission" on AMIs w/o going in and setting some permission yourself. Anyways, this just needs to be tested.

jparsons04:
Running `l0-setup init` and `l0-setup apply` works for me as described. :white_check_mark:

jparsons04:
I am however having an issue when I run `l0-setup init` and specify a different region in the wizard:

https://gist.github.com/jparsons04/d179c10a7f734d71ce7ad01b8a06f5ba

diemonster:
@jparsons04 is the AMI you got back for `us-east-1` a different AMI than what you got in `us-west-2`?

jparsons04:
`l0-setup plan` shows me a different ami id when I build it in us-west-2:

```  + module.layer0.module.api.aws_launch_configuration.api
      id:                                       <computed>
      associate_public_ip_address:              "false"
      ebs_block_device.#:                       <computed>
      ebs_optimized:                            <computed>
      enable_monitoring:                        "true"
      iam_instance_profile:                     "${aws_iam_instance_profile.ecs.id}"
      image_id:                                 "ami-62d35c02"
      instance_type:                            "t2.small"
      key_name:                                 "xfra-dev"
      name:                                     <computed>
      name_prefix:                              "l0-jlpamitest-api-"
      root_block_device.#:                      <computed>
      security_groups.#:                        <computed>
      user_data:                                "656bca31b5f7cea3181e684bfea743272dcaefca"
```

setup/module/api/core.tf
diemonster: I removed this param for now as the matter of permissions for images should be out of ours hands, otherwise it somewhat defeats the purpose of this PR.

diemonster:
@jparsons04 at this moment in time, `ami-fad25980` should be the image returned for `us-east-1`. Very odd that it's returning a different ami.

diemonster:
I believe this is related to https://github.com/quintilesims/layer0/issues/444, in that we should be using terraform's data sources when looking up availability zones, etc. In `us-west-1`, our hardcoded zones are incorrect: 

```
* aws_subnet.private.1: Error creating subnet: InvalidParameterValue: Value (us-west-1b) for parameter availabilityZone is invalid. Subnets can currently only be created in the following availability zones: us-west-1a, us-west-1c.
	status code: 400, request id: ec73ebcd-16cb-4aad-9c9a-778f74adeef9
* module.layer0.module.vpc.aws_subnet.public[1]: 1 error(s) occurred:
```

diemonster:
@jparsons04 I think we should merge this PR, and solve the overall issue in https://github.com/quintilesims/layer0/issues/452. We can confirm that this PR works for `us-west-2` at least, which is already current behavior.

jparsons04:
I'm fine with that approach.

diemonster:
@jparsons04 I think the other problem is that you're sourcing the Layer0 terraform module from `github.com/quintilesims/layer0//setup/module` which will get the code from `master`, not the changes we have here. I just tried a test run using `github.com/quintilesims/layer0//setup/module`, and `l0-setup` ended up using our previously configured AMI from 2016.

PR #454, opened 2017-12-18
Use AWS AZ data source when creating Layer0 API subnets

### LINKED ISSUE #452: Configure l0-setup to use supported AWS regions
`l0-setup` successfully creates resources in `us-west-2` without an issue, but in other regions, such as `us-west-1` and `us-east-1`, it fails. We should ensure that `l0-setup` can provision in any of our supported regions.

First thought is that we should be using terraform's data sources for more things than we are currently.

sesh-kebab:
Updates the API's VPC module to rely on Availability data source rather than hardcoded values of a,b & c. This allows a layer0 instance to be created in any regions that don't have the expected AZs.

Create a new Layer0 Instance for a region that has more/less than the standard a, b & c zones like `us-west-1`. The creation via l0-setup should succeed. 

**Checklist**
- ~[ ] Unit tests~

closes #452

### DIFF
--- a/setup/module/vpc/main.tf
+++ b/setup/module/vpc/main.tf
@@ -1,3 +1,5 @@
+data "aws_availability_zones" "available" {}
+
 resource "aws_vpc" "mod" {
   count = "${var.count_hack}"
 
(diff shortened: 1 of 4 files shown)


setup/module/vpc/main.tf
```diff
@@ -46,18 +48,18 @@ resource "aws_route_table" "private" {
...
+  tags              = "${merge(var.tags, map("Tier", "Private"), map("Name", format("l0-%s-subnet-private-%s", var.name, element(data.aws_availability_zones.available.names, count.index))))}"
 }
 
 resource "aws_subnet" "public" {
   vpc_id            = "${aws_vpc.mod.id}"
-  cidr_block        = "${var.public_subnets[count.index]}"
-  availability_zone = "${element(var.azs, count.index)}"
-  count             = "${length(var.public_subnets) * var.count_hack}"
-  tags              = "${merge(var.tags, map("Tier", "Public"), map("Name", format("l0-%s-subnet-public-%s", var.name, element(var.azs, count.index))))}"
+  cidr_block        = "${cidrsubnet(aws_vpc.mod.cidr_block, 8, count.index + 1 + 100)}"
```
sesh-kebab: [nit] This is intentional. I could do this to make the intention clearer (count.index + 1)

Issue #455, opened 2017-12-19
SSH Key pair needs to be defined in task definition of layer0 api module


jparsons04:
For `api-refactor` and its descendants

### Expected behavior

`l0-setup` should be able to stand up a new layer0 API instance

### Actual behavior

When the api service's task attempts to run, the CloudWatch logs indicated that there is a missing variable: `ssh-key-pair`.

This is because there is a missing line in `setup/module/api/Dockerrun.aws.json`:
`{ "name": "LAYER0_AWS_SSH_KEY_PAIR", "value": "${ssh_key_pair}" }`

### Steps to reproduce the behavior

`l0-setup init`
`l0-setup apply`

Note that the tasks running for the api service never run

jparsons04:
Fix merged.

PR #458, opened 2017-12-20
456: Allow the Layer0 API Health check to not require basic auth

### LINKED ISSUE #456: Layer0 API health check HTTP:80/admin/health not allowing EC2 instances to attach to ELB
For `api-refactor` and its descendants

### Expected behavior

When standing up an new layer0 API, the ECS container instances running the API service should be attaching to an ELB and the health check HTTP:80/admin/health should be returning an HTTP 200 response.

### Actual behavior

Instead, we observe that the EC2 instances get attached to the API's ELB, then detach after the health check fails.

Stand up an new Layer0 instance and then observe the API service's ELB through the AWS console and note the instance list over time.

jparsons04:
The admin endpoint /admin/health should not require basic auth.
The way the fireball controller is coded, even if the ELB health check ping target is changed to TCP:port, the controller will still return an HTTP 401 response. 

This change removes /admin/health and instead creates a new endpoint, /health that will not require basic auth.

Create a new layer0 instance (as of this writing, include the fix in issue #455). When `l0-setup apply` is run, the instance should come up healthy and the apply should finish without issue.

**Checklist**
~- [ ] Unit tests~

closes #456

### DIFF
--- a/api/controllers/health.go
+++ b/api/controllers/health.go
@@ -1,33 +1,26 @@
 package controllers
 
 import (
-	"github.com/quintilesims/layer0/common/config"
 	"github.com/zpatrick/fireball"
 )
 
-type HealthController struct {
-	Config  config.APIConfig
-	Version string
+type HealthController struct{}
+
+func NewHealthController() *HealthController {
+	return &HealthController{}
 }
 
-func NewHealthController(c config.APIConfig, version string) *HealthController {
-	return &HealthController{
-		Config:  c,
-		Version: version,
-	}
-}
-
-func (a *HealthController) Routes() []*fireball.Route {
+func (h *HealthController) Routes() []*fireball.Route {
 	return []*fireball.Route{
 		{
 			Path: "/health",
 			Handlers: fireball.Handlers{
-				"GET": a.GetHealth,
+				"GET": h.GetHealth,
 			},
 		},
 	}
 }
 
-func (a *HealthController) GetHealth(c *fireball.Context) (fireball.Response, error) {
+func (h *HealthController) GetHealth(c *fireball.Context) (fireball.Response, error) {
 	return fireball.NewJSONResponse(200, "")
 }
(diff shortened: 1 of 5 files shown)


api/main.go
```diff
@@ -105,6 +104,15 @@ func main() {
 		routes = append(routes, controllers.NewTagController(tagStore).Routes()...)
 		routes = append(routes, controllers.NewTaskController(taskProvider, jobStore, tagStore).Routes()...)
 
+		adminRoutes := controllers.NewAdminController(cfg, Version).Routes()
```
zpatrick: This can be simplified:  just make a new endpoint, `/health` with a different controller; there is no need for the health check to be tied to the `admin` controller. Then, `append` the health route to `routes` after the routes have been decorated.

jparsons04: Ah! Sure thing. That should be straightforward enough.

api/controllers/health.go
```diff
@@ -0,0 +1,33 @@
+package controllers
+
+import (
+	"github.com/quintilesims/layer0/common/config"
+	"github.com/zpatrick/fireball"
+)
+
+type HealthController struct {
+	Config  config.APIConfig
```
zpatrick: why does it need either of these?

api/main.go
```diff
@@ -114,6 +114,8 @@ func main() {
 			fireball.LogDecorator(),
 			fireball.BasicAuthDecorator(user, pass))
 
+		routes = append(routes, controllers.NewHealthController(cfg, Version).Routes()...)
```
zpatrick: this should probably have a comment above, like `// don't decorate health endpoint with auth`

api/controllers/health.go
jparsons04: :thinking: I suppose given that the API doesn't have any sort of cli flag or environment variable that corresponds with this function, we wouldn't need to have this. If the API changes in some way that would add a cli flag or env var for this controller, then I would suppose we would need these.

I'll remove them.

api/main.go
jparsons04: :white_check_mark:

PR #468, opened 2018-01-02
[v0.10.4] 466: Implement retry and delay logic against AWS API

### LINKED ISSUE #466: [v0.10.4] Implement retry and delay logic against AWS API
### Expected behavior
In reference to issue #423. This adds retry and delay configuration options to help with rate limiting issues that have previously been encountered.

### Actual behavior

The code does not have a configurable way to incorporate rate limiting against the AWS API and instead uses defaults.

jparsons04:
Recognizes the environment variable LAYER0_AWS_TIME_BETWEEN_REQUESTS.

LAYER0_AWS_TIME_BETWEEN_REQUESTS is a time duration for the min amount
of time allowed between AWS requests

Max retry constant is also introduced and set to 999.

Unit tests for job and tag store were modified to include these new configurable values. Unit tests should pass.

**Checklist**
- [x] Unit tests

closes #466


jparsons04:
Gist of manual local API testing where time between requests is set to 1.5 seconds and where I printed ticker invocations to the terminal: https://gist.github.com/jparsons04/104efa2a35bef07da95f397bef3c3bb6

common/config/config.go
```diff
@@ -214,6 +218,14 @@ func TestDynamoJobTableName() string {
 	return get(TEST_AWS_JOB_DYNAMO_TABLE)
 }
 
+func AWSTimeBetweenRequests() string {
+	return getOr(AWS_TIME_BETWEEN_REQUESTS, DEFAULT_TIME_BETWEEN_REQUESTS)
+}
+
+func AWSMaxRetries() string {
```
zpatrick: I don't think we need to expose this; we should just use a const of 999

common/db/job_store/dynamo_test.go
```diff
@@ -24,7 +27,23 @@ func NewTestJobStore(t *testing.T) *DynamoJobStore {
 		Region:      aws.String(config.AWSRegion()),
 	}
 
+	maxRetries, err := strconv.Atoi(config.AWSMaxRetries())
```
zpatrick: we don't need this for a test

common/db/tag_store/dynamo_test.go
```diff
@@ -48,7 +51,23 @@ func NewTestTagStore(t *testing.T) *DynamoTagStore {
 		Region:      aws.String(config.AWSRegion()),
 	}
 
+	maxRetries, err := strconv.Atoi(config.AWSMaxRetries())
```
zpatrick: ditto

common/aws/provider/connection.go
```diff
@@ -75,7 +79,24 @@ var getConfig = func(credProvider CredProvider, region string) (sess *session.Se
 		Credentials: creds,
 		Region:      aws.String(region),
 	}
+
+	maxRetries, err := strconv.Atoi(config.AWSMaxRetries())
```
zpatrick: in theory, this logic should only need to exist in 1 location; I'm seeing a lot of copies everywhere.

common/config/config.go
```diff
@@ -38,15 +38,18 @@ const (
...
 // bGF5ZXIwOm5vaGF4cGx6 = layer0:nohaxplz, base64 encoded (basic http auth)
 const (
-	DEFAULT_AUTH_TOKEN   = "bGF5ZXIwOm5vaGF4cGx6"
-	DEFAULT_API_ENDPOINT = "http://localhost:9090/"
-	DEFAULT_API_PORT     = "9090"
-	DEFAULT_AWS_REGION   = "us-west-2"
+	DEFAULT_AUTH_TOKEN            = "bGF5ZXIwOm5vaGF4cGx6"
+	DEFAULT_API_ENDPOINT          = "http://localhost:9090/"
+	DEFAULT_API_PORT              = "9090"
+	DEFAULT_AWS_REGION            = "us-west-2"
```
diemonster: how does this work when the user specifies a different region as part of `l0-setup`?

diemonster: (i.e. does the region always get set properly as an env var)

jparsons04: It is either passed in as a CLI flag at setup time or uses the default:

https://github.com/quintilesims/layer0/blob/466-api-retry-delay-logic-jparsons/setup/command/aws.go#L59

jparsons04: Additionally, if `LAYER0_AWS_REGION` is set as an environment variable, then the `AWSRegion()` function will pull that

https://github.com/quintilesims/layer0/blob/466-api-retry-delay-logic-jparsons/common/config/config.go#L156

diemonster: right, but this is problematic since `l0-setup endpoint` (as well as `-d`) does not include the aws region in its' list of exports. We need to add that otherwise the default region of `us-west-2` will always be used.

common/startup/startup.go
```diff
@@ -126,12 +128,17 @@ func GetLogic(backend *ecsbackend.ECSBackend) (*logic.Logic, error) {
 
 func getNewTagStore() (tag_store.TagStore, error) {
 	creds := credentials.NewStaticCredentials(config.AWSAccessKey(), config.AWSSecretKey(), "")
-	awsConfig := &aws.Config{
-		Credentials: creds,
-		Region:      aws.String(config.AWSRegion()),
+	session := session.New(config.GetAWSConfig(creds, config.AWSRegion()))
```
zpatrick: shouldn't this use the new helper?

common/startup/startup.go
```diff
@@ -143,12 +150,17 @@ func getNewTagStore() (tag_store.TagStore, error) {
...
+	delay, err := time.ParseDuration(config.AWSTimeBetweenRequests())
+	if err != nil {
+		return nil, err
 	}
 
-	session := session.New(awsConfig)
+	ticker := time.Tick(delay)
+	session.Handlers.Send.PushBack(func(r *request.Request) {
+		<-ticker
```
zpatrick: ditto

common/startup/startup.go
jparsons04: The helper is `config.GetAWSConfig()`, are you suggesting that I should make the helper get the entire Session object as opposed to getting the AWS configuration?

zpatrick: I'm suggesting that this block of logic should not be copy+pasted; it should only exist in a single location of our codebase. How you accomplish that is up to you.

jparsons04: :white_check_mark: I implemented a helper function inside this file to handle the redundant logic.

common/startup/startup.go
```diff
@@ -170,6 +158,20 @@ func getNewJobStore() (job_store.JobStore, error) {
 	return store, nil
 }
 
+func handleSessionPushBack(session *session.Session) error {
```
diemonster: [nit] not a big fan of this helper's name, feels like we're using AWS terminology when we don't have to. What is this function actually doing in simpler terms?

jparsons04: It's what actually enforces the time delay between each AWS request.

diemonster: then my minor suggestion would be something like `requestTimeDelay()` or `sessionTimeDelay()`

jparsons04: :white_check_mark: That's reasonable.

Issue #463, opened 2018-01-02
[v0.10.4] l0-setup AMI, region lookup


diemonster:
Enable v0.10.4 to use l0-setup AM and regionI lookup ala code in `api-refactor`

PR #471, opened 2018-01-02
update terraform module to use region, ami data sources

diemonster:
Configures Layer0 module and terraform to use data sources for AMIs and regions.  

Test by running l0-setup init and apply. The correct AMI for your region should be interpolated.

Testing Caveat: you'll need to select the correct branch of the layer0 terraform module when running l0-setup terraform_ami_update. Unfortunately, selecting this branch as the module source when prompted by l0-setup seems to be broken. I was able to get around this by having terraform get pull down master, and then I manually switched branches in ~/.layer0/.terraform/....

**Checklist**
- [X] Manual testing

closes #
https://github.com/quintilesims/layer0/issues/463


setup/module/api/core.tf
```diff
@@ -104,3 +104,29 @@ resource "aws_iam_group_policy" "mod" {
...
+  filter {
+    name   = "owner-alias"
+    values = ["amazon"]
+  }
+
+  filter {
+    name = "name"
+    values = ["Windows_Server-2016-English-Full-ECS_Optimized-2017.11.24"]
+  }
+}
```
zpatrick: EOF character?

diemonster: there's a couple EOF characters in this PR, but I don't recall explicitly adding them. Might be my editor doing that automatically. Regardless, it didn't impact my testing.

setup/instance/inputs.go
```diff
@@ -56,12 +56,12 @@ Layer0 API will use its own key with limited permissions to provision AWS resour
 
 const INPUT_AWS_REGION_DESCRIPTION = `
 AWS Region: The region input variable specifies which region to provision the
-AWS resources required for Layer0. The following regions can be used: 
+AWS resources required for Layer0. Regions that support the following services can be used:
```
jparsons04: This sentence is grammatically awkward and it makes the prompt non-intuitive. It almost makes it look like you want the user to enter a service, not a region.

diemonster: Which sentence is grammatically awkward... I'm assuming the second one? This is copy & paste from @sesh-kebab 's previously merged patch. I do agree that if you're not reading it carefully, you might think one of the bulleted items should be an option for this prompt. However, I don't think we should hardcode regions that have the services listed available here.

jparsons04: Or maybe another possible sentence is "Regions where the following services are allowed may be entered".

jparsons04: Still, the form of this input prompt differs slightly from other prompts to a degree where you're listing services underneath and someone squinting might naturally think they have to enter a service.

diemonster: I would prefer to table this for now and see if anyone else gets tripped up on it during testing. To me, the problem lies in that if you separate out those services in a list, it will seem like maybe they should be entered into the prompt here. But if you don't put those services into a list, it makes for a longer, weirder sentence that would likely be ignored.

PR #472, opened 2018-01-02
[v0.10.4] scaler patch

### LINKED ISSUE #461: [v.0.10.4] Fix disconnected ECS agent scaling issue
There is no reason to actually be checking for the connectivity of the ECS agent on an instance when allocating resource consumers onto resource providers.

tlake:
Removes a check for the ECS agent being connected when gathering resource providers; an instance should still be considered to exist even when the agent happens to be disconnected.

Run unit tests. Manually perform and monitor scaling operations.

**Checklist**
- [x] Unit tests

closes #461

### DIFF
--- a/api/backend/ecs/resource_manager.go
+++ b/api/backend/ecs/resource_manager.go
@@ -61,6 +61,10 @@ func (r *ECSResourceManager) getResourceProvider(ecsEnvironmentID id.ECSEnvironm
 		return nil, false
 	}
 
+	if !pbool(instance.AgentConnected) {
+		r.logger.Infof("Instance %s agent is disconnected", instanceID)
+	}
+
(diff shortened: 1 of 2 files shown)


api/backend/ecs/resource_manager.go
```diff
@@ -61,11 +61,6 @@ func (r *ECSResourceManager) getResourceProvider(ecsEnvironmentID id.ECSEnvironm
 		return nil, false
 	}
 
-	if !pbool(instance.AgentConnected) {
-		r.logger.Errorf("Instance %s agent is disconnected", instanceID)
```
zpatrick: we should still log

tlake: Okeydoke. I've got it at `Info` level right now - maybe `Debug` level is more appropriate...?

PR #473, opened 2018-01-03
Update list service to reduce describe ecs calls.

nehayward:
This is a patch fix to list services to reduce number of api calls.

**Checklist**
- [x] Unit tests


api/backend/ecs/service_manager.go
```diff
@@ -57,6 +57,21 @@ func (this *ECSServiceManager) ListServices() ([]*models.Service, error) {
 	return services, nil
 }
 
+func (this *ECSServiceManager) ListServiceNames() ([]id.L0ServiceID, error) {
```
zpatrick: You should update `ListServices` instead of adding this new function in this instance

api/backend/ecs/service_manager.go
```diff
@@ -57,6 +57,21 @@ func (this *ECSServiceManager) ListServices() ([]*models.Service, error) {
...
+	serviceARNs, err := this.ECS.Helper_ListServices(id.PREFIX)
```
zpatrick: Drop any function thats `Helper_...`; just use your own that we are sure is optimal

api/backend/ecs/service_manager.go
```diff
@@ -57,6 +57,21 @@ func (this *ECSServiceManager) ListServices() ([]*models.Service, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	servicesID := make([]id.L0ServiceID, len(serviceARNs))
+	for i, arn := range serviceARNs {
+		ecsServiceID := id.ServiceARNToECSServiceID(*arn)
```
zpatrick: the `ecs` package should return `[]string` of service arns; no need to deal with pointer ugliness here

api/backend/ecs/task_manager.go
```diff
@@ -160,14 +159,6 @@ func modelFromTasks(tasks []*ecs.Task) (*models.Task, error) {
 
 			details = append(details, detail)
 		}
-
-		copy := models.TaskCopy{
```
zpatrick: this is not meant to be removed

zpatrick: undo any changes you've made to this file

api/backend/interface.go
```diff
@@ -20,6 +20,7 @@ type Backend interface {
 	DeleteDeploy(deployID string) error
 
 	ListServices() ([]*models.Service, error)
+	ListServiceNames() ([]id.L0ServiceID, error)
```
zpatrick: drop

api/logic/environment_resource_getter.go
```diff
@@ -151,17 +151,12 @@ func (c *EnvironmentResourceGetter) getPendingTaskResourcesInJobs(environmentID
...
 				if req.EnvironmentID == environmentID {
-					// note that this isn't exact if the job has started some, but not all of the tasks
-					deployIDCopies := map[string]int{
-						req.DeployID: int(req.Copies),
-					}
-
 					// resource consumer ids are just used for debugging purposes
-					generateID := func(deployID, containerName string, copy int) string {
-						return fmt.Sprintf("Task: %s, Deploy: %s, Container: %s, Copy: %d", req.TaskName, deployID, containerName, copy)
+					generateID := func(deployID, containerName string) string {
```
zpatrick: not sure why any of this is changing? copy is fine even if it's always 1. I feel like this is introducing new bugs, would rather we not touch this unless we have to.

PR #474, opened 2018-01-04
[v0.10.4] Decrease ECS API Calls

### LINKED ISSUE #462: [v0.10.4] Decrease number of ECS API calls
We need to reduce the number of ECS API calls for the 10.4 release. The biggest culprit is the list/describe task API calls. We should also investigate how to reduce the number of list/describe calls for services, clusters, and task definitions (if possible).

zpatrick:
This PR decreases the total number of ECS API calls by using the following methods:
* It uses `arn` tags to map Layer0 Tasks to ECS Tasks 
* It optimizes List and Read calls for environment, service, and task
* It changes the scaler to run every hour instead of every 5 minutes
* It removes the `copies` param from create task
* It add functions `GetEnvironmentServices()` and `GetEnvironmentTasks()` to the logic layer. 
This is used by the scaler, which was calling get service/task once for each service/task in an environment. 

Unit, Smoke, System, and Stress tests

**Checklist**
- [ ] Unit tests
~[ ] Smoke tests (if applicable)~
~[ ] System tests (if applicable)~
~[ ] Documentation (if applicable)~

closes #462


api/backend/ecs/task_manager.go
```diff
@@ -171,13 +152,9 @@ func modelFromTasks(tasks []*ecs.Task) (*models.Task, error) {
...
-		EnvironmentID: id.ClusterARNToECSEnvironmentID(*tasks[0].ClusterArn).L0EnvironmentID(),
-		PendingCount:  pendingCount,
-		RunningCount:  runningCount,
-		DesiredCount:  int64(len(tasks)),
-		TaskID:        id.ECSTaskID(*tasks[0].StartedBy).L0TaskID(),
-		Copies:        copies,
-		DeployID:      id.TaskDefinitionARNToECSDeployID(*tasks[0].TaskDefinitionArn).L0DeployID(),
+		RunningCount: runningCount,
+		PendingCount: pendingCount,
+		Copies:       copies,
```
zpatrick: This seems to be add odds with how we removed `copies` as an argument, but for clarity, this is/was a slice of details about each copy of the task that was running. We still want to keep these details around, but since we only allow 1 copy of a task, this object is still a slice but will just always have a size of 1.

api/logic/task_logic.go
```diff
@@ -33,12 +37,17 @@ func (this *L0TaskLogic) ListTasks() ([]*models.TaskSummary, error) {
 }
 
 func (this *L0TaskLogic) GetTask(taskID string) (*models.Task, error) {
```
tlake: Possibly out of the scope of this PR, but we should be moving away from `this` when we can.

api/backend/ecs/service_manager.go
```diff
@@ -40,21 +41,25 @@ func NewECSServiceManager(
 	}
 }
 
-func (this *ECSServiceManager) ListServices() ([]*models.Service, error) {
-	serviceARNs, err := this.ECS.Helper_ListServices(id.PREFIX)
+func (this *ECSServiceManager) ListServices() ([]id.ECSServiceID, error) {
```
diemonster: would it be worth renaming this to `ListServiceIDS()`?

api/backend/ecs/task_manager.go
```diff
@@ -50,88 +52,66 @@ func (this *ECSTaskManager) ListTasks() ([]string, error) {
...
-	reason := "Task stopped by User"
-
-	for _, taskARN := range taskARNs {
-		if err := this.ECS.StopTask(ecsEnvironmentID.String(), reason, *taskARN); err != nil {
-			return err
-		}
-	}
-
-	return nil
+	return this.ECS.StopTask(ecsEnvironmentID.String(), taskARN, StopTaskReason)
```
diemonster: why don't we have to loop through `taskARNs` anymore?

api/backend/ecs/task_manager.go
diemonster: 👍

api/backend/ecs/task_manager.go
```diff
@@ -50,88 +52,66 @@ func (this *ECSTaskManager) ListTasks() ([]string, error) {
...
-
 	ecsOverrides := []*ecs.ContainerOverride{}
 	for _, override := range overrides {
 		o := ecs.NewContainerOverride(override.ContainerName, override.EnvironmentOverrides)
 		ecsOverrides = append(ecsOverrides, o)
 	}
 
 	startedBy := id.PREFIX
-	tasks, failed, err := this.ECS.RunTask(ecsEnvironmentID.String(), ecsDeployID.TaskDefinition(), 1, stringp(startedBy), ecsOverrides)
+	task, err := this.ECS.RunTask(ecsEnvironmentID.String(), ecsDeployID.TaskDefinition(), startedBy, ecsOverrides)
```
diemonster: was this an SDK change that `failed` is no longer returned?

api/backend/ecs/task_manager.go
```diff
@@ -171,13 +152,9 @@ func modelFromTasks(tasks []*ecs.Task) (*models.Task, error) {
...
 
 	model := &models.Task{
-		EnvironmentID: id.ClusterARNToECSEnvironmentID(*tasks[0].ClusterArn).L0EnvironmentID(),
-		PendingCount:  pendingCount,
-		RunningCount:  runningCount,
-		DesiredCount:  int64(len(tasks)),
-		TaskID:        id.ECSTaskID(*tasks[0].StartedBy).L0TaskID(),
-		Copies:        copies,
-		DeployID:      id.TaskDefinitionARNToECSDeployID(*tasks[0].TaskDefinitionArn).L0DeployID(),
+		RunningCount: runningCount,
```
diemonster: Why don't we want `EnvironmentID` in this model?

tests/smoke/task.bats
```diff
@@ -12,8 +12,8 @@
     l0 task create --wait test task1 alpine:latest
 }
 
-@test "task create --copies 2 --env alpine:key=val test task2 alpine:latest" {
-    l0 task create --copies 2 --env alpine:key=val test task2 alpine:latest
+@test "task create --env alpine:key=val test task2 alpine:latest" {
```
diemonster: OK, yes, so this would be a breaking change. Not the end of the world, but we need to have a good explanation as for why we wouldn't want `copies` as a param any more.

tlake: +1

api/logic/task_logic.go
zpatrick: Agreed, and also think it's out of scope.

api/backend/ecs/service_manager.go
zpatrick: I think that would just be more tedious than anything tbh

api/backend/ecs/task_manager.go
zpatrick: tasks only have a single copy

api/backend/ecs/task_manager.go
zpatrick: no, we changed our aws layer to make this call more user-friendly.

api/backend/ecs/task_manager.go
zpatrick: that is handled in the logic layer

tests/smoke/task.bats
zpatrick: Doesn't "breaking change" imply we would need to bump to v0.11? If so, that is not true; these changes are still compatible with v0.10 APIs.

diemonster: it's a breaking change insofar as Chunnels re-implemented their workflows to use `copies`. Also, it changes the params for task create which in turn changes the request model, doesn't it?

api/backend/ecs/service_manager.go
diemonster: It is more tedious, but as it stands, it's a bit inconsistent with similar methods that do list operations --- we usually return a model, and here's we're not.

tests/smoke/task.bats
diemonster: Also, I'm still not entirely clear on how removing `copies` reduces the amount of API calls.

api/backend/interface.go
```diff
@@ -19,19 +19,21 @@ type Backend interface {
...
 	UpdateService(environmentID, serviceID, deployID string) (*models.Service, error)
 	GetServiceLogs(environmentID, serviceID, start, end string, tail int) ([]*models.LogFile, error)
 
-	CreateTask(envID, taskName, deployVersion string, overrides []models.ContainerOverride) (*models.Task, error)
+	CreateTask(environmentID, deployID string, overrides []models.ContainerOverride) (string, error)
 	ListTasks() ([]string, error)
-	GetTask(envID, taskID string) (*models.Task, error)
-	DeleteTask(envID, taskID string) error
-	GetTaskLogs(environmentID, taskID, start, end string, tail int) ([]*models.LogFile, error)
+	GetTask(environmentID, taskARN string) (*models.Task, error)
```
diemonster: I get why you are using `taskARN` versus `taskID` here, but it force the caller to know about critical helper functions like `lookupTaskARN()` without having much context into why. Do we leak the concept of an ARN anywhere else in the logic layer besides things relating to `task`s?

tests/smoke/task.bats
zpatrick: It does change the request model but not in an incompatible way. Using an older CLI to create tasks against this would just ignore the copies param.

zpatrick: However, to make the chunnels lives easier, we could actually keep the `copies` param in the cli. under the covers, it would just make a bunch of calls to the api.

api/backend/interface.go
zpatrick: nope

common/aws/ecs/ecs.go
```diff
@@ -25,13 +29,16 @@ type Provider interface {
...
+	DescribeClusterServices(clusterName, prefix string) ([]*Service, error)
+	DescribeServices(cluster string, service []string) ([]*Service, error)
 	Helper_DescribeServices(prefix string) ([]*Service, error)
 
 	DescribeTaskDefinition(familyAndRevision string) (*TaskDefinition, error)
 	Helper_DescribeTaskDefinitions(prefix string) ([]*TaskDefinition, error)
 
-	DescribeTasks(cluster string, taskArns []*string) ([]*Task, error)
+	DescribeTask(cluster string, taskARN string) (*Task, error)
+	DescribeTasks(clusterName string, taskARNs []*string) ([]*Task, error)
```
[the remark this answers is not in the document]
zpatrick: yes, keep in mind this is a patch, not a big feature set; we add stuff as we needed instead of going in and refactoring a bunch.

PR #476, opened 2018-01-04
[v0.10.4] 475: Remove AWS Region default and require region for API operation

### LINKED ISSUE #475: [v0.10.4] L0 API Region configuration should not have a default and should be required
### Expected behavior

When the L0 API is run, the environment variable and cli flag for the AWS Region should be provided and required either as a passed in CLI flag or as part of the user's environment

### Actual behavior

The AWS Region when not provided defaults to us-west-2. This is a dangerous assumption given that users might need to work in other regions.

jparsons04:
Removes DEFAULT_AWS_REGION and requires that the environment variable LAYER0_AWS_REGION be specified when launching the L0 API

Manually by running the L0 API locally. If LAYER0_AWS_REGION is not specified in your environment, the API will not run and you will be notified that the environment variable needs to be set.

**Checklist**
~- [ ] Unit tests~

closes #475 and #444


diemonster:
@jparsons04 should we add this variable as part of `l0-setup endpoint -d` ?

jparsons04:
Ah, that's an interesting thought. Probably, yeah.

PR #479, opened 2018-01-08
444: Fix l0-setup apply issue for instances created in another region

### LINKED ISSUE #444: failure to push remote tf state when not using default region

jparsons04:
When a new AWS Provider needs to be created or an AWS session needs to be established, aws.Config needs to contain the correct region for interactions with AWS to work.

In the case where l0 instances are initialized and set up in a region other than the default us-west-2, the resources will create in the region specified in the init wizard, but pushing terraform state to the S3 bucket fails because it was always assumed that the AWS region would be us-west-2.

To fix this, anytime an AWS Provider is needed, the instance derives what region it's in (assuming that l0-setup apply ran to completion), and then passes that region into the AWS Provider helper function so that aws.Config sets the correct region for API operations via the WithRegion() method.

The cli flag `--aws-region` is now no longer required in any l0-setup operation. The region the user specifies in `l0-setup init` is going to be the region used for AWS Provider operations through `l0-setup`

Unit tests, making a new l0 instance in a region other than us-west-2. `l0-setup apply` should correctly push terraform state to the bucket after apply. `l0-setup push` and `l0-setup pull` should also work.

**Checklist**
- [x] Unit tests

closes #444

### DIFF
--- a/setup/command/apply.go
+++ b/setup/command/apply.go
@@ -3,6 +3,7 @@ package command
 import (
 	"fmt"
 
+	"github.com/quintilesims/layer0/setup/instance"
 	"github.com/urfave/cli"
 )
 
@@ -27,18 +28,23 @@ func (f *CommandFactory) Apply() cli.Command {
 				return err
 			}
 
-			instance := f.NewInstance(args["NAME"])
-			if err := instance.Apply(!c.Bool("quick")); err != nil {
+			i := f.NewInstance(args["NAME"])
+			if err := i.Apply(!c.Bool("quick")); err != nil {
 				return err
 			}
 
 			if c.Bool("push") {
-				provider, err := f.newAWSProviderHelper(c)
+				region, err := i.Output(instance.OUTPUT_AWS_REGION)
 				if err != nil {
 					return err
 				}
 
-				if err := instance.Push(provider.S3); err != nil {
+				provider, err := f.newAWSProviderHelper(c, region)
+				if err != nil {
+					return err
+				}
+
+				if err := i.Push(provider.S3); err != nil {
 					return err
 				}
 			}
--- a/setup/command/apply.go
+++ b/setup/command/apply.go
@@ -3,7 +3,7 @@ package command
 import (
 	"fmt"
 
-	setup_instance "github.com/quintilesims/layer0/setup/instance"
+	"github.com/quintilesims/layer0/setup/instance"
 	"github.com/urfave/cli"
 )
 
@@ -28,13 +28,13 @@ func (f *CommandFactory) Apply() cli.Command {
 				return err
 			}
 
-			instance := f.NewInstance(args["NAME"])
-			if err := instance.Apply(!c.Bool("quick")); err != nil {
+			inst := f.NewInstance(args["NAME"])
+			if err := inst.Apply(!c.Bool("quick")); err != nil {
 				return err
 			}
 
 			if c.Bool("push") {
-				region, err := instance.Output(setup_instance.OUTPUT_AWS_REGION)
+				region, err := inst.Output(instance.OUTPUT_AWS_REGION)
 				if err != nil {
 					return err
 				}
@@ -44,7 +44,7 @@ func (f *CommandFactory) Apply() cli.Command {
 					return err
 				}
 
-				if err := instance.Push(provider.S3); err != nil {
+				if err := inst.Push(provider.S3); err != nil {
 					return err
 				}
 			}
--- a/setup/command/apply.go
+++ b/setup/command/apply.go
@@ -3,7 +3,7 @@ package command
 import (
 	"fmt"
 
-	"github.com/quintilesims/layer0/setup/instance"
+	setup_instance "github.com/quintilesims/layer0/setup/instance"
 	"github.com/urfave/cli"
 )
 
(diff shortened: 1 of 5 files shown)


setup/command/aws.go
```diff
@@ -23,14 +23,9 @@ var awsFlags = []cli.Flag{
...
-	cli.StringFlag{
-		Name:   "aws-region",
-		Usage:  "AWS region",
-		EnvVar: config.AWS_REGION,
-	},
 }
 
-func (f *CommandFactory) newAWSProviderHelper(c *cli.Context) (*aws.Provider, error) {
+func (f *CommandFactory) newAWSProviderHelper(c *cli.Context, region string) (*aws.Provider, error) {
 	// use default credentials and region settings
```
diemonster: [nit] is this comment still relevant?

setup/command/aws.go
```diff
@@ -23,14 +23,9 @@ var awsFlags = []cli.Flag{
...
 	},
-	cli.StringFlag{
-		Name:   "aws-region",
-		Usage:  "AWS region",
-		EnvVar: config.AWS_REGION,
-	},
 }
 
-func (f *CommandFactory) newAWSProviderHelper(c *cli.Context) (*aws.Provider, error) {
+func (f *CommandFactory) newAWSProviderHelper(c *cli.Context, region string) (*aws.Provider, error) {
```
diemonster: [nit] is `region` unavailable from any other existing object we have access to?

setup/command/list.go
```diff
@@ -58,7 +59,7 @@ func (f *CommandFactory) List() cli.Command {
 }
 
 func (f *CommandFactory) addRemoteInstances(c *cli.Context, current map[string]status) error {
-	provider, err := f.newAWSProviderHelper(c)
+	provider, err := f.newAWSProviderHelper(c, aws.DEFAULT_AWS_REGION)
```
diemonster: [nit] I'm mildly confused about using the name `DEFAULT_AWS_REGION` at this point --- is that really what this is?

jparsons04: So my thinking here is that `l0-setup list` is sort of a weird case. We need the AWS Provider, but the region doesn't matter at all because all list does when it checks for remote instances is it checks S3 (which is region agnostic for lists) and finds buckets that correlate to l0 instances. So because I modified the helper function, I need to pass in some sort of region, but the region doesn't matter, so I opt for the default region.

diemonster: can you pass in `nil` for the region, then? My concern here is that the code is a little misleading without the context you just gave.

jparsons04: The problem with sending `nil` is that I use `config.WithRegion(region)`, and the `aws.Config` struct has a field called Region which is required:
https://github.com/aws/aws-sdk-go/blob/master/aws/config.go#L70

So I have to pass a valid string (and a valid region) into `WithRegion()`

setup/command/aws.go
jparsons04: It's not worded very well. It is relevant because it initially grabs default values and then overrides them later in this function. I'll reword this to make it clearer.

setup/command/aws.go
jparsons04: It would have been available through the CLI context, but my code removes that as a possibility. With it gone from CLI context, the only other place it could come from is Instance context.

setup/command/apply.go
```diff
@@ -27,18 +28,23 @@ func (f *CommandFactory) Apply() cli.Command {
 				return err
 			}
 
-			instance := f.NewInstance(args["NAME"])
-			if err := instance.Apply(!c.Bool("quick")); err != nil {
+			i := f.NewInstance(args["NAME"])
```
zpatrick: [nit] the variable `i` is typically reserved for `for` loops

setup/command/pull.go
```diff
@@ -18,13 +19,18 @@ func (f *CommandFactory) Pull() cli.Command {
 				return err
 			}
 
-			provider, err := f.newAWSProviderHelper(c)
+			i := f.NewInstance(args["NAME"])
+			region, err := i.Output(instance.OUTPUT_AWS_REGION)
```
zpatrick: Unfortunately, this introduces a 'chicken and egg' problem. A user would need the instance files locally to get the output, but the purpose of this command is to get the instance files locally.  I recommend using a region flag for this command, or update the terraform module + user data script to always specify `us-west-2` when dealing w buckets.

jparsons04: It's going to have to be the first option, because when you hard code the terraform module (the s3 bucket resource) to create in us-west-2, then I actually get this error message on the first apply attempt: https://github.com/terraform-providers/terraform-provider-aws/issues/867

When the aws provider and the s3 bucket resource's region don't match, this happens. The bucket gets created but then terraform can't read from it.

setup/command/apply.go
```diff
@@ -3,7 +3,7 @@ package command
 import (
 	"fmt"
 
-	"github.com/quintilesims/layer0/setup/instance"
+	setup_instance "github.com/quintilesims/layer0/setup/instance"
```
zpatrick: We shouldn't rename packages unless there is a conflict among imports. My suggestion earlier to not use `i` as a variable name was meant to steer you towards using something like `inst := f.NewInstance(...)`

setup/command/push.go
```diff
@@ -3,7 +3,7 @@ package command
 import (
 	"fmt"
 
-	"github.com/quintilesims/layer0/setup/instance"
+	setup_instance "github.com/quintilesims/layer0/setup/instance"
```
zpatrick: ditto

PR #480, opened 2018-01-09
MAINT: Make l0 deploy list only return active task definitions

jparsons04:
The methods called in the AWS SDK that pull back Task Definitions should only be retrieving Task Definitions that are active. Without filtering Task Definitions by status, inactive Task Definitions can still be retrieved. For Layer0 instances that have been up for a long time, inactive Task Definitions can accumulate and has been known to slow the operation `l0 deploy list` down considerably.

Running `l0 deploy list` operations (both with and without the `--all` flag)

**Checklist**
~- [ ] Unit tests~

### DIFF
--- a/common/aws/ecs/ecs.go
+++ b/common/aws/ecs/ecs.go
@@ -973,6 +973,8 @@ func (this *ECS) ListTaskDefinitionFamiliesPages(prefix string) ([]*string, erro
 		FamilyPrefix: &prefix,
 	}
 
+	input.SetStatus(ecs.TaskDefinitionFamilyStatusActive)
+
 	connection, err := this.Connect()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


jparsons04:
This was mainly cherry picked by looking at what I did in this commit: https://github.com/quintilesims/layer0/commit/bd06b90c4578d7d00f676db3e4377cc20716bbf0

diemonster:
I'm fine with merging this, but in order to look at potential downsides, I'm struggling to think of a time when a Layer0 user would ever want/need an inactive taskDefinition. Possibly rollbacks, but even then the workflow of just making a new `deploy` has become the norm versus rolling back to an older version. Thoughts @zpatrick ?

zpatrick:
I don't think you can even target an inactive task definition for rollbacks, so I say merge as-is.

PR #481, opened 2018-01-09
Add check for layer0-api-endpoint

nehayward:
A simple fix to flow.sh to ensure the api endpoint is set for layer0

### DIFF
--- a/scripts/flow.sh
+++ b/scripts/flow.sh
@@ -124,6 +124,11 @@ if [ -z $LAYER0_PREFIX ]; then
     exit 1
 fi
 
+if [ -z $LAYER0_API_ENDPOINT ]; then
+    echo "LAYER0_API_ENDPOINT not set!"
+    exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


scripts/flow.sh
```diff
@@ -124,6 +124,11 @@ if [ -z $LAYER0_PREFIX ]; then
     exit 1
 fi
 
+if [ -z $LAYER0_API_ENDPOINT ]; then
```
zpatrick: where is this used?

nehayward: It needed it to do the delete flow. Or is flow only for local use? In which case I just needed the api running locally.

zpatrick: It can be used for either

nehayward: It's your call then to add this or not.

zpatrick: I'm going to not add this because I think it's more useful to have both options available.

PR #482, opened 2018-01-09
[v0.10.4] Fix System and Stress tests

zpatrick:
Updates our system and stress tests to run successfully for this version of Layer0 on top of Terraform v0.11.1. 
Includes an updated version of `tftest` along with fixing what appears to be bad merges.


tests/clients/layer0_test_client.go
```diff
@@ -1,18 +1,20 @@
...
 	"github.com/quintilesims/layer0/common/models"
 )
 
+type Tester interface {
+	Fatal(...interface{})
+}
+
 type Layer0TestClient struct {
-	T      *testing.T
+	T      Tester
```
tlake: I remember making this sort of change in the past - I don't now remember why it didn't stick. xD

Good change, though. :+1:

tests/stress/Makefile
```diff
@@ -1,7 +1,10 @@
...
+init:
+	pushd module ; \
+	ln -s -f $$(which terraform-provider-layer0) . ; \
+	terraform init ; \
+	popd ; \
 
-benchmark:
-	go test -v -run NOTHING -bench . -timeout 2h
+test: init
+	go test -v -run NOTHING -bench . -timeout 2h
```
tlake: [nit/suggestion] There may be some small amount of worth in adding the `-debug` flag here? When you ran these tests you had expressed some concern about how long they took, and `-debug` should spit out the names of each test case as they're run.

zpatrick: ah good call

PR #483, opened 2018-01-10
update windows ECS agent version

diemonster:
Updates the Windows ECS Agent to the latest version as specified in the canonical Windows ECS script https://github.com/aws/amazon-ecs-agent/blob/master/misc/windows-deploy/user-data.ps1#L7:17

Also removes a test target from our smoketests.

Create a Windows environment with `-os windows` and a sample windows service.

**Checklist**
- [X] Manual tests


tlake:
I encountered an error when trying to create a deploy from the sample Windows taskdef:

`InternalServerError (code: 21): Custom family names are currently unsupported in Layer0`

The sample task definition found at https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ECS_Windows_getting_started.html#create_windows_service requires a change in order to work with Layer0: the `"family"` key must be changed to `"familyName"`.

diemonster:
@tlake calling out for historical purposes that the "correct" task def is here https://github.com/quintilesims/guides/tree/master/windows . I'll be updating the corresponding README and such to reflect that we have proper support for Windows containers now.

Issue #361, opened 2017-09-26
Unclear error message when missing aws-access-key


nehayward:
### Expected behavior
After following [layer0 install](http://layer0.ims.io/setup/install/) and creating an Access AWS Key and entering it into the l0-setup, you still can't access aws without having aws-cli installed. This is because you are missing the aws-access key environment variables. Although it can be set with `l0-setup init --aws-access-key=...` it's not clear in the initial setup. 

### Actual behavior
After finishing the setup you get this message. I believe there should be at least a warning or error message and message to install aws-cli. This should also be added to the initial prerequisites [here](https://github.com/quintilesims/layer0/blob/develop/docs-src/docs/setup/install.md) too for the install docs. 

`No valid AWS credentials found. Please specify an AWS access key and secret key using their corresponding flags or environment variables`
<img width="1005" alt="error" src="https://user-images.githubusercontent.com/6733030/30864112-5f2c6bd0-a298-11e7-9d6f-d4fe00f2e140.png">

The error message occurs [here](https://github.com/quintilesims/layer0/blob/c815bb05a3a0945b9b853c8df66acc9b9c99feda/setup/command/aws.go#L47-L56) 

### Steps to reproduce the behavior
Delete aws-cli and run l0-setup init (instance-name)

- Layer0 Setup version v0.10.3

jparsons04:
There are a couple of areas in the code I feel need to be looked at.

1) It seems to me that the logic of `newAWSProviderHelper` needs to be reexamined, which is invoked during the [apply](https://github.com/quintilesims/layer0/blob/develop/setup/command/apply.go#L36) phase:

```
func (f *CommandFactory) newAWSProviderHelper(c *cli.Context) (*aws.Provider, error) {
     // use default credentials and region settings
     config := defaults.Get().Config
     
     // use static credentials if passed in by the user
     accessKey := c.String("aws-access-key")
     secretKey := c.String("aws-secret-key") 
     if accessKey != "" && secretKey != "" {
         staticCreds := credentials.NewStaticCredentials(accessKey, secretKey, "")
         config.WithCredentials(staticCreds)
     } else {
         logrus.Debugf("aws-access-key or aws-secret-key was not specified. Using default credentials")
     }
 
     // ensure credentials are available
     if _, err := config.Credentials.Get(); err != nil {
         if err, ok := err.(awserr.Error); ok && err.Code() == "NoCredentialProviders" {
[...]

PR #484, opened 2018-01-11
Update docs

nehayward:
This update the walkthrough docs to support the new terraform init as well as clarifying l0-setup init.

This can be checked by running the `make build; make serve` in /doc-src

closes #
#361


docs-src/mkdocs.yml
```diff
@@ -39,9 +39,17 @@ pages:
...
-    custom_dir: 'material'   
+  name: 'material'
+  palette:
+    primary: 'indigo'
+    accent: 'indigo'
+
+extra:
+    version: v0.10.3
+    logo: 'static/logo_icon.png'
+    author: xfra
```
jparsons04: So here, the `theme` block and the `extra` block changed with the new version of mkdocs (I'm using mkdocs v0.17.2)

I think the newer version of mkdocs did away with the `extra` block. Some stuff that you have here in the `extra` block actually needs to go in the `theme` block. What happens when I build this is our color scheme isn't preserved.

This is how I have the `theme` block:

```
 theme:
     name: 'material'
     custom_dir: 'material'
     palette:
         primary: 'red'
         accent:  'grey'
     font:
         text: 'Roboto'
         code: 'Roboto Mono'
     logo: 'static/logo_icon.png'
```

Give this a try and you should see the color scheme align with what is currently available at layer0.ims.io.

nehayward: Perfect! I have that version of mkdocs too. I'll update my pr

PR #495, opened 2018-01-16
Update release notes and add script to update version

### LINKED ISSUE #117: Migrate Layer0 Binaries to Github

nehayward:
This adds a little script to update the version number. The Release.md was updated as well

**Checklist**

closes #117


zpatrick:
Can you please do this for each location mentioned in the [release docs](https://github.com/quintilesims/layer0/blob/develop/RELEASE.md#update-documentation)?

Also, given this can probably be done in a handful of lines, how about we try putting this into a new target in the top-level `Makefile` instead of adding a new file into `/scripts`?

zpatrick:
Also, this doesn't close #117 since it's only addressing part of that issue.

nehayward:
@zpatrick I will add that part in too.

.travis.yml
```diff
@@ -1,19 +1,22 @@
...
 
-script:
-  - make unittest
-  - ./scripts/ci-smoketest.sh
+stage: Build Release
+script: make github-release
 
-after_script:
-  - make destroy-smoketest
+deploy:
```
zpatrick: We should have the team chime in, but I don't think we want/need to release the binaries on Github. We already have an easier pathway for users to download from S3 (because they can get the release for their specific OS), and adding that here just seems redundant and more code to maintain.

Makefile
```diff
@@ -17,6 +18,25 @@ release:
 		aws s3 cp build/$$os/layer0_$(L0_VERSION)_$$os.zip s3://xfra-layer0/release/$(L0_VERSION)/layer0_$(L0_VERSION)_$$os.zip ; \
 	done
 
+github-release:
```
zpatrick: If we are to keep the Github binaries, this should run at the end of the existing `release` target.  
This can be accomplished in one or two lines, because all of the binaries exist in `layer0/build` by the end of that target.

Makefile
```diff
@@ -1,5 +1,6 @@
 SHELL:=/bin/bash
-L0_VERSION?=$(shell git describe --tags)
+L0_VERSION:=$(shell git describe --tags | sed "s/v//")
```
zpatrick: This would break our versioning, this must be reverted back. Remember, we run this all the time, not just on "clean" releases like v0.10.4, but also on working branches where the version may be some long hash.

Makefile
```diff
@@ -1,5 +1,6 @@
...
+RELEASE:=$(shell cat docs-src/docs/releases.md | sed -n 3p | sed 's/0\.10\../'$(L0_VERSION)'/g')
```
zpatrick: This should be taken out - it would execute every time we ran `make`, regardless of the target.

Makefile
```diff
@@ -17,6 +18,25 @@ release:
...
+	$(MAKE) -C plugins/terraform release
+
+	rm -rf binaries
+	cp -R cli/build/ binaries 
+	cp -R setup/build/ binaries
+	cp -R plugins/terraform/build/  binaries 
+
+	zip -r binaries.zip binaries
+
+update-release:
```
zpatrick: Updating the docs' version references should be handled in a `release` target in `layer0/docs-src/Makefile`.

For example, in `layer0/Makefile`:
```
release:
        $(MAKE) -C api release
        $(MAKE) -C cli release
        $(MAKE) -C docs release                     <- added here
        $(MAKE) -C runner release
        $(MAKE) -C setup release
        $(MAKE) -C plugins/terraform release

        rm -rf build
        ...
```

and in `layer0/docs-src/Makefile`
```
build:
        mkdocs build --clean

release:
    # sed commands ...
    $(MAKE) build
```

RELEASE.md
```diff
@@ -60,11 +54,12 @@ Commit your changes and push them to the `release` branch when you are done.
 
 ## Update Documentation
 
-There are a couple of references in the [docs](https://github.com/quintilesims/layer0/tree/release/docs-src/docs) section that will need to be updated with the latest version:
+After running `make update-release` this will update a version number for these files and the
```
zpatrick: [KISS](https://en.wikipedia.org/wiki/KISS_principle)!!

We already have a release process that, aside from `git` housekeeping, works with a single command: 
```
make release
```
Let's try to keep it that way unless we absolutely can't.

RELEASE.md
```diff
@@ -60,11 +54,12 @@ Commit your changes and push them to the `release` branch when you are done.
...
 
-There are a couple of references in the [docs](https://github.com/quintilesims/layer0/tree/release/docs-src/docs) section that will need to be updated with the latest version:
+After running `make update-release` this will update a version number for these files and the 
 
 * [mkdocs.yml](https://github.com/quintilesims/layer0/blob/release/docs-src/mkdocs.yml#L40)
 * [index.md](https://github.com/quintilesims/layer0/blob/release/docs-src/docs/index.md)
 * [releases.md](https://github.com/quintilesims/layer0/blob/release/docs-src/docs/releases.md)
+* [README.MD](https://github.com/quintilesims/layer0/blob/release/README.md)
 
 Run `make build` from the `docs-src/` directory to compile the docs, commit your changes, and push them to the release branch when you are done.
```
zpatrick: As stated above, this should already run as part of the top-level `make release`  (something we should have done in the past, but would be a nice addition to this changeset).

RELEASE.md
```diff
@@ -114,4 +109,4 @@ summary of the release contents. For example:
 
 # Merge Master into Develop
 
-To bring the `develop` branch up-to-date with `master`, create a [create a pull request](https://github.com/quintilesims/layer0/compare) from the `master` branch targeting the `develop` branch (_base: develop, compare: master_).
+To bring the `develop` branch up-to-date with `master`, create a [create a pull request](https://github.com/quintilesims/layer0/compare) from the `master` branch targeting the `develop` branch (_base: develop, compare: master_). **Please squash and merge.**
```
zpatrick: rebase and merge is preferred, if possible

.travis.yml
zpatrick: Another note, we don't want to run this in Travis. For one thing, Travis doesn't have the permissions to push to S3 or our Dockerhub account. Also, if we did go the Travis route for releases, we should run system tests first, which is a very large unit of work. Any Travis related stuff should be dropped from this PR.

tlake: Oooh, great point about travis' permissions - I hadn't even thought about that.

As for releasing binaries on GitHub in addition to our docs, I'm sort of on the fence. I think that if we _do_ release on GitHub as well, we should definitely break it up by OS in the same way we do in the docs. But whether or not we _should_ do that... I don't know just yet. I like the idea of having binary releases so strongly coupled to the tagged version notes; I think that it's a good pattern found in a lot of other OSS repositories. It is redundant, though, given that we also link to binaries stored in S3 in our docs. Maybe the release process could be that the version notes on GitHub and the download links in our docs both link to the same source of binaries?

I guess what I like about releasing binaries on GitHub is that we're not tied to AWS resources paid for by IQVIA - if anything goes south and we lose access to those resources, our OSS tool is unaffected and still available on GitHub.

nehayward: Great thoughts everyone, just to clarify with the deploy portion. 
It'd only be building the binaries the current layer0.ims.io/#download which can be done on in travis. 

As far as s3 and DockerHub go we can continue to do that locally or figure out if it can be deployed via travis, which I believe can be done, ie https://docs.travis-ci.com/user/deployment/s3/

zpatrick: I guess my concern is more about releasing in multiple locations. To tlake's point, releasing on Github makes a lot of sense, and if we went that route, I'd like to see the S3 stuff go away.

docs-src/Makefile
```diff
@@ -1,10 +1,27 @@
...
 deps:
 	pip install mkdocs
 	pip install -Iv mkdocs-material==2.3.0
 
 build:
 	mkdocs build --clean
 
+release:
+	# Remove v from Version
+	$(eval L0_VERSION := $(shell echo $(L0_VERSION) | cut -c 2-))
```
zpatrick: For `README.md`, I think using `sed` along with stripping the `v` prefix is fine. 

For the rest of the files, I worry that using the stripped prefix could cause inconsistencies. 
I think we should consider using a tool like [expenv](https://github.com/blang/expenv) which replaces environment variables in template files. That way, we could include templated files, e.g. `docs-src/index.md.template` that contained:
```
...
## Download
| Download **$L0_VERSION** | |
...
```

and in this `Makefile`, we would have:
```
expenv -f docs/index.md.template > docs/index.md
```

For the problem of adding a new table entry, we could add a new template, reverse the order of our table, and use the append `>>` function: 
```
expenv -f docs/release_table.template >> docs/releases.md
```

.travis.yml
```diff
@@ -8,15 +8,23 @@ before_install:
...
+    fi
 
 deploy:
   provider: releases
+  script: make release
   api_key:
     secure: O0g3yP0OWCF2I96p2sY9SsmFrurbVWS1A4DthtWw+xVl8PYhAby7hzgYsy/Ap6C3L5b0PAdP8EtDKoYBy8A7tIImKtoIG7XZPqEH11WhFyUGlpr52DgeROdwZhpKsB3AjELzzSY+XYqW1pKxhcQPIbEBRSHfcREVJ3mL1yuXJkjrXDFCqnLe7oalR6rXaYfv27dJ9PK/Dmtpcu87I21rAX3Jmr94NFPfhXSaIHx9v4WOriMRYKNXvS0hxMH/UNlPNyKBFP6bMzMJFuIusmiJOx9ew/PmxyWen46rRytnJXj9YtaTeaK2Vn+6hEQ11PtwDn7ulou40ZIUiruxM1lmwN/uSIPmq5LVkx6Xsqfppu/jE8JLzitlme2D15xwJybdxNFaXCYG6TjlHjus4DPc72lLtO9OLNQr5bRc0IWrMlHPY49THkMnwMmqNxSLKhz56GcOOPUU/sNwCbReMlT0Ml/CKWwqqF8ul4MD/6quQjiemKsbGEXUTzy4myzO681R7/MQSioI6XF2fPaf+29CBB28uxUva7jbQSb7f1MaRagBBPL2a7P0Pk7ofw9nX49LOHLF8z9SZAbXT0/R4HgIBBt4yXOD1IyIXUitnNHYfbCQLUA4Mq7NFglGKfUPs9HG0afTm+eUJXxjUqEjXpNjrLC6TJOzutYXamBpTVaZe1w=
   file: 
-   - binaries.zip
+   - build/macOS.zip
```
zpatrick: not sure why `macOS` is the only non-capitalized name

README.md.b
```diff
@@ -0,0 +1,19 @@
+# ![logo](https://quintilesims.github.io/layer0/static/logo_rs.png)
```
zpatrick: I assume this should be removed?

docs-src/docs/releases.md
```diff
@@ -1,9 +1,9 @@
...
-| v0.10.0 | [macOS](https://s3.amazonaws.com/xfra-layer0/release/v0.10.0/layer0_v0.10.0_darwin.zip) | [Linux](https://s3.amazonaws.com/xfra-layer0/release/v0.10.0/layer0_v0.10.0_linux.zip) | [Windows](https://s3.amazonaws.com/xfra-layer0/release/v0.10.0/layer0_v0.10.0_windows.zip)
-| v0.9.0 | [macOS](https://s3.amazonaws.com/xfra-layer0/release/v0.9.0/layer0_v0.9.0_darwin.zip) | [Linux](https://s3.amazonaws.com/xfra-layer0/release/v0.9.0/layer0_v0.9.0_linux.zip) | [Windows](https://s3.amazonaws.com/xfra-layer0/release/v0.9.0/layer0_v0.9.0_windows.zip)
-| v0.8.4 | [macOS](https://s3.amazonaws.com/xfra-layer0/release/v0.8.4/layer0_v0.8.4_darwin.zip) | [Linux](https://s3.amazonaws.com/xfra-layer0/release/v0.8.4/layer0_v0.8.4_linux.zip) | [Windows](https://s3.amazonaws.com/xfra-layer0/release/v0.8.4/layer0_v0.8.4_windows.zip)
+| v0.10.4 | [macOS](https://github.com/quintilesims/layer0/releases/download/v0.10.4/macOS.zip) | [Linux](https://github.com/quintilesims/layer0/releases/download/v0.10.4/Linux.zip) | [Windows](https://github.com/quintilesims/layer0/releases/download/v0.10.4/Windows.zip)
```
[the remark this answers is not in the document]
nehayward: Yes it's possible and I will add them before closing this.

.travis.yml
nehayward: That's just how they name there OS's. like iOS, tvOS, etc.

docs-src/Makefile
nehayward: Okay sweet, I was wondering about using a template file but wasn't sure if the added resources would be wanted. I'll switch to this.

README.md.b
nehayward: Yup, missed that.

docs-src/docs/releases.md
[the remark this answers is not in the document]
diemonster: I don't think there's any value is having the older versions pre-v0.10 posted

RELEASE_NOTES.md
```diff
@@ -1,104 +0,0 @@
-# Release Notes
```
zpatrick: Why is this being deleted?

docs-src/Makefile
```diff
@@ -1,10 +1,24 @@
+L0_VERSION:=$(shell git describe --tags)
+export L0_VERSION_V:=""
```
zpatrick: i dont think this is required

docs-src/Makefile
```diff
@@ -1,10 +1,24 @@
...
 
 build:
 	mkdocs build --clean
 
+release:
+	# Remove v from Version
+	$(eval L0_VERSION_V := $(shell echo $(L0_VERSION) | cut -c 2-))
+
+	# Update Version to Latest and clean up
+	expenv -f index.md.template > docs/index.md
```
zpatrick: See my last comment -> `README.md` is the **only** place it is ok to use the stripped version, because we know it's specifically for the badge link. We don't want/need a template for README.md, so the `sed` should be done in place.

As I said before, any other replacements need to be done with the **untouched** `L0_VERSION` variable.

docs-src/README.md.template
```diff
@@ -0,0 +1,19 @@
+# ![logo](https://quintilesims.github.io/layer0/static/logo_rs.png)
```
zpatrick: this should be removed -> can use `sed` for this instance.

docs-src/index.md.template
```diff
@@ -0,0 +1,24 @@
...
+</center>
+
+
+## Meet Layer0
+Layer0 is a framework that helps you deploy web applications to the cloud with minimal fuss. Using a simple command line interface (CLI), you can manage the entire life cycle of your application without having to focus on infrastructure.
+
+Ready to learn more about Layer0? See our [introduction page](intro.md) to learn about some important concepts. When you're ready to get started, take a look at the [installation page](setup/install.md) for information about setting up Layer0.
+
+## Download
+| Download **v0.10.4** | | |
```
zpatrick: why is `v0.10.4` hardcoded in the template?

RELEASE_NOTES.md
nehayward: It seems redundant. I can bring it back, but I migrated all the notes to [Github](https://github.com/quintilesims/layer0/releases)

docs-src/Makefile
nehayward: I needed it for the expenv, otherwise the env wasn't being found. Not sure if it's a bug or something I did wrong.

docs-src/Makefile
nehayward: Okay. Thanks for clarifying I'll fix this.

docs-src/index.md.template
nehayward: Hmmm 😕 I need to fix that. I thought I checked this file out before I pushed it.

docs-src/README.md.template
nehayward: 👍

RELEASE_NOTES.md
zpatrick: Ah cool, didn't realize that 👍

docs-src/Makefile
zpatrick: since, from our convo below, `expenv` commands will be using `L0_VERSION` instead of `L0_VERSION-V`, I think this can be removed?

docs-src/docs/releases.md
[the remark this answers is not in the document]
nehayward: @diemonster I'll remove them.

docs/releases/index.html
```diff
@@ -594,46 +594,46 @@ <h1>Releases</h1>
...
+<td><a href="https://github.com/quintilesims/layer0/releases/download/v0.10.2/Linux.zip">Linux</a></td>
+<td><a href="https://github.com/quintilesims/layer0/releases/download/v0.10.2/Windows.zip">Windows</a></td>
 </tr>
 <tr>
-<td>v0.9.0</td>
-<td><a href="https://s3.amazonaws.com/xfra-layer0/release/v0.9.0/layer0_v0.9.0_darwin.zip">macOS</a></td>
-<td><a href="https://s3.amazonaws.com/xfra-layer0/release/v0.9.0/layer0_v0.9.0_linux.zip">Linux</a></td>
-<td><a href="https://s3.amazonaws.com/xfra-layer0/release/v0.9.0/layer0_v0.9.0_windows.zip">Windows</a></td>
+<td>v0.10.3</td>
+<td><a href="https://github.com/quintilesims/layer0/releases/download/v0.10.3/macOS">macOS</a></td>
```
jparsons04: I'm not sure how this happened if it was autogenerated, but somehow the .zip extension is missing from this link.

nehayward: Good catch. I'm not sure what happened I'll check that the autogen is still working

docs-src/Makefile
```diff
@@ -1,10 +1,22 @@
+L0_VERSION:=$(shell git describe --tags)
+export L0_VERSION
```
nehayward: Needed for expenv

docs-src/Makefile
zpatrick: This regex needs to be something like `v[0-9]+.[0-9]+.[0-9]+` 

-> `v<1 to many ints>.<1 to many ints>.<1 to many ints>-` e.g. `v111.2222.33333` is perfectly valid.

README.md
```diff
@@ -1,7 +1,7 @@
 # ![logo](https://quintilesims.github.io/layer0/static/logo_rs.png)
 
 [![Build Status](https://travis-ci.org/quintilesims/layer0.svg?branch=master)](https://travis-ci.org/quintilesims/layer0)
-[![Version](https://img.shields.io/badge/version-0.10.4-brightgreen.svg)](https://github.com/quintilesims/layer0/releases/tag/v0.10.4)
+[![Version](https://img.shields.io/badge/version-v0.10.5-7-g63bde91c-brightgreen.svg)](https://github.com/quintilesims/layer0/releases/tag/v0.10.5-7-g63bde91c)
```
zpatrick: need to remove extra stuff on the end (this is fine; we only expect to run this on tagged releases)

nehayward: ✅

docs-src/Makefile
```diff
@@ -1,10 +1,20 @@
...
 	pip install mkdocs
 	pip install -Iv mkdocs-material==2.3.0
+	go get github.com/blang/expenv
+	go install github.com/blang/expenv
 
 build:
 	mkdocs build --clean
 
+release:
+	$(shell sed -E -i '' 's/(v[0-9].[0-9]+.[0-9])/'$(L0_VERSION)'/g' ../README.md)
```
zpatrick: regex needs updated (see previous comment)

README.md
```diff
@@ -1,7 +1,7 @@
 # ![logo](https://quintilesims.github.io/layer0/static/logo_rs.png)
 
 [![Build Status](https://travis-ci.org/quintilesims/layer0.svg?branch=master)](https://travis-ci.org/quintilesims/layer0)
-[![Version](https://img.shields.io/badge/version-0.10.4-brightgreen.svg)](https://github.com/quintilesims/layer0/releases/tag/v0.10.4)
+[![Version](https://img.shields.io/badge/version-v0.10.4-brightgreen.svg)](https://github.com/quintilesims/layer0/releases/tag/v0.10.4)
```
jparsons04: Do we want to run the script now so that references to 0.10.4 like here and elsewhere update the version to 0.10.5?

nehayward: Yeah I could do that now. It'll make the merge easier.

PR #497, opened 2018-01-17
add overlapping az stipulation to vpc_id input

### LINKED ISSUE #496: tagged subnets on existing vpcs cannot overlap AZs
When creating a Layer0 instance on an existing VPC, the tagged subnets for each Tier (e.g. `Public` or `Private`) cannot have overlapping availability zones. Running `l0-setup apply` with this stipulation will case the following error to occur:

```
* module.layer0.module.api.aws_elb.api: 1 error(s) occurred:

* aws_elb.api: InvalidConfigurationRequest: ELB cannot be attached to multiple subnets in the same AZ.
```

We should call this out when describing the input variable for `vpc`.

zpatrick:
Adds AZ overlap stipulation to `vpc_id` input variable for `l0-setup`. 

closes #496

### DIFF
--- a/setup/instance/inputs.go
+++ b/setup/instance/inputs.go
@@ -94,6 +94,7 @@ created for you. Existing VPCs must satisfy the following constraints:
     - Have access to the public internet (through a NAT instance or gateway)
     - Have at least 1 public and 1 private subnet
     - Each subnet must be tagged with ["Tier": "Private"] or ["Tier": "Public"]
+    - The subnets in each tier cannot have overlapping availability zones
 
 Note that changing this value will destroy and recreate any existing resources.
 `


setup/instance/inputs.go
```diff
@@ -94,6 +94,7 @@ created for you. Existing VPCs must satisfy the following constraints:
     - Have access to the public internet (through a NAT instance or gateway)
     - Have at least 1 public and 1 private subnet
     - Each subnet must be tagged with ["Tier": "Private"] or ["Tier": "Public"]
+    - The subnets in each tier cannot have overlapping availability zones
```
jparsons04: What if you said "No more than one tagged subnet per tier per availability zone"?

zpatrick: that's better, but still not great. let me think a bit

diemonster: what about "Each subnet must reside in a unique availability zone"?

zpatrick:
going to merge as-is; can revisit if we think of better wording

PR #511, opened 2018-01-23
510: Modify flow.sh to fix the delete environment stage

### LINKED ISSUE #510: flow.sh delete in api-refactor doesn't actually delete environments
### Expected behavior

In the `api-refactor` branch, running `flow.sh delete` should delete any non-api environments, deploys and jobs.

### Actual behavior

The deploys and jobs are deleted but not the environments.

Create a few environments and then run `flow.sh delete` in the `scripts/` directory. Do this in the `api-refactor` branch.

jparsons04:
Changes `scripts/flow.sh` to fix an issue where the environments are not deleted due to a CLI syntax change made for `l0 environment delete`.

The --wait flag has been removed from l0 environment delete, and each of the iterations of the environments to delete should not be sent to the background.

Create a few environments and deploys and then run `flow.sh delete`. Confirm that everything except for the api environment and deploy are deleted.

**Checklist**
~- [ ] Unit tests~

closes #510

### DIFF
--- a/scripts/flow.sh
+++ b/scripts/flow.sh
@@ -33,7 +33,7 @@ delete() {
     for id in $environment_ids; do
         if [ "$id" != "api" ]; then
             echo -e $BULLET "$id"
-            l0 environment delete $id --wait > /dev/null &
+            l0 environment delete $id > /dev/null
         fi
     done
 


scripts/flow.sh
```diff
@@ -33,7 +33,7 @@ delete() {
     for id in $environment_ids; do
         if [ "$id" != "api" ]; then
             echo -e $BULLET "$id"
-            l0 environment delete $id --wait > /dev/null &
+            l0 environment delete $id > /dev/null
```
diemonster: [nit] you could probably keep the `&` if you added a `wait` command after the for loop.

jparsons04: We would need to capture every PID that is spawned. This is a simpler solution because it simply just executes each delete and waits for the environment to delete before moving on to the next one. I'm a little concerned about backgrounding tasks like this. I'm not sure we'd want to for instance spawn 100 background tasks at the same time to handle cleanup.

PR #514, opened 2018-01-23
[WIP] Environment -recursive delete

### LINKED ISSUE #502: Check dependencies on Environment Delete
### api/provider/aws
* EnvironmentDelete() should check if there are any existing services, load balancers, and/or tasks in the environment before it starts to delete. If so, it should send back a nice error message.

### cli/command
* Add `l0 environment delete -r/--recursive NAME` bool flag 
  * If `-r` is set, the cli will delete all services, load balancers, and tasks before calling environment delete
  * Otherwise, immediately call environment delete (the API should communicate the "nice error message" from above if this would cause a dependency delete conflict).

nehayward:
Updates the api and the cli to delete dependencies on env.

Still needs to be tested, waiting on api-refactor init.

**Checklist**
- [x] Unit tests

closes #502


api/provider/aws/environment_delete.go
```diff
@@ -53,6 +59,43 @@ func (e *EnvironmentProvider) Delete(environmentID string) error {
...
 
+func (e *EnvironmentProvider) checkEnvironmentDependencies(environmentID string) error {
+	loadBalancerTags, err := e.TagStore.SelectByType("load_balancer")
+	if err != nil {
+		return err
+	}
+
+	dependentLoadBalancers := loadBalancerTags.WithKey("environment_id").WithValue(environmentID)
+	if len(dependentLoadBalancers) > 0 {
+		msg := fmt.Sprintf("Cannot delete non-empty environment '%s' (environment contains one or more load balancers).", environmentID)
```
zpatrick: For each of these errors, we should add the id of the dependent entity. 
Also, the `msg` line is too long, we should break it down into multiple lines.

cli/command/environment.go
```diff
@@ -63,6 +63,16 @@ func (e *EnvironmentCommand) Command() cli.Command {
 				Usage:     "delete an environment",
 				ArgsUsage: "ENVIRONMENT_NAME",
 				Action:    e.delete,
+				Flags: []cli.Flag{
+					cli.BoolFlag{
+						Name:  "r",
```
zpatrick: This is how you specify aliases with the cli package
```go
cli.BoolFlag{
    Name: "r, recursive",
    ...
}
```

cli/command/base.go
```diff
@@ -75,6 +75,20 @@ func (b *CommandBase) deleteHelper(c *cli.Context, entityType string, deleteFN f
 	return nil
 }
 
+func (b *CommandBase) entityIDHelper(c *cli.Context, entityType string) (string, error) {
```
zpatrick: not needed, should be removed

cli/command/environment.go
```diff
@@ -166,11 +176,62 @@ func (e *EnvironmentCommand) create(c *cli.Context) error {
 }
 
 func (e *EnvironmentCommand) delete(c *cli.Context) error {
+	if c.Bool("r") || c.Bool("recursive") {
+		environmentID, err := e.entityIDHelper(c, "environment")
+		if err != nil {
+			return err
+		}
+
+		if err := e.deleteDependecies(environmentID); err != nil {
```
zpatrick: [nit] I'm not convinced this helper is giving us much of a 'win'.
Personally, I'd either remove it or break it down by type, e.g.
```go
if c.Bool("recursive") {
    if err := e.deleteEnvironmentLoadBalancers(environmentID); err != nil {
        return err
    }

    if err := e.deleteEnvironmentServices(environmentID); err != nil {
        return err
    }

    if err := e.deleteEnvironmentTasks(environmentID); err != nil {
        return err
    }
}
```

api/provider/aws/environment_delete.go
tlake: ✅

cli/command/environment.go
nehayward: ✅

cli/command/environment.go
nehayward: ✅

cli/command/base.go
nehayward: ✅

api/provider/aws/test_aws/environment_delete_test.go
```diff
@@ -117,3 +118,54 @@ func TestDeleteEnvironmentIdempotence(t *testing.T) {
...
+	}
+
+	for _, tag := range tags {
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	target := provider.NewEnvironmentProvider(mockAWS.Client(), tagStore, mockConfig)
+	if err := target.Delete("env_id"); err.Error() != error.Error(errors.Newf(errors.DependencyError, "Cannot delete environment 'env_id' because it contains dependent services: svc_id")) {
```
zpatrick: you should only check that error code are equal

api/provider/aws/test_aws/environment_delete_test.go
```diff
@@ -117,3 +118,54 @@ func TestDeleteEnvironmentIdempotence(t *testing.T) {
...
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockAWS := awsc.NewMockClient(ctrl)
+	tagStore := tag.NewMemoryStore()
+	mockConfig := mock_config.NewMockAPIConfig(ctrl)
+
+	mockConfig.EXPECT().Instance().Return("test").AnyTimes()
+
+	tags := models.Tags{
```
zpatrick: there is too much unused setup and assertions in this test; anything this isn't directly related to testing the specific behavior should be taken out. Also, we should test for each dependent entity type: loadbalancer, service, and task.

cli/command/environment.go
```diff
@@ -166,6 +172,54 @@ func (e *EnvironmentCommand) create(c *cli.Context) error {
 }
 
 func (e *EnvironmentCommand) delete(c *cli.Context) error {
+	if c.Bool("r") || c.Bool("recursive") {
```
zpatrick: only use the long name, not the alias.

cli/command/environment.go
```diff
@@ -166,6 +172,54 @@ func (e *EnvironmentCommand) create(c *cli.Context) error {
...
+		args, err := extractArgs(c.Args(), "NAME")
+		if err != nil {
+			return err
+		}
+
+		environmentID, err := e.CommandBase.resolveSingleEntityIDHelper("environment", args["NAME"])
```
zpatrick: this has to be done regardess, so it should be done before the if block.

cli/command/environment.go
```diff
@@ -166,6 +172,54 @@ func (e *EnvironmentCommand) create(c *cli.Context) error {
...
+
+		// Delete Load Balancers
+		loadBalancers, err := e.client.ListLoadBalancers()
+		if err != nil {
+			return err
+		}
+
+		for _, loadBalancer := range loadBalancers {
+			if loadBalancer.EnvironmentID == environmentID {
+				e.client.DeleteLoadBalancer(loadBalancer.LoadBalancerID)
```
zpatrick: error swallowing!

cli/command/environment.go
```diff
@@ -166,6 +172,54 @@ func (e *EnvironmentCommand) create(c *cli.Context) error {
...
+
+		// Delete Tasks
+		tasks, err := e.client.ListTasks()
+		if err != nil {
+			return err
+		}
+
+		for _, task := range tasks {
+			if task.EnvironmentID == environmentID {
+				e.client.DeleteTask(task.TaskID)
```
zpatrick: ditto

cli/command/environment.go
```diff
@@ -166,6 +172,54 @@ func (e *EnvironmentCommand) create(c *cli.Context) error {
...
+
+		// Delete Services
+		services, err := e.client.ListServices()
+		if err != nil {
+			return err
+		}
+
+		for _, service := range services {
+			if service.EnvironmentID == environmentID {
+				e.client.DeleteTask(service.ServiceID)
```
zpatrick: ditto

cli/command/environment_test.go
```diff
@@ -212,6 +212,57 @@ func TestDeleteEnvironment(t *testing.T) {
...
+			JobID:  "job_id",
+			Status: models.CompletedJobStatus,
+			Result: "entity_id",
+		}
+
+		base.Resolver.EXPECT().
+			Resolve("environment", "env_name").
+			Return([]string{"env_id"}, nil)
+
+		base.Client.EXPECT().
```
zpatrick: always specify what you are returning

cli/command/environment_test.go
```diff
@@ -212,6 +212,57 @@ func TestDeleteEnvironment(t *testing.T) {
 	})
 }
 
+func TestDeleteEnvironmentRecursive(t *testing.T) {
```
zpatrick: this test isn't testing the behavior the name suggests.

api/provider/aws/environment_delete.go
```diff
@@ -53,6 +59,61 @@ func (e *EnvironmentProvider) Delete(environmentID string) error {
 	return nil
 }
 
+func (e *EnvironmentProvider) checkEnvironmentDependencies(environmentID string) error {
+	loadBalancerTags, err := e.TagStore.SelectByType("load_balancer")
+	if err != nil {
+		return err
+	}
+
+	dependentLoadBalancerTags := loadBalancerTags.WithKey("environment_id").WithValue(environmentID)
```
zpatrick: all of this logic follows the same pattern: get a slice of tags, filter by `WithKey("environment_id").WithValue(environmentID)`, if the length of the filtered tags are > 0, generate a new slice from the filtered tags, then throw an error

cli/command/environment.go
nehayward: ✅

cli/command/environment.go
nehayward: ✅

cli/command/environment.go
nehayward: ✅

cli/command/environment.go
nehayward: ✅

cli/command/environment_test.go
nehayward: ✅

api/provider/aws/test_aws/environment_delete_test.go
```diff
@@ -117,3 +118,59 @@ func TestDeleteEnvironmentIdempotence(t *testing.T) {
...
+	if err := tagStore.Insert(envTag); err != nil {
+		t.Fatal(err)
+	}
+
+	var dependencies = []struct {
+		entityID   string
+		entityType string
+		err        string // expected result
+	}{
+		{"lb_id", "load_balancer", error.Error(errors.Newf(errors.DependencyError, "Cannot delete environment 'env_id' because it contains dependent load balancers: lb_id"))},
```
nehayward: How can reduce this part?

api/provider/aws/test_aws/environment_delete_test.go
```diff
@@ -117,3 +118,59 @@ func TestDeleteEnvironmentIdempotence(t *testing.T) {
...
+		}
+
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+
+		target := provider.NewEnvironmentProvider(mockAWS.Client(), tagStore, mockConfig)
+		actual := target.Delete("env_id")
+
+		if actual.Error() != dependency.err {
```
nehayward: First shot, let me know what you think.

zpatrick: Better, but can still be improved. This will panic if `actual == nil` (also, use the established verbiage please: `err` or possibly `result`). 

You should look at other examples of unit tests in our code base that check error codes. The idempotent entity deletes are a good place.

api/provider/aws/test_aws/environment_delete_test.go
zpatrick: Take a step back and find the absolute **smallest possible amount** of things need to change between test cases while still testing what you need to. Then, build your table tests around a slice or map of that type (this is usually an [anonymous struct](https://talks.golang.org/2012/10things.slide#3) when there are multiple things that need to change).

Without giving you the answer, I can tell you that we can run these table tests over a slice of strings, e.g.
```
cases := []string{"...", "..."}
for _, c := range cases {
    ...
}
```

cli/command/environment_test.go
```diff
@@ -212,6 +212,60 @@ func TestDeleteEnvironment(t *testing.T) {
 	})
 }
 
+func TestDeleteEnvironmentDependencies(t *testing.T) {
```
zpatrick: This isn't testing what the name suggests; currently, you don't assert that we call delete on each dependent entity for that environment.

api/provider/aws/environment_delete.go
zpatrick: I think we can reduce a lot of this code given that it all follows a similar pattern.

api/provider/aws/test_aws/environment_delete_test.go
```diff
@@ -117,3 +119,50 @@ func TestDeleteEnvironmentIdempotence(t *testing.T) {
...
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockAWS := awsc.NewMockClient(ctrl)
+	tagStore := tag.NewMemoryStore()
+	mockConfig := mock_config.NewMockAPIConfig(ctrl)
+
+	mockConfig.EXPECT().Instance().Return("test").AnyTimes()
+
+	envTag := models.Tag{
```
zpatrick: [nit] even though it's a little less efficient, it may read better to keep the tag insertions together, e.g.:
```go
for _, c := range cases {
    tags := []models.Tag{
        { environment tag here },
        { entity tag here },
    }

    for _, tag := range tags {
        if err := tagStore.Insert(tag); err != nil { ... }
    }

    ...
}
```

api/provider/aws/test_aws/environment_delete_test.go
```diff
@@ -117,3 +119,50 @@ func TestDeleteEnvironmentIdempotence(t *testing.T) {
...
+		}
+
+		target := provider.NewEnvironmentProvider(mockAWS.Client(), tagStore, mockConfig)
+		if err := target.Delete("env_id"); err != nil {
+			if !strings.Contains(err.Error(), errors.DependencyError.String()) {
+				t.Errorf("Error: %s", err)
+			}
+		}
+
+		if err := tagStore.Delete(tag.EntityType, tag.EntityID, tag.Key); err != nil {
```
zpatrick: I think it would be cleaner to create a new `tagStore` for each case instead of doing cleanup. This may be more of a personal philosophy, but I think readability is more important than speed/efficiency for unit tests.

api/provider/aws/test_aws/environment_delete_test.go
```diff
@@ -117,3 +119,50 @@ func TestDeleteEnvironmentIdempotence(t *testing.T) {
...
+			Value:      "env_id",
+		}
+
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+
+		target := provider.NewEnvironmentProvider(mockAWS.Client(), tagStore, mockConfig)
+		if err := target.Delete("env_id"); err != nil {
+			if !strings.Contains(err.Error(), errors.DependencyError.String()) {
```
zpatrick: We should be doing assertions on error codes here. Some other things need to be worked out here:
- what happens if `err == nil`?
- what happens if `err` is not of type `*errors.ServerError`?
- why does this test fail if the expected error codes match?

cli/command/environment.go
```diff
@@ -166,6 +172,60 @@ func (e *EnvironmentCommand) create(c *cli.Context) error {
...
+
+		for _, loadBalancer := range loadBalancers {
+			if loadBalancer.EnvironmentID == environmentID {
+				if _, err := e.client.DeleteLoadBalancer(loadBalancer.LoadBalancerID); err != nil {
+					return err
+				}
+			}
+		}
+
+		// Delete Tasks
```
zpatrick: [nit] the comments seem redundant.

cli/command/environment.go
```diff
@@ -166,6 +172,60 @@ func (e *EnvironmentCommand) create(c *cli.Context) error {
...
+
+		// Delete Services
+		services, err := e.client.ListServices()
+		if err != nil {
+			return err
+		}
+
+		for _, service := range services {
+			if service.EnvironmentID == environmentID {
+				if _, err := e.client.DeleteTask(service.ServiceID); err != nil {
```
zpatrick: should be `DeleteService` (this is an example why it is important to test these code paths in the unit tests).

cli/command/environment_test.go
```diff
@@ -212,6 +212,60 @@ func TestDeleteEnvironment(t *testing.T) {
 	})
 }
 
+func TestCheckEnvironmentDependencies(t *testing.T) {
```
zpatrick: This test doesn't assert the behavior it claims it does: there are no calls to delete an environment's dependencies.

docs-src/docs/reference/cli.md
```diff
@@ -249,6 +249,7 @@ l0 environment delete [--wait] environmentName
 
 #### Optional arguments
 * `--wait` - Wait until the deletion is complete before exiting.
+* `-r --recursive` - This will attempt to delete all dependencies (loadbalancers, services, tasks) on environment and then delete environment.
```
zpatrick: Don't forget to add/update smoketests whenever you update the CLI

cli/command/environment_test.go
nehayward: What should we name it? It doesn't claim to delete. It says check.

api/provider/aws/test_aws/environment_delete_test.go
nehayward: Yeah I was on the fence about this but I agree.

api/provider/aws/environment_delete.go
nehayward: ✅

cli/command/environment_test.go
zpatrick: Ah, my mistake, I thought this was `TestDeleteEnvironmentDependencies`. I think think it should be the name I just said, and it should test each delete codepath.

tests/smoke/environment.bats
```diff
@@ -28,6 +28,10 @@
     l0 environment create --os windows test4
 }
 
+@test "environment create test5" {
```
nehayward: Not sure if we should add a service or task to the environment to test the recursive delete, breaks patterns of bats.

api/provider/aws/environment_delete.go
```diff
@@ -53,6 +59,42 @@ func (e *EnvironmentProvider) Delete(environmentID string) error {
 	return nil
 }
 
+func (e *EnvironmentProvider) checkEnvironmentDependencies(environmentID string) error {
+	deleteEntity := func(entity string) error {
```
zpatrick: this variable name is misleading, maybe something like `checkEntityTags`?

api/provider/aws/environment_delete.go
```diff
@@ -53,6 +59,42 @@ func (e *EnvironmentProvider) Delete(environmentID string) error {
...
+		tags, err := e.TagStore.SelectByType(entity)
+		if err != nil {
+			return err
+		}
+
+		dependentTags := tags.WithKey("environment_id").WithValue(environmentID)
+		if len(dependentTags) > 0 {
+			IDs := []string{}
```
zpatrick: even with acronyms, it's best practice to always have lowercase variable names - maybe `entityIDs`?

api/provider/aws/environment_delete.go
```diff
@@ -53,6 +59,42 @@ func (e *EnvironmentProvider) Delete(environmentID string) error {
...
+			for _, tag := range dependentTags {
+				IDs = append(IDs, tag.EntityID)
+			}
+
+			msg := "Cannot delete environment '%s' because it contains dependent %s: %s"
```
zpatrick: [nit] should be `... contains dependent %ss: %s` -> `... contains dependent services`

api/provider/aws/environment_delete.go
```diff
@@ -53,6 +59,42 @@ func (e *EnvironmentProvider) Delete(environmentID string) error {
...
+			msg = fmt.Sprintf(msg, environmentID, entity, strings.Join(IDs, ", "))
```
zpatrick: [nit] this is a little different then how we typically do long strings. I think it's a littler clearer what's going on by doing `fmt.Sprintf` in-line as opposed to having that in separate lines. For example:
```
msg := fmt.Sprintf("Cannot delete environment while it contains dependent %ss: ", entityType)
msg += strings.Join(entityIDs, ", ")
return errors.New(...)
```

api/provider/aws/job_runner.go
```diff
@@ -170,54 +170,7 @@ func (r *JobRunner) deleteDeploy(jobID, deployID string) (string, error) {
 }
 
 func (r *JobRunner) deleteEnvironment(jobID, environmentID string) (string, error) {
-	loadBalancers, err := r.loadBalancerProvider.List()
```
zpatrick: You'll need to merge from `api-refactor`

api/provider/aws/test_aws/environment_delete_test.go
```diff
@@ -117,3 +118,47 @@ func TestDeleteEnvironmentIdempotence(t *testing.T) {
...
+		}
+
+		for _, tag := range tags {
+			if err := tagStore.Insert(tag); err != nil {
+				t.Fatal(err)
+			}
+		}
+
+		target := provider.NewEnvironmentProvider(mockAWS.Client(), tagStore, mockConfig)
+		if err := target.Delete("env_id"); err != nil {
```
zpatrick: what if `err == nil`?

api/provider/aws/test_aws/environment_delete_test.go
```diff
@@ -117,3 +118,47 @@ func TestDeleteEnvironmentIdempotence(t *testing.T) {
...
+			if serverError, ok := err.(*errors.ServerError); ok {
```
zpatrick: what if this conversion fails?

api/provider/aws/test_aws/environment_delete_test.go
```diff
@@ -117,3 +118,47 @@ func TestDeleteEnvironmentIdempotence(t *testing.T) {
...
+				if serverError.Code != errors.DependencyError {
+					t.Errorf("Error: %s", err)
```
zpatrick: If someone who didn't write this test saw this error message, I don't think they would know what is going on. I think we should have a message here explaining what went wrong vs. what was expected to happen.

cli/command/environment_test.go
zpatrick: To clarify, because `TestDeleteEnvironmentDependencies` executes all of these code paths and extends beyond that, it makes this test redundant; because of that, I don't think we need this test.

tests/smoke/environment.bats
zpatrick: ah good call. I can see an argument for both, I personally like the way you've done it here because it keeps the environment "suite" of tests fast.

api/provider/aws/test_aws/environment_delete_test.go
nehayward: Wouldn't that mean the tests fail? Since we expect a Server Error. Just trying to clarify. I'm learning a lot so thanks for all the help.

api/provider/aws/environment_delete.go
nehayward: Should we do it based on count, or is that an overkill 🤔

api/provider/aws/environment_delete.go
nehayward: Yeah I agree with this. It makes it easier to read.

api/provider/aws/environment_delete.go
```diff
@@ -53,6 +59,42 @@ func (e *EnvironmentProvider) Delete(environmentID string) error {
...
+		}
+
+		dependentTags := tags.WithKey("environment_id").WithValue(environmentID)
+		if len(dependentTags) > 0 {
+			entityIDs := []string{}
+			for _, tag := range dependentTags {
+				entityIDs = append(entityIDs, tag.EntityID)
+			}
+
+			msg := fmt.Sprintf("Cannot delete environment '%s' because it contains dependent %ss:", environmentID, entityType)
```
zpatrick: [nit] you'll probably want a space after the last `:`

api/provider/aws/environment_delete.go
```diff
@@ -53,6 +59,42 @@ func (e *EnvironmentProvider) Delete(environmentID string) error {
...
+			msg += strings.Join(entityIDs, ", ")
+			return errors.Newf(errors.DependencyError, msg)
+		}
+		return nil
+	}
+
+	if err := checkDependentEntityTags("load_balancer"); err != nil {
```
tlake: [nit] We could avoid some manual repetition here
```
for _, entity := range []string{"load_balancer", "service", "task"} {
    if err := checkDependentEntityTags(entity); err != nil {
        return err
    }
}
```

nehayward:
I need to make some changes with the refactor.

api/provider/aws/test_aws/environment_delete_test.go
```diff
@@ -117,3 +118,48 @@ func TestDeleteEnvironmentIdempotence(t *testing.T) {
...
+			if err := tagStore.Insert(tag); err != nil {
+				t.Fatal(err)
+			}
+		}
+
+		target := provider.NewEnvironmentProvider(mockAWS.Client(), tagStore, mockConfig)
+		err := target.Delete("env_id")
+		if serverError, ok := err.(*errors.ServerError); ok {
+			if serverError.Code != errors.DependencyError {
+				t.Errorf("Expected DependencyError for entity type %s\nActual Error: %s", c, err)
```
zpatrick: it would probably be easier to use `assert` pkg here:
```go
err := target.Delete("env_id")
if err, ok := err.(*errors.ServerError); ok {
    assert.Equal(t, errors.DependencyError, err.Code)
} else {
    t.Fatalf("Error was not of type *errors.ServerError: %#v", err)
}
```

note that using `%#v` will show the object and its type

cli/command/environment_test.go
```diff
@@ -387,3 +387,90 @@ func TestUnlinkEnvironmentsInputErrors(t *testing.T) {
 		})
 	}
 }
+
+func TestEnvironmentRecursiveDelete(t *testing.T) {
```
zpatrick: 👍 good test! Now you just need to merge from `api-refactor` and fix this up a bit to satisfy those changes.

api/provider/aws/environment_delete.go
nehayward: I like

tests/smoke/environment.bats
```diff
@@ -23,7 +23,67 @@
   l0 environment unlink --bi-directional env_name1 env_name2
 }
 
+@test "environment create --user-data common/user_data.sh test2" {
```
zpatrick: Something went wonky while merging I'm guessing? Looks like most of these "new" tests have been deleted or are duplicates.

nehayward: Good catch, yeah I'm looking into that now.

api/provider/aws/test_aws/environment_delete_test.go
nehayward: Yeah that reads better. Any obvious advantage for choosing one over the other?

zpatrick: If you mean the `assert` package, it's less lines of code which tends to make things more readable. 

If you mean `%s` vs `%#v`, the latter gives more detailed output when the test fails, something like:
```
--- FAIL: TestThing (0.00s)
        some_test.go:6: Error was not of type *errors.ServerError: &errors.errorString{s:"foo"}
```
vs
```
--- FAIL: TestThing (0.00s)
        some_test.go:6 Assertion to ServerError failed
            Error: foo
```

cli/command/task_test.go
```diff
@@ -23,18 +24,31 @@ func TestCreateTask(t *testing.T) {
...
 		},
 	}
 
+	validateOverride := func(req models.CreateTaskRequest) {
+		assert.Equal(t, "tsk_name", req.TaskName)
+		assert.Equal(t, "dpl_id", req.DeployID)
+		assert.Equal(t, "env_id", req.EnvironmentID)
+
+		assert.Len(t, req.ContainerOverrides, 2)
+		for _, override := range req.ContainerOverrides {
```
zpatrick: we should iterate through `expected` here instead of `req.ContainerOverrides`

nehayward:
I had to update travis to 1.9 for terraform [related issue](https://github.com/mcuadros/terraform-provider-helm/issues/22)

Issue #389, opened 2017-10-11
Container exit codes should be non-0 when hitting "CannotPullContainerError"


zpatrick:
### Expected behavior
A task's exit code should be non-zero when a  `CannotPullContainerError` error occurs. 

### Actual behavior
When tasks get a `CannotPullContainerError`, the associated exit code is 0 (since the task never started). 
```
{
     "container_name": "foo",
     "exit_code": 0,
    "last_status": "STOPPED",
    "reason": "CannotPullContainerError: Error: image foo/foo not found"
}
```

diemonster:
Should we be fixing specifically to errors around `CannotPullContainer`? It seems like any case where we get an exit code of 0 and `last_status` of `STOPPED` would qualify for Layer0 presenting the exit code as != 0.

diemonster:
In testing, I noticed some additional behavior we should consider fixing. 

For one, `CannotPullContainerError` doesn't actually return an exit code, which is why we return 0. That's easy enough to fix, however we're oddly handling other non-zero errors for containers. For example, I created a task def with a container's `COMMAND` that would intentionally return a `1` exit code (e.g. `cat file`). It returned as `1` via the AWS ECS Console, but we still return "COMPLETED" via Layer0. 

It seems to we need another column of information that gives users' the container's exit code (or lack thereof). Otherwise, I'm not clear on what we're really telling users via the "STATUS" column other than the task was processed by the ECS API.

zpatrick:
I'm not seeing any unintended behavior in the first paragraph. There is already an `exit_code` field in the `Container` model in the `api-refactor` branch.

diemonster:
Will address UX issue here https://github.com/quintilesims/layer0/issues/508

tribaljack:
fixed, closing

PR #516, opened 2018-01-23
Add non-zero exit code when 'CannotPullContainerError` is returned

diemonster:
When a container cannot successfully pull down an image, ECS does return an error code. This PR manually sets the error code to '1'.

Still needs to be manually tested.

**Checklist**
- [x] Unit tests

closes #
https://github.com/quintilesims/layer0/issues/389


api/provider/aws/task_read.go
```diff
@@ -42,6 +42,10 @@ func (t *TaskProvider) Read(taskID string) (*models.Task, error) {
 
 	model.Containers = make([]models.Container, len(task.Containers))
 	for i, c := range task.Containers {
+		// When a container cannot pull an image, no exit code is returned from AWS. We set it to 1 manually to signal an error.
+		if strings.Contains(aws.StringValue(c.Reason), "CannotPullContainerError") {
+			c.SetExitCode(1)
+		}
```
zpatrick: [nit] newline after closing bracket

api/provider/aws/test_aws/task_read_test.go
```diff
@@ -110,3 +110,102 @@ func TestTaskRead(t *testing.T) {
 
 	assert.Equal(t, expected, result)
 }
+
+func TestTaskRead_CannotPullContainer(t *testing.T) {
```
zpatrick: there is too much unused setup and assertions in this test; anything this isn't directly related to testing the specific behavior should be taken out

zpatrick: There are still unnecessary setup and assertions in this test after the last update; I will call them out below.

api/provider/aws/test_aws/task_read_test.go
```diff
@@ -110,3 +110,96 @@ func TestTaskRead(t *testing.T) {
...
+	mockConfig.EXPECT().Instance().Return("test").AnyTimes()
+
+	tags := models.Tags{
+		{
+			EntityID:   "tsk_id",
+			EntityType: "task",
+			Key:        "name",
+			Value:      "tsk_name",
+		},
+		{
```
zpatrick: I believe the `arn` tag is the only tag required to do the proper lookup; all others should be removed.

api/provider/aws/test_aws/task_read_test.go
```diff
@@ -110,3 +110,96 @@ func TestTaskRead(t *testing.T) {
...
+		},
+	}
+
+	for _, tag := range tags {
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	describeTaskInput := &ecs.DescribeTasksInput{}
```
zpatrick: we don't care about if the `describeTaskInput` is correct in this test; remove it any use `gomock.Any()` in the `EXPECT()` statement

api/provider/aws/test_aws/task_read_test.go
```diff
@@ -110,3 +110,96 @@ func TestTaskRead(t *testing.T) {
...
+	describeTaskInput.SetCluster("l0-test-env_id")
+	describeTaskInput.SetTasks([]*string{aws.String("arn:aws:ecs:region:012345678910:task/arn")})
+
+	containerECS := &ecs.Container{}
+	containerECS.SetReason("CannotPullContainerError: API error (404): repository test not found: does not exist or no pull access\n")
+	containerECS.SetExitCode(0)
+
+	task := &ecs.Task{}
+	task.SetTaskArn("arn:aws:ecs:region:012345678910:task/arn")
+	task.SetTaskDefinitionArn("arn:aws:ecs:region:account:task-definition/dpl_id:deployVersion")
```
zpatrick: We don't care about the returned task's `TaskDefinitionArn` in this test, and I'm pretty sure we don't care about the `TaskArn` either.

api/provider/aws/test_aws/task_read_test.go
```diff
@@ -110,3 +110,96 @@ func TestTaskRead(t *testing.T) {
...
+		DescribeTasks(describeTaskInput).
+		Return(describeTaskOutput, nil)
+
+	target := provider.NewTaskProvider(mockAWS.Client(), tagStore, mockConfig)
+	result, err := target.Read("tsk_id")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	container := models.Container{
```
zpatrick: This section is asserting the entire model is correct, which is out of scope for this test. 
 For this test, we only care that the container's `ExitCode` was flipped from `0` to `1`. This whole section of of the test should be replaced with:
```go
    ...
    result, err := target.Read("tsk_id")
    if err != nil {
        t.Fatal(err)
    }

    assert.Len(t, 1, result.Containers)
    assert.Equal(t, 1, result.Containers[0].ExitCode)
}
```

api/provider/aws/test_aws/task_read_test.go
diemonster: `environment_id` is also required; I thought about removing the other tags but your original comment was ambiguous

api/provider/aws/test_aws/task_read_test.go
diemonster: `TaskDefinitionArn` must be set due to the way `Read()` is implemented. I opted to keep `SetTaskArn` as well so that param wasn't coming out of nowhere.

api/provider/aws/test_aws/task_read_test.go
diemonster: [nit] assert.Len(t, result.Containers, 1) :trollface:

zpatrick: I added the length assertion so we don't hit a `panic` if we try to do ` result.Containers[0].ExitCode` when the length is 0.

diemonster:
FYI I'm going to hold off on merging this until it's fully testable.

PR #525, opened 2018-01-26
498 Implement dynamic environments

### LINKED ISSUE #498: Implement Static and Dynamic Environments
### common/models
  * Remove/replace Environment model's MinScale, CurrentScale, MaxScale fields. 
  * Add EnvironmentType field
```go
type Environment struct {
    ...
    CurrentScale    int
    DesiredScale    int
    EnvironmentType string
    ...
}
```
* Remove/Replace scale field in CreateEnvironmentRequest, add EnvironmentType
```go
type CreateEnvironmentRequest struct {
    ...
    Scale           int
    EnvironmentType string
}
```
* Update CreateEnvironmentRequest.Validate() to make sure they didn't specify a Scale & dynamic EnvironmentType
* Update `EnvironmentEnvironmentRequest` to take `Scale int` as the only param. 

### cli/command
* `l0 environment create` should take `--scale` flag
  * If `--scale` is specified, then it is a static environment
  * If `--scale` is not specified, then it is a dynamic environment
* replace `setmincount` and `setmaxcount` with `l0 environment scale NAME SCALE`

### cli/printer
* Show `SCALE` and `TYPE` when printing environments
```
l0 environment get foo*
ID       ...   SCALE   TYPE
foo123   ...   3/3     static
foo457   ...   -       dynamic
```

### api/providers/aws
* EnvironmentCreate() should add a tag for `environment_type`. This is either `static` or `dynamic`. 
* We should probably add a const in this package for the two environment types:
```go
const (
    EnvironmentTypeStatic  = "static"
    EnvironmentTypeDynamic = "dynamic"
)
```
* EnvironmentCreate() and EnvironmentDelete() should only Create/Delete LaunchConfigs and ASGs if they are static environments. 
* INVESTIGATION: Does fargate work with Windows? 
* EnvironmentRead() on static environments should populate DesiredScale and CurrentScale
  * DesiredScale should be taken from the [ASG's DesiredCapacity](https://github.com/aws/aws-sdk-go/blob/master/service/autoscaling/api.go#L9206).
  * CurrentScale should be taken from the [Cluster's RegisteredContainerInstancesCount](https://github.com/aws/aws-sdk-go/blob/master/service/ecs/api.go#L3837)
* EnvironmentUpdate() should error on dynamic environments, set the Min, Desired & Max Capacity for the environment's ASG on static environments.

### plugins/terraform
* Add `environment_type` as a field for the environment resource
* Make sure to read the new field(s) on the environment models for both the environment resource and data source.

sesh-kebab:
Adds new environment_type attribute to environments where the value can be either dynamic (default) or static.

CLI: 
* Create environment with and without `--scale` flag to create static & dynamic environments
* Run environment list and get commands to ensure the environment output is as expected
* Run `set-scale` on dynamic and static environments and ensure output is as expected

Terraform:
* Create environment with and without `environment_type` attribute (value: 'static' or 'dynamic' [default])
* Update scale flag for both environment types
* Update environment_type 

**Checklist**
- [x] Unit tests
- [x] Smoke tests 
- [x] System tests 
- [x] Documentation

closes #498


api/provider/aws/environment_create.go
```diff
@@ -81,39 +49,69 @@ func (e *EnvironmentProvider) Create(req models.CreateEnvironmentRequest) (strin
...
-		aws.StringValue(securityGroup.GroupId),
-		instanceType,
-		e.Config.InstanceProfile(),
-		e.Config.SSHKeyPair(),
-		amiID,
-		userData); err != nil {
-		return "", err
-	}
+	// creating asg, lc isn't required for dynamic environments
+	if strings.ToLower(req.EnvironmentType) != models.EnvironmentTypeDynamic {
```
zpatrick: [nit] best practice is to use "positive" bool comparisons. It's also the more conservative path in this instance (e.g. if there were a third type of environment, we wouldn't necessarily want this block to execute).

api/provider/aws/environment_update.go
```diff
@@ -18,24 +19,23 @@ import (
...
+		environmentType := ""
+		if tag, ok := tags.WithKey("type").First(); ok {
+			environmentType = tag.Value
 		}
 
-		maxSize := aws.Int64Value(asg.MaxSize)
-		if req.MaxScale != nil {
-			maxSize = int64(*req.MaxScale)
+		if environmentType != models.EnvironmentTypeStatic {
+			return fmt.Errorf("updating dynamic environment's size isn't supported")
```
zpatrick: [nit] should this be `environment's scale`?

api/provider/aws/environment_update.go
```diff
@@ -110,11 +110,11 @@ func (e *EnvironmentProvider) Update(environmentID string, req models.UpdateEnvi
 	return nil
 }
 
-func (e *EnvironmentProvider) updateASGSize(autoScalingGroupName string, minSize, maxSize int64) error {
+func (e *EnvironmentProvider) updateASGSize(autoScalingGroupName string, size int) error {
 	input := &autoscaling.UpdateAutoScalingGroupInput{}
 	input.SetAutoScalingGroupName(autoScalingGroupName)
-	input.SetMinSize(minSize)
-	input.SetMaxSize(maxSize)
+	input.SetMinSize(int64(size))
```
zpatrick: Don't we need to set desired as well?

cli/command/environment.go
```diff
@@ -78,17 +73,13 @@ func (e *EnvironmentCommand) Command() cli.Command {
 			},
 			{
 				Name:      "set-scale",
```
zpatrick: [nit] would you consider using `scale` instead of `set-scale`? It's a personal preference, but I like that `scale` is consistent with the other single-verb commands, `list`, `delete`, etc.

cli/command/environment.go
```diff
@@ -78,17 +73,13 @@ func (e *EnvironmentCommand) Command() cli.Command {
...
-				Usage:     "update the min/max scale of an environment cluster",
+				Usage:     "update the scale of a static environment cluster",
 				Action:    e.setScale,
 				ArgsUsage: "ENVIRONMENT_NAME",
```
zpatrick: This should be `ENVIRONMENT_NAME SCALE` since the scale is a required arg, and flags denote optional params.

cli/command/environment_test.go
```diff
@@ -263,11 +261,9 @@ func TestEnvironmentSetScale(t *testing.T) {
 			Resolve("environment", "env_name").
 			Return([]string{"env_id"}, nil)
 
-		minScale := 2
-		maxScale := 5
+		Scale := 2
```
zpatrick: accidentally caps?

cli/command/environment_test.go
```diff
@@ -134,8 +133,8 @@ func TestCreateEnvironment(t *testing.T) {
 		req := models.CreateEnvironmentRequest{
 			EnvironmentName:  "env_name",
 			InstanceType:     "t2.small",
-			MinScale:         2,
-			MaxScale:         5,
+			Scale:            2,
+			EnvironmentType:  "static",
```
zpatrick: We will need a test to cover the "dynamic" code path.

cli/printer/text.go
```diff
@@ -74,7 +74,11 @@ func (t *TextPrinter) PrintDeploySummaries(deploys ...*models.DeploySummary) err
 
 func (t *TextPrinter) PrintEnvironments(environments ...*models.Environment) error {
 	getScale := func(e *models.Environment) string {
-		return fmt.Sprintf("%d:%d:%d", e.MinScale, e.CurrentScale, e.MaxScale)
+		if e.EnvironmentType != models.EnvironmentTypeDynamic {
+			return fmt.Sprintf("%d/%d", e.CurrentScale, e.DesiredScale)
+		}
+
+		return "n/a"
```
zpatrick: [nit] would displaying a `-` look better?

cli/printer/text.go
```diff
@@ -85,11 +89,12 @@ func (t *TextPrinter) PrintEnvironments(environments ...*models.Environment) err
 		return e.Links[i]
 	}
 
-	rows := []string{"ENVIRONMENT ID | ENVIRONMENT NAME | OS | SCALE | INSTANCE TYPE | LINKS"}
+	rows := []string{"ENVIRONMENT ID | ENVIRONMENT NAME | TYPE | OS | SCALE | INSTANCE TYPE | LINKS"}
```
zpatrick: [nit] I would argue that `INSTANCE_TYPE` should be dropped from this table - the table is getting pretty large and I think it is the least viewed/important field here.

common/config/api.go
```diff
@@ -176,7 +176,7 @@ func (c *ContextAPIConfig) Validate() error {
 
 	for _, name := range requiredVars {
 		if !c.C.IsSet(name) {
-			return fmt.Errorf("Required Variable %s is not set!", name)
+			return fmt.Errorf("Required Variable '%s' is not set!", name)
```
zpatrick: 👍 always appreciate small cleanup items like this

common/models/create_environment_request.go
```diff
@@ -21,16 +22,13 @@ func (r CreateEnvironmentRequest) Validate() error {
...
-	if r.MaxScale < 0 {
-		return fmt.Errorf("MaxScale must be a positive integer")
-	}
-
-	if r.MaxScale < r.MinScale {
-		return fmt.Errorf("MaxScale must be greater than or equal to MinScale")
+	if strings.EqualFold(r.EnvironmentType, EnvironmentTypeDynamic) &&
+		!strings.EqualFold(r.OperatingSystem, LinuxOS) {
+		return fmt.Errorf("%s is not a supported OS for dynamic environments", r.OperatingSystem)
 	}
```
zpatrick: don't we need to update the swagger def?

common/models/create_environment_request.go
```diff
@@ -21,16 +22,13 @@ func (r CreateEnvironmentRequest) Validate() error {
...
+		return fmt.Errorf("Scale must be a positive integer")
 	}
 
-	if r.MaxScale < 0 {
-		return fmt.Errorf("MaxScale must be a positive integer")
-	}
-
-	if r.MaxScale < r.MinScale {
-		return fmt.Errorf("MaxScale must be greater than or equal to MinScale")
+	if strings.EqualFold(r.EnvironmentType, EnvironmentTypeDynamic) &&
```
zpatrick: I'm not sure why we are using `strings.EqualFold` here; why isn't default comparison (`==`) sufficient?

common/models/create_environment_request.go
```diff
@@ -21,16 +22,13 @@ func (r CreateEnvironmentRequest) Validate() error {
...
+		!strings.EqualFold(r.OperatingSystem, LinuxOS) {
+		return fmt.Errorf("%s is not a supported OS for dynamic environments", r.OperatingSystem)
```
zpatrick: Seems like we are missing some other error scenarios:
* `EnvironmentType == EnvironmentTypeDynamic && Scale != 0`
* `!(EnvironmentType == EnvironmentTypeDynamic || EnvironmentType == EnvironmentTypeStatic)`

zpatrick: I realize that since `EnvironmentType` has a default, our pattern has been to allow empty strings in that field, so while it may make sense to not do that validation here, it still needs to be done somewhere. 
Also, we may want to reconsider allowing empty strings like that (the main reason I used that pattern is because I was impatient adding each field in swagger; definitely not the best reason to keep it around).

plugins/terraform/layer0/data_source_layer0_environment.go
```diff
@@ -77,9 +73,9 @@ func dataSourceLayer0EnvironmentRead(d *schema.ResourceData, meta interface{}) e
 	d.SetId(environment.EnvironmentID)
 	d.Set("name", environment.EnvironmentName)
 	d.Set("instance_type", environment.InstanceType)
-	d.Set("min_scale", environment.MinScale)
-	d.Set("current_scale", environment.CurrentScale)
-	d.Set("max_scale", environment.MaxScale)
+	d.Set("environment_type", environment.EnvironmentType)
+	d.Set("environment", environment.EnvironmentType)
```
zpatrick: line 77 seems like an accidental copy+paste and needs to be deleted `d.Set("environment", environment.EnvironmentType)`

tests/system/cases/windows_service/main.tf
```diff
@@ -7,7 +7,8 @@ provider "layer0" {
 resource "layer0_environment" "ws" {
   name = "ws"
   os   = "windows"
-  size = "m3.large"
+  instance_type = "m3.large"
```
zpatrick: this needs a scale of at least 1 to pass, right?

tests/system/system_test.go
```diff
@@ -22,15 +23,22 @@ func NewSystemTest(t *testing.T, dir string, vars map[string]string) *SystemTest
 	vars["endpoint"] = os.Getenv(config.ENVVAR_ENDPOINT)
 	vars["token"] = os.Getenv(config.ENVVAR_TOKEN)
 
+	fmt.Println("[DEBUG] endpoint", vars["endpoint"])
```
zpatrick: always use `log` package to log! Also, this should probably be cleaned up a bit using `Printf`

tests/system/system_test.go
```diff
@@ -22,15 +23,22 @@ func NewSystemTest(t *testing.T, dir string, vars map[string]string) *SystemTest
...
 		tftest.DryRun(*dry))
 
 	layer0 := clients.NewLayer0TestClient(t, vars["endpoint"], vars["token"])
 
-	// download modules using terraform get
-	tfContext.Terraformf("get")
+	// download modules using terraform init
+	stdoutStderr, err := tfContext.Terraformf("init")
+	if err != nil {
+		fmt.Println("[ERROR]", err)
```
zpatrick: Shouldn't we error immediately? Use `t.Fatal` for that.

tests/system/system_test.go
```diff
@@ -22,15 +23,22 @@ func NewSystemTest(t *testing.T, dir string, vars map[string]string) *SystemTest
...
+	}
+	fmt.Printf("%s\n", stdoutStderr)
```
zpatrick: seems like the printing I'm seeing here is leftover for live debugging more than actual code intended to stay?

api/provider/aws/environment_create.go
sesh-kebab: Not convinced about "positive" comparisons being best practice.

I actually ended up changing this from `== static` to `== dynamic` with the justification that existing behavior is the default behavior. Any new environment types should use the existing as the baseline. But I can see the argument the other way too.

api/provider/aws/environment_create.go
```diff
@@ -81,39 +49,69 @@ func (e *EnvironmentProvider) Create(req models.CreateEnvironmentRequest) (strin
...
+		}
 	}
 
 	clusterName := fqEnvironmentID
 	if err := e.createCluster(clusterName); err != nil {
 		return "", err
 	}
 
-	if err := e.createTags(environmentID, req.EnvironmentName, req.OperatingSystem); err != nil {
+	if err := e.createTags(environmentID, req.EnvironmentName, req.EnvironmentType, req.OperatingSystem); err != nil {
```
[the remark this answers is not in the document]
sesh-kebab: Was checking for an edge case in Validate but added the check for static & dynamic. ✅

api/provider/aws/environment_create_test.go
```diff
@@ -13,7 +13,7 @@ func TestEnvironment_createTags(t *testing.T) {
 	tagStore := tag.NewMemoryStore()
 	environment := NewEnvironmentProvider(nil, tagStore, nil)
 
-	if err := environment.createTags("env_id", "env_name", "env_os"); err != nil {
+	if err := environment.createTags("env_id", "env_name", "static", "env_os"); err != nil {
```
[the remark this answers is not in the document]
sesh-kebab: I would think it would be redundant in this case as you're testing the same code path in this test. `static` is essentially `env_type`.

api/provider/aws/environment_update.go
sesh-kebab: ✅

api/provider/aws/environment_update.go
sesh-kebab: We weren't doing so before. I can confirm the scaling behavior works as expected where desired is set to min without explicitly setting it.

cli/command/environment_test.go

cli/command/environment_test.go

common/config/api.go
sesh-kebab: ✅

common/models/create_environment_request.go
sesh-kebab: ✅

plugins/terraform/layer0/data_source_layer0_environment.go
sesh-kebab: ✅

tests/system/cases/windows_service/main.tf
sesh-kebab: Yes, you're correct. ✅

common/models/create_environment_request.go
sesh-kebab: for a case insensitive comparison.

common/models/create_environment_request.go
sesh-kebab: If the `EnvironmentType` is explicitly set, I'm ignoring `Scale` intentionally. But I can see the value in informing the user when this is the case. ✅ 
> EnvironmentType == EnvironmentTypeDynamic && Scale != 0 

If we do, then it would make sense to call Validate from the Terraform provider to right?

cli/command/environment.go

cli/command/environment.go

cli/printer/text.go
sesh-kebab: Don't feel strongly either way. Will change it to `-`. ✅

cli/printer/text.go
sesh-kebab: Agree that this is getting large. If we remove this, I think we should drop SCALE as well.

SCALE, INSTANCE_TYPE will be a lot less relevant going forward but would you we want the user to resort to showing the json output to check these fields?

tests/system/system_test.go
sesh-kebab: ✅ using testing.T.Logf

tests/system/system_test.go
sesh-kebab: ✅

tests/system/system_test.go
sesh-kebab: Was for debugging, but switched to using t.Logf

cli/printer/text.go
zpatrick: I like the idea of dropping SCALE as well since it makes static environments feel a little more "legacy" and discouraged. With that in mind, I think it's ok to require users's specify json output to see that.

common/models/create_environment_request.go
zpatrick: I think adding that validation in, and including it in terraform, would make our code better.

tests/system/system_test.go
zpatrick: Taking a step back, there is already a `tfContext.Init()` command: https://github.com/quintilesims/tftest/blob/master/test_context.go#L20

tests/system/system_test.go
zpatrick: You may want to consider using `log.Printf("[DEBUG] ...")` over `t.Logf()`, because they have slightly different behaviors:
* Any `log.Printf("[DEBUG] ...)` statements print if the user runs `go test --debug`. These statements print immediately. 
* Any `t.Logf` statements print if the user runs `go test -v`. These statements print only after the test has completed or failed.

tests/system/system_test.go
zpatrick: see my `t.Logf` vs `log.Printf("[DEBUG] ...` above

cli/printer/text.go
sesh-kebab: ✅

common/models/create_environment_request.go
sesh-kebab: ✅

api/provider/aws/environment_create.go
zpatrick: I can't find where I read this, but the argument goes something like this:

Obviously context matters, but when you have an either-or situation, it's better to use positive comparisons because that's how humans naturally communicate. For example, it's more natural to say "If Seshi is a male, then he has XY chromosomes" than it is to say "If Seshi is not a female, then he has XY chromosomes".

api/provider/aws/environment_create.go
```diff
@@ -81,39 +49,69 @@ func (e *EnvironmentProvider) Create(req models.CreateEnvironmentRequest) (strin
...
+		instanceType := config.DefaultEnvironmentInstanceType
+		if req.InstanceType != "" {
+			instanceType = req.InstanceType
+		}
+
+		if len(req.UserDataTemplate) > 0 {
+			userDataTemplate = req.UserDataTemplate
+		}
+
+		userData, err := RenderUserData(fqEnvironmentID, e.Config.S3Bucket(), userDataTemplate)
```
zpatrick: Will discuss in Slack, but for record keeping: how do we do private docker auth with dynamic environments?

api/provider/aws/environment_update.go
```diff
@@ -18,24 +19,23 @@ import (
...
+		environmentType := ""
+		if tag, ok := tags.WithKey("type").First(); ok {
+			environmentType = tag.Value
 		}
 
-		maxSize := aws.Int64Value(asg.MaxSize)
-		if req.MaxScale != nil {
-			maxSize = int64(*req.MaxScale)
+		if environmentType != models.EnvironmentTypeStatic {
+			return fmt.Errorf("updating dynamic environment's scale isn't supported")
```
zpatrick: I'm pretty sure there is an `InvalidRequest` error code that we could use for this, e.g. `return errors.New(errors.InvalidRequest, "updating...")`. That'll be helpful since that will let the `controller` pkg know to return a 400 status code rather than a 500.

cli/command/environment.go
```diff
@@ -34,22 +35,17 @@ func (e *EnvironmentCommand) Command() cli.Command {
...
-						Name:  "min-scale",
+						Name:  "scale",
 						Value: 0,
-						Usage: "minimum allowed scale of the environment cluster",
-					},
-					cli.IntFlag{
-						Name:  "max-scale",
-						Value: config.DefaultEnvironmentMaxScale,
-						Usage: "maximum allowed scale of the environment cluster",
+						Usage: "cluster size. setting scale explicitly, results in a static environment",
```
zpatrick: I think this could be cleaned up a bit, maybe something like `specifies the number of instances in the cluster. Setting this value will create a static environment`

cli/command/environment.go
```diff
@@ -139,13 +125,21 @@ func (e *EnvironmentCommand) create(c *cli.Context) error {
...
 		UserDataTemplate: userData,
 		OperatingSystem:  c.String("os"),
 		AMIID:            c.String("ami"),
 	}
 
+	if c.IsSet("scale") {
+		req.EnvironmentType = models.EnvironmentTypeStatic
+	}
+
+	if !c.IsSet("os") {
```
zpatrick: We don't need to do this check since we already have the default specified when we create the flag.

cli/command/environment.go
```diff
@@ -211,20 +205,15 @@ func (e *EnvironmentCommand) read(c *cli.Context) error {
...
-	if c.IsSet("min-scale") {
-		minScale := c.Int("min-scale")
-		req.MinScale = &minScale
-	}
-
-	if c.IsSet("max-scale") {
-		maxScale := c.Int("max-scale")
-		req.MaxScale = &maxScale
+	scale, err := strconv.Atoi(args["SCALE"])
+	if err == nil {
```
zpatrick: This seems a little odd to me, not sure why we wouldn't use standard error checking here?
```go
scale, err := strconv...
if err != nil { 
    return err
}

req := models.UpdateEnvironmentRequest{
    Scale: &scale,
}
...
```

cli/command/environment_test.go
```diff
@@ -56,12 +55,9 @@ func TestCreateEnvironmentInputErrors(t *testing.T) {
...
 	contexts := map[string]*cli.Context{
 		"Missing NAME arg": testutils.NewTestContext(t, nil, nil),
-		"Negative MinScale": testutils.NewTestContext(t,
+		"Negative Scale": testutils.NewTestContext(t,
 			[]string{"env_name"},
-			map[string]interface{}{"min-scale": "-1"}),
-		"Negative MaxScale": testutils.NewTestContext(t,
-			[]string{"env_name"},
-			map[string]interface{}{"max-scale": "-1"}),
+			map[string]interface{}{"scale": "-1"}),
```
zpatrick: I think it would be useful to add a non-integer case, e.g. `{"scale": "one"}`

tests/system/cases/windows_service/main.tf
```diff
@@ -7,7 +7,9 @@ provider "layer0" {
 resource "layer0_environment" "ws" {
   name = "ws"
   os   = "windows"
-  size = "m3.large"
+  scale = 1
```
zpatrick: [nit] you can format terraform files with `terraform fmt`

tests/system/system_test.go
```diff
@@ -22,15 +22,22 @@ func NewSystemTest(t *testing.T, dir string, vars map[string]string) *SystemTest
...
 		tftest.Dir(dir),
 		tftest.Vars(vars),
 		tftest.DryRun(*dry))
 
 	layer0 := clients.NewLayer0TestClient(t, vars["endpoint"], vars["token"])
 
-	// download modules using terraform get
-	tfContext.Terraformf("get")
+	// download modules using terraform init
+	stdoutStderr, err := tfContext.Terraformf("init")
```
zpatrick: I think we should just use `tfContext.Init()` here

tests/system/system_test.go
```diff
@@ -22,15 +22,22 @@ func NewSystemTest(t *testing.T, dir string, vars map[string]string) *SystemTest
 	vars["endpoint"] = os.Getenv(config.ENVVAR_ENDPOINT)
 	vars["token"] = os.Getenv(config.ENVVAR_TOKEN)
 
+	t.Logf("[DEBUG] endpoint: %s", vars["endpoint"])
```
zpatrick: We don't need the `[DEBUG]` prefix when logging using `*testing.T` since that only prints based on whether `go test -v` was passed. However, we do need them when logging using the `log` package, because that's how  we determine whether to [print statements or not](https://github.com/quintilesims/layer0/blob/api-refactor/common/logging/logging.go#L17).

[nit] Personally, I don't like having more than one way to log, but I feel it was/is better to use `log.Print` over `t.Log` in system tests (and only `log.Print` to keep things simple) given those logs print in realtime. The latter option forces you to wait until the test completes or errors before getting any feedback, which can be problematic given the length of time system tests can run.

cli/command/environment.go
sesh-kebab: I thought so too but when running the test, the default wasn't being picked up for some reason. I'll double check what the issue could be.

api/provider/aws/environment_create.go
sesh-kebab: Yes.. I raised this issue previously as well. Private repo auth isn't supported in fargate as far as I'm aware.

The only solution I can see is modifying the task definition ourselves to inject a reference to an ECR image. We would also have to ensure aws permissions are setup to allow cross account access to an ECR.

tests/system/system_test.go
sesh-kebab: ✅

tests/system/cases/windows_service/main.tf
sesh-kebab: ✅

cli/command/environment.go

cli/command/environment.go
zpatrick: ah, that's an issue with `testutils.NewTestContext`. There's some functionality that `urfave/cli` does to allow defaults that I haven't found out how to duplicate yet. Would be awesome if you could figure it out :wink:

cli/command/environment.go
sesh-kebab: That is odd - not sure how that happened 😕 
✅

cli/command/environment.go
sesh-kebab: I'll create a new issue for this. We're not testing the command correctly and not passing the args correctly either. I have a fix for this but we might as well update all the cli tests with the slightly tweaked pattern.
✅

cli/command/environment_test.go
sesh-kebab: Not sure I agree here. This wouldn't be testing our code; rather that a IntFlag will throw an error if an invalid int is passed in.

zpatrick: oh shoot, sorry I got test mixed up with `EnvironmentScale` where that's an arg, not a flag.

api/provider/aws/environment_create.go
```diff
@@ -81,39 +49,69 @@ func (e *EnvironmentProvider) Create(req models.CreateEnvironmentRequest) (strin
...
-		launchConfigName,
-		aws.StringValue(securityGroup.GroupId),
-		instanceType,
-		e.Config.InstanceProfile(),
-		e.Config.SSHKeyPair(),
-		amiID,
-		userData); err != nil {
-		return "", err
-	}
+	// creating asg, lc isn't required for dynamic environments
```
diemonster: [nit] for this comment, took a minute for me to realize lc == launch config 🤦‍♂️

api/provider/aws/environment_create.go
```diff
@@ -81,39 +49,69 @@ func (e *EnvironmentProvider) Create(req models.CreateEnvironmentRequest) (strin
...
+	if strings.ToLower(req.EnvironmentType) != models.EnvironmentTypeDynamic {
+		var userDataTemplate []byte
+		var amiID string
+
+		switch strings.ToLower(req.OperatingSystem) {
+		case models.LinuxOS:
```
diemonster: [nit] what if the user is using a FreeBSD ami? :trollface:

api/provider/aws/environment_create.go
diemonster: Glad this was brought up --- it might have been the only useful thing AWS shared w us on the Fargate call. They mentioned several things worth noting here: 

1) Fargate ONLY supports private repos via ECR
2) You have to use https://github.com/awslabs/amazon-ecr-credential-helper
3) ecr credential helper needs the following policies set, which I'm betting we need to add still in Layer0: https://docs.aws.amazon.com/AmazonECR/latest/userguide/ecr_managed_policies.html
3) Supplementary article on the ECR credential helper (sort of relevant) in this context: https://aws.amazon.com/blogs/compute/authenticating-amazon-ecr-repositories-for-docker-cli-with-credential-helper/

docs-src/docs/intro.md
```diff
@@ -33,4 +33,4 @@ Your running Layer0 applications. We also use the term `service` for tools such
 
 ### Environments
 
-Logical groupings of services. Typically, you would make a single environment for each tier of your application, such as `dev`, `staging`, and `prod`.
+Logical groupings of services. Typically, you would make a single environment for each tier of your application, such as `dev`, `staging`, and `prod`. Additionally an environment can be either static or dynamic. Static environments should be used when you require more fine grained control over an EC2 instance that a container will run on. Dynamic environments (the default) allow you to run containers without having to worry about managing and scaling clusters of EC2 instances.
```
diemonster: This is a good start, but I think we're going to need additional docs to explain this difference --- pros and cons of either approach in detail, etc. That can be a separate issue we create, though.

diemonster: added this issue for now https://github.com/quintilesims/layer0/issues/537

PR #533, opened 2018-02-01
[WIP] 500: Set launch type for services and tasks

### LINKED ISSUE #500: Set LaunchType when creating Services and Tasks

### api/provider/aws

tlake:
Services and tasks now check the environment's type (static or dynamic) and use that value to determine the type of instance (Fargate or EC2) on which they should be deployed.

Spin up services and tasks in both dynamic and static environments, and confirm that dynamic environments yield Fargate instances and that static environments yield EC2 instances. Also, run unit tests at least for the `provider` package.

**Notes:**
- Fargate is only available in `us-east-1` right now, so make sure that's set in `l0-setup`.
- I've pushed an API binary to dockerhub: specify `tlaketest` in `l0-setup`.
- There are some requirements for a task definition file to be Fargate-compatible. Probably the most notable are the following four parameters (defined at the same level as `"containerDefinitions"`):
```
    "requiresCompatibilities": [ "FARGATE" ], 
    "networkMode": "awsvpc", 
    "cpu": "256", 
    "memory": "512",
```
More info at https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html - in particular, note that there are specific values allowed for `"cpu"` and `"memory"`.

**Checklist**
- [x] Unit tests

closes #500

### DIFF
--- a/api/provider/aws/common.go
+++ b/api/provider/aws/common.go
@@ -12,6 +12,8 @@ import (
 	"github.com/aws/aws-sdk-go/service/ecs/ecsiface"
 	"github.com/aws/aws-sdk-go/service/elb"
 	"github.com/aws/aws-sdk-go/service/elb/elbiface"
+	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/aws/aws-sdk-go/service/iam/iamiface"
 	"github.com/quintilesims/layer0/api/tag"
 	"github.com/quintilesims/layer0/common/errors"
 	"github.com/quintilesims/layer0/common/models"
@@ -72,6 +74,21 @@ func getLaunchTypeFromEnvironmentID(store tag.Store, environmentID string) (stri
 	}
 
 	return "", fmt.Errorf("Could not find instance launch type for environment '%s'", environmentID)
+}
+
+func getRoleARNFromRoleName(iamapi iamiface.IAMAPI, roleName string) (string, error) {
+	input := &iam.GetRoleInput{}
+	input.SetRoleName(roleName)
+	if err := input.Validate(); err != nil {
+		return "", err
+	}
+
+	output, err := iamapi.GetRole(input)
+	if err != nil {
+		return "", err
+	}
+
+	return aws.StringValue(output.Role.Arn), nil
 }
 
 func lookupDeployIDFromTaskDefinitionARN(store tag.Store, taskDefinitionARN string) (string, error) {
--- a/api/provider/aws/common.go
+++ b/api/provider/aws/common.go
@@ -14,6 +14,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/elb/elbiface"
 	"github.com/quintilesims/layer0/api/tag"
 	"github.com/quintilesims/layer0/common/errors"
+	"github.com/quintilesims/layer0/common/models"
 )
 
 func describeLoadBalancer(elbapi elbiface.ELBAPI, loadBalancerName string) (*elb.LoadBalancerDescription, error) {
@@ -51,6 +52,26 @@ func describeTaskDefinition(ecsapi ecsiface.ECSAPI, taskDefinitionARN string) (*
 	}
 
 	return output.TaskDefinition, nil
+}
+
+func getLaunchTypeFromEnvironmentID(store tag.Store, environmentID string) (string, error) {
+	var launchType string
+	tags, err := store.SelectByTypeAndID("environment", environmentID)
+	if err != nil {
+		return "", err
+	}
+
+	if tag, ok := tags.WithKey("type").First(); ok {
+		if tag.Value == models.EnvironmentTypeDynamic {
+			launchType = ecs.LaunchTypeFargate
+		}
+
+		if tag.Value == models.EnvironmentTypeStatic {
+			launchType = ecs.LaunchTypeEc2
+		}
+	}
+
+	return launchType, nil
 }
 
 func lookupDeployIDFromTaskDefinitionARN(store tag.Store, taskDefinitionARN string) (string, error) {
--- a/api/provider/aws/common.go
+++ b/api/provider/aws/common.go
@@ -55,23 +55,23 @@ func describeTaskDefinition(ecsapi ecsiface.ECSAPI, taskDefinitionARN string) (*
 }
 
 func getLaunchTypeFromEnvironmentID(store tag.Store, environmentID string) (string, error) {
-	var launchType string
 	tags, err := store.SelectByTypeAndID("environment", environmentID)
 	if err != nil {
 		return "", err
 	}
 
 	if tag, ok := tags.WithKey("type").First(); ok {
-		if tag.Value == models.EnvironmentTypeDynamic {
-			launchType = ecs.LaunchTypeFargate
-		}
-
-		if tag.Value == models.EnvironmentTypeStatic {
-			launchType = ecs.LaunchTypeEc2
+		switch tag.Value {
+		case models.EnvironmentTypeDynamic:
+			return ecs.LaunchTypeFargate, nil
+		case models.EnvironmentTypeStatic:
+			return ecs.LaunchTypeEc2, nil
+		default:
+			return "", fmt.Errorf("Could not find instance launch type for environment '%s'", environmentID)
 		}
 	}
 
-	return launchType, nil
+	return "", fmt.Errorf("Could not find instance launch type for environment '%s'", environmentID)
 }
 
 func lookupDeployIDFromTaskDefinitionARN(store tag.Store, taskDefinitionARN string) (string, error) {
--- a/api/provider/aws/common.go
+++ b/api/provider/aws/common.go
@@ -67,7 +67,7 @@ func getLaunchTypeFromEnvironmentID(store tag.Store, environmentID string) (stri
 		case models.EnvironmentTypeStatic:
 			return ecs.LaunchTypeEc2, nil
 		default:
-			return "", fmt.Errorf("Could not find instance launch type for environment '%s'", environmentID)
+			return "", fmt.Errorf("Environment '%s' has invalid 'type' tag: '%s'.", environmentID, tag.Value)
 		}
 	}
 
--- a/api/provider/aws/common.go
+++ b/api/provider/aws/common.go
@@ -12,8 +12,6 @@ import (
 	"github.com/aws/aws-sdk-go/service/ecs/ecsiface"
 	"github.com/aws/aws-sdk-go/service/elb"
 	"github.com/aws/aws-sdk-go/service/elb/elbiface"
-	"github.com/aws/aws-sdk-go/service/iam"
-	"github.com/aws/aws-sdk-go/service/iam/iamiface"
 	"github.com/quintilesims/layer0/api/tag"
 	"github.com/quintilesims/layer0/common/errors"
 	"github.com/quintilesims/layer0/common/models"
@@ -74,21 +72,6 @@ func getLaunchTypeFromEnvironmentID(store tag.Store, environmentID string) (stri
 	}
 
 	return "", fmt.Errorf("Could not find instance launch type for environment '%s'", environmentID)
-}
-
-func getRoleARNFromRoleName(iamapi iamiface.IAMAPI, roleName string) (string, error) {
-	input := &iam.GetRoleInput{}
-	input.SetRoleName(roleName)
-	if err := input.Validate(); err != nil {
-		return "", err
-	}
-
-	output, err := iamapi.GetRole(input)
-	if err != nil {
-		return "", err
-	}
-
-	return aws.StringValue(output.Role.Arn), nil
 }
 
 func lookupDeployIDFromTaskDefinitionARN(store tag.Store, taskDefinitionARN string) (string, error) {
--- a/api/provider/aws/common_test.go
+++ b/api/provider/aws/common_test.go
@@ -3,10 +3,85 @@ package aws
 import (
 	"testing"
 
+	"github.com/aws/aws-sdk-go/service/ecs"
 	"github.com/quintilesims/layer0/api/tag"
 	"github.com/quintilesims/layer0/common/models"
 	"github.com/stretchr/testify/assert"
 )
+
+func Test_getLaunchTypeFromEnvironmentID(t *testing.T) {
+	tagStore := tag.NewMemoryStore()
+
+	envIDs := []string{"env_id0", "env_id1"}
+
+	tags := models.Tags{
+		{
+			EntityID:   envIDs[0],
+			EntityType: "environment",
+			Key:        "type",
+			Value:      models.EnvironmentTypeDynamic,
+		},
+		{
+			EntityID:   envIDs[1],
+			EntityType: "environment",
+			Key:        "type",
+			Value:      models.EnvironmentTypeStatic,
+		},
+	}
+
+	for _, tag := range tags {
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	cases := map[string]string{
+		envIDs[0]: ecs.LaunchTypeFargate,
+		envIDs[1]: ecs.LaunchTypeEc2,
+	}
+
+	for id, expected := range cases {
+		result, err := getLaunchTypeFromEnvironmentID(tagStore, id)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		assert.Equal(t, expected, result)
+	}
+}
+
+func Test_getLaunchTypeFromEnvironmentID_Errors(t *testing.T) {
+	tagStore := tag.NewMemoryStore()
+
+	envIDs := []string{"env_id0", "env_id1"}
+
+	tags := models.Tags{
+		{
+			EntityID:   envIDs[0],
+			EntityType: "environment",
+			Key:        "type",
+			Value:      "",
+		},
+		{
+			EntityID:   envIDs[1],
+			EntityType: "environment",
+			Key:        "",
+			Value:      "",
+		},
+	}
+
+	for _, tag := range tags {
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	for _, id := range envIDs {
+		if _, err := getLaunchTypeFromEnvironmentID(tagStore, id); err == nil {
+			t.Fatal("Err was nil!")
+		}
+	}
+}
 
 func Test_lookupDeployIDFromTaskDefinitionARN(t *testing.T) {
 	tagStore := tag.NewMemoryStore()
--- a/api/provider/aws/common_test.go
+++ b/api/provider/aws/common_test.go
@@ -52,7 +52,7 @@ func Test_getLaunchTypeFromEnvironmentID(t *testing.T) {
 	}
 }
 
-func Test_getLaunchTypeFromEnvironmentID_Errors(t *testing.T) {
+func Test_getLaunchTypeFromEnvironmentID_errors(t *testing.T) {
 	tagStore := tag.NewMemoryStore()
 
 	envIDs := []string{"env_id0", "env_id1", "env_id2"}
--- a/api/provider/aws/common_test.go
+++ b/api/provider/aws/common_test.go
@@ -41,19 +41,21 @@ func Test_getLaunchTypeFromEnvironmentID(t *testing.T) {
 	}
 
 	for id, expected := range cases {
-		result, err := getLaunchTypeFromEnvironmentID(tagStore, id)
-		if err != nil {
-			t.Fatal(err)
-		}
+		t.Run(id, func(t *testing.T) {
+			result, err := getLaunchTypeFromEnvironmentID(tagStore, id)
+			if err != nil {
+				t.Fatal(err)
+			}
 
-		assert.Equal(t, expected, result)
+			assert.Equal(t, expected, result)
+		})
 	}
 }
 
 func Test_getLaunchTypeFromEnvironmentID_Errors(t *testing.T) {
 	tagStore := tag.NewMemoryStore()
 
-	envIDs := []string{"env_id0", "env_id1"}
+	envIDs := []string{"env_id0", "env_id1", "env_id2"}
 
 	tags := models.Tags{
 		{
@@ -64,6 +66,12 @@ func Test_getLaunchTypeFromEnvironmentID_Errors(t *testing.T) {
 		},
 		{
 			EntityID:   envIDs[1],
+			EntityType: "environment",
+			Key:        "type",
+			Value:      "neither static nor dynamic",
+		},
+		{
+			EntityID:   envIDs[2],
 			EntityType: "environment",
 			Key:        "",
 			Value:      "",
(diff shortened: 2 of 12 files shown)


api/provider/aws/task_create.go
```diff
@@ -16,6 +16,22 @@ func (t *TaskProvider) Create(req models.CreateTaskRequest) (string, error) {
 	taskID := entityIDGenerator(req.TaskName)
 	fqEnvironmentID := addLayer0Prefix(t.Config.Instance(), req.EnvironmentID)
 
+	var launchType string
+	tags, err := t.TagStore.SelectByTypeAndID("environment", req.EnvironmentID)
+	if err != nil {
+		return "", err
+	}
+
+	if tag, ok := tags.WithKey("type").First(); ok {
```
zpatrick: seems like this should be a helper function that both `service_create` and `task_create` can call

zpatrick: Also, we should handle if the `type` tag doesn't exist and throw a meaningful error there. Currently, I assume we'd get some error from AWS and it wouldn't make any sense to the user what was going on.

api/provider/aws/common.go
```diff
@@ -53,6 +54,26 @@ func describeTaskDefinition(ecsapi ecsiface.ECSAPI, taskDefinitionARN string) (*
...
 
+func getLaunchTypeFromEnvironmentID(store tag.Store, environmentID string) (string, error) {
+	var launchType string
+	tags, err := store.SelectByTypeAndID("environment", environmentID)
+	if err != nil {
+		return "", err
+	}
+
+	if tag, ok := tags.WithKey("type").First(); ok {
+		if tag.Value == models.EnvironmentTypeDynamic {
```
zpatrick: I think we should throw an error if we don't find the environment's `type` tag. 
Maybe a more conservative approach here could be something like:
```go
if tag, ok := tags.WithKey("type").First(); ok {
    switch tag.Value {
        case models.EnvironmentTypeDynamic:
            return ecs.LaunchTypeFargate, nil
        case models.EnvironmentTypeStatic:
            return ecs.LaunchTypeEc2, nil
        default:
            return "", fmt.Errorf("...")
    }
}

return "", fmt.Errorf("...")
```

tlake: @zpatrick Oh good - I've been thinking about this recently, actually. I was going back and forth between erroring out or having a default. Do we have a way to retroactively generate environment type tags for existing environments when a user updates?

diemonster: we do have a way to do that, don't we? We have an API for tags, so we can create new arbitrary tags as needed.

zpatrick: No we don't retroactively add tags; keep in mind this is a minor version change so we don't need to worry about backwards compatibility

api/provider/aws/test_aws/service_create_test.go
```diff
@@ -104,6 +110,7 @@ func TestServiceCreate(t *testing.T) {
 	createServiceInput := &ecs.CreateServiceInput{}
 	createServiceInput.SetCluster("l0-test-env_id")
 	createServiceInput.SetDesiredCount(1)
+	createServiceInput.SetLaunchType(ecs.LaunchTypeFargate)
```
diemonster: I believe we also need to set the networking mode to `awsvpc`, which is the only networking mode that works w Fargate.

diemonster: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ECS_AWSCLI_Fargate.html as a point of reference

api/provider/aws/test_aws/task_create_test.go
```diff
@@ -72,6 +78,7 @@ func TestTaskCreate(t *testing.T) {
 
 	runTaskInput := &ecs.RunTaskInput{}
 	runTaskInput.SetCluster("l0-test-env_id")
+	runTaskInput.SetLaunchType(ecs.LaunchTypeFargate)
```
diemonster: see comment re: setting networking mode

api/provider/aws/test_aws/service_create_test.go
tlake: I can't find a `SetNetworkingMode` method anywhere (it looks like `networkMode` is a field in a task definition file), but there _does_ exist `SetNetworkConfiguration`, which is talked about being set for tasks/services that _use_ the `awsvpc` networking mode. From what I can tell, a Fargate instance automatically/by-default uses the `awsvpc` networking mode. I haven't yet figured out what (if anything) needs to be set for the network configuration, though.

In any event, it looks like I have some more research to do. :+1:

diemonster: One way to test is to just create a sample fargate service / task and see if ECS renders it with the `awsvpc` networkMode option in the resulting task definition in the ECS UI. Maybe it doesn't need to be defined here explicitly?

more details on `networkMode`: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#network_mode

more details on `awsvpc` mode, specifically:
https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-networking.html

tlake: ✅

api/provider/aws/common.go
```diff
@@ -53,6 +54,26 @@ func describeTaskDefinition(ecsapi ecsiface.ECSAPI, taskDefinitionARN string) (*
...
+	}
+
+	if tag, ok := tags.WithKey("type").First(); ok {
+		switch tag.Value {
+		case models.EnvironmentTypeDynamic:
+			return ecs.LaunchTypeFargate, nil
+		case models.EnvironmentTypeStatic:
+			return ecs.LaunchTypeEc2, nil
+		default:
+			return "", fmt.Errorf("Could not find instance launch type for environment '%s'", environmentID)
```
zpatrick: [nit] would be useful to show what the tag is, e.g. `"Environment '%s' has invalid 'type' tag: %s" environmentID, tag.Value)`

api/provider/aws/common_test.go
```diff
@@ -3,11 +3,86 @@ package aws
...
+			EntityID:   envIDs[0],
+			EntityType: "environment",
+			Key:        "type",
+			Value:      "",
+		},
+		{
+			EntityID:   envIDs[1],
+			EntityType: "environment",
+			Key:        "",
+			Value:      "",
```
zpatrick: we should add a case for an unknown value for the `type` tag

api/provider/aws/common_test.go
```diff
@@ -3,11 +3,86 @@ package aws
...
+	}
+
+	for _, tag := range tags {
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	for _, id := range envIDs {
+		if _, err := getLaunchTypeFromEnvironmentID(tagStore, id); err == nil {
```
zpatrick: [nit] we typically run each case in a table test as a subtest using [t.Run](https://golang.org/pkg/testing/#T.Run)

api/provider/aws/common_test.go
```diff
@@ -3,11 +3,86 @@ package aws
...
+		}
+	}
+
+	cases := map[string]string{
+		envIDs[0]: ecs.LaunchTypeFargate,
+		envIDs[1]: ecs.LaunchTypeEc2,
+	}
+
+	for id, expected := range cases {
+		result, err := getLaunchTypeFromEnvironmentID(tagStore, id)
```
zpatrick: [nit] we typically run each case in a table test as a subtest using [t.Run](https://golang.org/pkg/testing/#T.Run)

api/provider/aws/deploy_create.go
```diff
@@ -47,6 +47,46 @@ func (d *DeployProvider) createTaskDefinition(taskDefinition *ecs.TaskDefinition
...
+		//
+		// For the most part, there are four things that should be specified in a Fargate task definition
+		// that aren't required for an EC2 task definition:
+		//   "requiresCompatibilities": [ "FARGATE" ],
+		//   "networkMode": "awsvpc",
+		//   "cpu": "<an appropriate value>",
+		//   "memory": "<an appropriate value>"
+		if aws.StringValue(compatibility) == ecs.LaunchTypeFargate {
+			cpu := aws.StringValue(taskDefinition.Cpu)
+			memory := aws.StringValue(taskDefinition.Memory)
```
zpatrick: so there is memory and memory reservation; if memory is not set, should we check for memory reservation and use that if it is set?

api/provider/aws/deploy_create.go
```diff
@@ -47,6 +47,46 @@ func (d *DeployProvider) createTaskDefinition(taskDefinition *ecs.TaskDefinition
...
+
+			// Question: Do we want to be proactive here and error for certain configurations of a task
+			// definition that we know to be bad? They would come out of AWS when the user tried to use the
+			// deploy in a service or a task, but it might be nice UX to save them a step.
```
zpatrick: 👍 very much agree w this logic

api/provider/aws/deploy_create.go
```diff
@@ -47,6 +47,46 @@ func (d *DeployProvider) createTaskDefinition(taskDefinition *ecs.TaskDefinition
...
+
+			// hard-coding an ARN isn't usually the best thing, but we know exactly what this ARN will look like
+			// and the only variable in it is the account ID
+			accountID := d.Config.AccountID()
+			ecsTaskExecutionRoleARN := fmt.Sprintf("arn:aws:iam::%s:role/ecsTaskExecutionRole", accountID)
+			input.SetExecutionRoleArn(ecsTaskExecutionRoleARN)
+		}
+	}
+
+	input.SetRequiresCompatibilities(requiresCompatibilities)
```
zpatrick: I'm not sure why we're recreating `requiredCompatibilities` instead of just updating in place?

api/provider/aws/service_create.go
```diff
@@ -12,6 +12,20 @@ func (s *ServiceProvider) Create(req models.CreateServiceRequest) (string, error
 	fqEnvironmentID := addLayer0Prefix(s.Config.Instance(), req.EnvironmentID)
 	cluster := fqEnvironmentID
 
+	var securityGroupIDs []*string
+	environmentSecurityGroupName := getEnvironmentSGName(fqEnvironmentID)
+	environmentSecurityGroup, err := readSG(s.AWS.EC2, environmentSecurityGroupName)
```
zpatrick: [nit] isn't this only required if the `launchType` is fargate? If so, it would be a little more efficient to do this only in that case.

api/provider/aws/service_create.go
```diff
@@ -58,13 +82,22 @@ func (s *ServiceProvider) Create(req models.CreateServiceRequest) (string, error
 		scale = 1
 	}
 
+	privateSubnets := s.Config.PrivateSubnets()
```
zpatrick: Rule of Thumb! Don't let aws-sdk-go's ugliness leak out of helper functions (when you can help it). The helper should take `[]string` instead of `[]*string`, and manage the ugly conversions within itself.

api/provider/aws/service_create.go
```diff
@@ -77,17 +110,33 @@ func (s *ServiceProvider) Create(req models.CreateServiceRequest) (string, error
...
+
+	if launchType == ecs.LaunchTypeFargate {
+		awsvpcConfig := &ecs.AwsVpcConfiguration{}
+		awsvpcConfig.SetAssignPublicIp(ecs.AssignPublicIpDisabled)
+		awsvpcConfig.SetSecurityGroups(securityGroupIDs)
+		awsvpcConfig.SetSubnets(subnets)
+		networkConfig := &ecs.NetworkConfiguration{}
+		networkConfig.SetAwsvpcConfiguration(awsvpcConfig)
+		input.SetNetworkConfiguration(networkConfig)
+		input.SetPlatformVersion("LATEST")
```
zpatrick: [nit] can't help but call out whenever I see "latest" vs a hardcoded version; is this the best option here?

api/provider/aws/task_create.go
```diff
@@ -27,7 +41,21 @@ func (t *TaskProvider) Create(req models.CreateTaskRequest) (string, error) {
 	taskDefinitionVersion := deployVersion
 	taskOverrides := convertContainerOverrides(req.ContainerOverrides)
 
-	task, err := t.runTask(clusterName, startedBy, taskDefinitionFamily, taskDefinitionVersion, taskOverrides)
+	privateSubnets := t.Config.PrivateSubnets()
+	subnets := make([]*string, len(privateSubnets))
```
zpatrick: ditto

api/provider/aws/test_aws/environment_create_test.go
```diff
@@ -261,30 +260,6 @@ func TestEnvironmentCreateDefaults(t *testing.T) {
 		AuthorizeSecurityGroupIngress(gomock.Any()).
 		Return(&ec2.AuthorizeSecurityGroupIngressOutput{}, nil)
 
-	// ensure we pass the default instance type, ami id, and user data to the launch configuration
```
zpatrick: ah! no! This behavior hasn't been removed so we should not remove/reduce this test. Instead, we should add a new test to test the new behavior.

api/provider/aws/common.go

api/provider/aws/test_aws/environment_create_test.go
tlake: Ah, I had mistakenly thought that this was duplicated in an earlier test; this functionality should be moved to a different test then, since the new default is to use a dynamic (Fargate) environment and that doesn't create LauncConfigurations or AutoScalingGroups.

api/provider/aws/common_test.go
tlake: Ooh, good catch - I forgot about `t.Run`. Thanks! ✅

api/provider/aws/common_test.go
tlake: ✅

api/provider/aws/common_test.go
tlake: ✅

api/provider/aws/deploy_create.go
tlake: `memoryReservation` is a container-level parameter, not a task-definition-level parameter. Specifying `memoryReservation` at the task-definition level at all should be an error, just like any other parameter that doesn't exist.

If you wanted to get fancy, we could inspect the container definitions, compile all the `memory` and `memoryReservation` specs, and do some maths to figure out what a proper `memory` value should be, but that seems like a lot of hidden magic and I'd prefer to avoid it.

api/provider/aws/deploy_create.go
tlake: Huh, me neither. xD ✅

api/provider/aws/service_create.go
tlake: At the moment, it seems like the only platform version available is `"1.0.0"`, and that `"LATEST"` is currently synonymous; using "LATEST" here was just a holdover from following guides and just getting something to work; it's probably safer to manage this value ourselves, and I think it's probably a good idea to set this in the config somewhere instead of setting it here.

https://docs.aws.amazon.com/AmazonECS/latest/developerguide/platform_versions.html

api/provider/aws/task_create.go

api/provider/aws/common_test.go
```diff
@@ -3,11 +3,96 @@ package aws
...
 )
 
+func Test_getLaunchTypeFromEnvironmentID(t *testing.T) {
+	tagStore := tag.NewMemoryStore()
+
+	envIDs := []string{"env_id0", "env_id1"}
+
+	tags := models.Tags{
+		{
+			EntityID:   envIDs[0],
```
nehayward: I'm not sure we do this pattern anywhere else. Maybe just use 
`env_id1, env_id2`
[example](https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/test_aws/environment_list_test.go#L28)

api/provider/aws/common_test.go
```diff
@@ -3,11 +3,96 @@ package aws
...
+	}
+
+	for _, tag := range tags {
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	cases := map[string]string{
+		envIDs[0]: ecs.LaunchTypeFargate,
```
nehayward: Here too.

api/provider/aws/common_test.go
```diff
@@ -3,11 +3,96 @@ package aws
...
+
+			assert.Equal(t, expected, result)
+		})
+	}
+}
+
+func Test_getLaunchTypeFromEnvironmentID_Errors(t *testing.T) {
+	tagStore := tag.NewMemoryStore()
+
+	envIDs := []string{"env_id0", "env_id1", "env_id2"}
```
nehayward: Same.

api/provider/aws/deploy_create.go
diemonster: +1 for avoiding magic w `memory` and `memoryReservation`. AWS should throw an error if either of the params are left out. We should force the users to be explicit with those values.

api/provider/aws/task_create.go
```diff
@@ -65,12 +97,39 @@ func convertContainerOverrides(overrides []models.ContainerOverride) *ecs.TaskOv
...
+	if launchType == ecs.LaunchTypeFargate {
+		s := make([]*string, len(subnets))
+		for i := range subnets {
+			s[i] = aws.String(subnets[i])
+		}
+
+		awsvpcConfig := &ecs.AwsVpcConfiguration{}
+		awsvpcConfig.SetAssignPublicIp(ecs.AssignPublicIpDisabled)
+		awsvpcConfig.SetSecurityGroups(securityGroupIDs)
+		awsvpcConfig.SetSubnets(s)
```
nehayward: [nit pickity picky] newline here maybe?

api/provider/aws/test_aws/environment_create_test.go
```diff
@@ -257,6 +257,50 @@ func TestEnvironmentCreateDefaults(t *testing.T) {
...
+		CreateCluster(gomock.Any()).
+		Return(&ecs.CreateClusterOutput{}, nil)
+
+	target := provider.NewEnvironmentProvider(mockAWS.Client(), tagStore, mockConfig)
+	if _, err := target.Create(req); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestEnvironmentCreateDefaults_static(t *testing.T) {
```
nehayward: Is there a dynamic/stateless test?

api/provider/aws/common_test.go
tlake: @nehayward Yeah, I also don't think we do the pattern anywhere else. I introduced it looking for some feedback on whether we like this micropattern; I can see it not providing a benefit in `Test_getLaunchTypeFromEnvironmentID()` beyond a mild maintenance streamline wherein updates only have to be done in one place (and maybe that's benefit enough to merit keeping the pattern?), but I _do_ think there's a good reason for the micropattern in `Test_getLaunchTypeFromEnvironmentID_errors()` when we `range` over the `envIDs`.

Happy to strip it out if it's overengineering, though. :)

api/provider/aws/task_create.go

api/provider/aws/test_aws/environment_create_test.go
tlake: Yeah - the default behavior of `environment create` in this new release will be to create dynamic/stateless/fargate environments, so `TestEnvironmentCreateDefaults()` is the dynamic counterpart to this `TestEnvironmentCreateDefaults_static()` func. Naming things is hard, though - which is better?
- `TestEnvironmentCreateDefaults()`, which indicates calling `environment create` with the minimum required arguments, and which only _implies_ that it creates defaults for dynamic/stateless environments by virtue of dynamic/stateless environments being the default for `environment create`?
- `TestEnvironmentCreateDefaults_dynamic()`, which is explicit about creating the defaults for dynamic/stateless environments, but which doesn't communicate that it's also the generalized default for `environment create`?

api/provider/aws/common_test.go
```diff
@@ -3,11 +3,96 @@ package aws
...
+		},
+	}
+
+	for _, tag := range tags {
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	for _, id := range envIDs {
```
nehayward: Based off your comment I think it makes sense here, because you range over the array. I'd just change it to env_id1 instead of env_id0.

api/provider/aws/common_test.go
nehayward: I think it makes sense in the `Test_getLaunchTypeFromEnvironmentID_Errors` but we should revert to the old pattern for consistency in Test_getLaunchTypeFromEnvironmentID. Just my opinion though.

tlake: I like it. ✅

api/provider/aws/common_test.go
tlake: ✅ (https://github.com/quintilesims/layer0/pull/533#discussion_r170091137)

api/provider/aws/common_test.go
tlake: ✅ (https://github.com/quintilesims/layer0/pull/533#discussion_r170091137)

api/provider/aws/common_test.go
tlake: ✅

api/provider/aws/deploy_create.go
tlake: It should probably be noted that there exists a `memory` parameter at _both_ the task-definition-and container-definition-levels. Here's a summary of the relevant parts of the documentation at https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#task_size via a sample task definition file:

```
// task definition file
{
  // `memory` at the task-definition-level (TDL):
  // - is a string
  // - is required for Fargate instances, but optional for EC2 instances
  // - is coupled to the TDL `cpu` parameter in terms of valid values
  "memory": "512",
  "cpu": "256",
  "containerDefinitions": [
    {
      // `memory` at the container-definition-level (CDL):
      // - is an int
      // - is optional for both Fargate and EC2 instances, provided that
      //   at least one of `memory` and `memoryReservation` is supplied;
      //   if both are supplied, `memory` must be greater than
      //   `memoryReservation`.
      "memory": 256,
      "memoryReservation": 128,
    }
  ]
}
```

In terms of proactive inspection, `deploy_create.go` currently only checks that neither TDL-`memory` nor TDL-`cpu` are omitted. Given the summary above, do we want to do more? I could definitely write a more robust inspector that validates task- and container-definitions to those commented specifications, and while that will duplicate some of AWS' validation, it would also provide feedback to the user at the `deploy` step instead of later at the `service` step (which was the point of validating TDL-`memory` and -`cpu`).

api/provider/aws/deploy_create.go
tlake: (This is relevant to [this comment](https://github.com/quintilesims/layer0/pull/533#discussion_r170122962), just above.)

api/provider/aws/deploy_create.go
diemonster: We should not couple `deploy`s to `launchType`s, which I think we would invariably end up doing if we add extra 'helper' logic to `deploy create`. 

Aside from that, my vote is to let AWS handle the errors and leave it at that. I know that is more minimalistic than we do in other areas of the code, but given how new Fargate is and how many changes I can foresee it having in the future, it seems like a lot less maintenance to defer to AWS for errors and not try to be too clever in advance.

api/provider/aws/service_create.go
zpatrick: I think keeping it as-is is fine - it's maybe not perfectly in-line with the current pattern, but doesn't seem like a big deal to me.

tlake:
I'm closing this pull request in favor of the implementation in PR #567. We can reopen this PR in the event that this path is deemed to be the correct one for us to pursue.

PR #526, opened 2018-01-26
524: Apply Terraform SSL Cert ARN feature to api-refactor

jparsons04:
**What does this pull request do?**

With the release of v0.10.5, the feature that allows load balancer resources to have the ARN of an IAM or an ACM certificate needs to be folded into api-refactor

**How should this be tested?**
Through the CLI and terraform. Unit tests should pass.

**Checklist**
- [x] Unit tests
~- [ ] Smoke tests (if applicable)~
~- [ ] System tests (if applicable)~
~- [ ] Documentation (if applicable)~

closes #524

PR #534, opened 2018-02-01
Set the LoadBalancer's port's CertificateARN 

### LINKED ISSUE #531: Load Balancer TF Resource unable to read its state correctly when a cert is specified
### Expected behavior
Consecutive Updates to a Load Balancer Resource (via terraform) where you specify a Cert, results in the resource being recreated every time.

**sample terraform**
```
data "aws_acm_certificate" "oce" {
  provider = "aws.certificate"
  domain   = "*.oce-marketing.com"
  statuses = ["ISSUED"]
}

resource "layer0_load_balancer" "haproxy" {
  name        = "${var.haproxy_environment_name}-lb"
  environment = "${layer0_environment.haproxy.id}"
  private     = false

  port {
    host_port      = 443
    container_port = 80
    protocol       = "https"
    certificate    = "${data.aws_acm_certificate.oce.arn}"
  }
}
```

We seem to be converting the cert arn to name when performing a read. On initial thought, this could cause tf to register a change as there would be a mismatch between the tfstate and what load balancer read returns as far as LoadBalancer.Ports[*].Certificate is concerned.

https://github.com/quintilesims/layer0/blob/aaf3f886e51d4c16d4a96dfe6ead544deb67d616/api/backend/ecs/load_balancer_manager.go#L223

- `l0 admin version` and `l0 --version`
v0.10.5

linked to #491

sesh-kebab:
It sets a CertificateARN property on a port as well as the CertificateName. The TF Provider will use the ARN to populate its `certificate` property in the schema. 

Use the sample terraform config below
```
  name = "a"
}

resource "layer0_load_balancer" "haproxy" {
  name        = "test-lb"
  private     = false

  port {
    host_port      = 443
    container_port = 80
    protocol       = "https"
    certificate    = "<redacted>"
  }
}
```

Apply the changes using `terraform apply`

Confirm that a consecutive apply won't detect any changes and try to recreate any resources.

**Checklist**
- ~[ ] Unit tests~

closes #531

### DIFF
--- a/api/backend/ecs/load_balancer_manager.go
+++ b/api/backend/ecs/load_balancer_manager.go
@@ -220,6 +220,7 @@ func (e *ECSLoadBalancerManager) listenerToPort(listener *awselb.Listener) model
 	}
 
 	if listener.SSLCertificateId != nil {
+		port.CertificateARN = *listener.SSLCertificateId
 		port.CertificateName = id.CertificateARNToName(*listener.SSLCertificateId)
 	}
 


api/backend/ecs/load_balancer_manager.go
```diff
@@ -220,6 +220,7 @@ func (e *ECSLoadBalancerManager) listenerToPort(listener *awselb.Listener) model
 	}
 
 	if listener.SSLCertificateId != nil {
+		port.CertificateARN = *listener.SSLCertificateId
```
zpatrick: Are we sure this value will be the ACM certificates ARN and not it's ID? Also, don't we need some sort of update to terraform to ensure these values sync up (or is purposely omitted?)

sesh-kebab: According to the documentation, yes. https://docs.aws.amazon.com/sdk-for-go/api/service/elb/#Listener

> The Amazon Resource Name (ARN) of the server certificate.

L0 Provider doesn't require any changes.

diemonster:
@sesh-kebab is the intention here to release another layer0 hotfix? Does this also need to be applied to the `api-refactor` branch?

sesh-kebab:
@diemonster yes, I think this should be released as another hotfix. Also yes, it does need to be ported to the `api-refactor` branch also.

jparsons04:
I addressed the api-refactor port in my PR https://github.com/quintilesims/layer0/pull/526. It appears the code in API refactor already does this when it needs to read from the load balancer. See here: https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/load_balancer_read.go#L32

sesh-kebab:
@jparsons04 this isn't an issue in your PR. The port model only has Certificate in `api-refactor`. In Develop right now, we have both CertificateName & CertificateARN and the read in the provider was only setting the CertificateName which is what gave rise to the issue.

PR #539, opened 2018-02-02
Refactor cli tests

### LINKED ISSUE #536: CLI Command tests need to be refactored
### Problem
The tests for commands don't correctly use the Command to be tested. This results in defaults not being available within tests among other things.

### Solution
```go
// instead of:
...
c := testutils.NewTestContext(t, []string{"env_name"}, flags)
if err := command.create(c); err != nil {
	t.Fatal(err)
}

// do this
command := NewEnvironmentCommand(base.Command()).Command()
...
c := testutils.NewTestContext(t, []string{"environment", "create", "env_name"}, flags)
if err := command.Run(c); err != nil {
	t.Fatal(err)
}
```

This will require modifying NewTestContext to add arguments correctly.

- `l0 admin version` and `l0 --version`
v0.10.5

zpatrick:
closes #536


cli/command/environment_test.go
```diff
@@ -1,17 +1,17 @@
...
 	"github.com/quintilesims/layer0/common/models"
 	"github.com/quintilesims/layer0/common/testutils"
-	"github.com/urfave/cli"
 )
 
 func TestCreateEnvironment(t *testing.T) {
 	base, ctrl := newTestCommand(t)
 	defer ctrl.Finish()
-	command := NewEnvironmentCommand(base.Command())
+	command := NewEnvironmentCommand(base.CommandBase()).Command()
```
nehayward: Is there a way to reduce this? Seems a little verbose/ reads weird?

sesh-kebab: It does look/read weird but if you look at the source it will make sense.

cli/command/service_test.go
```diff
@@ -86,49 +72,35 @@ func TestDeleteService(t *testing.T) {
 		DeleteService("svc_id").
 		Return(nil)
 
-	c := testutils.NewTestContext(t, []string{"svc_name"}, nil)
-	if err := command.delete(c); err != nil {
+	if err := testutils.RunApp(command, "l0 service delete svc_name"); err != nil {
```
nehayward: Should we do all inputs, for consistency? Makes it easier to read in my opinion.
```
input := "l0 service delete svc_name"
```

cli/command/service_test.go
```diff
@@ -138,34 +110,21 @@ func TestReadService(t *testing.T) {
...
 		Return(&models.Service{}, nil)
 
-	c := testutils.NewTestContext(t, []string{"svc_name"}, nil)
-	if err := command.read(c); err != nil {
+	if err := testutils.RunApp(command, "l0 service get svc_name"); err != nil {
 		t.Fatal(err)
 	}
 }
 
 func TestReadServiceInputErrors(t *testing.T) {
```
nehayward: Should we change get to read? for the cli?

cli/command/task_test.go
```diff
@@ -182,34 +146,16 @@ func TestReadTaskLogs(t *testing.T) {
...
 
-	flags := map[string]interface{}{
-		"tail":  100,
-		"start": "start",
-		"end":   "end",
-	}
-
-	c := testutils.NewTestContext(t, []string{"tsk_name"}, flags)
-	if err := command.logs(c); err != nil {
+	input := "l0 task logs --tail 100 --start start --end end tsk_name"
```
nehayward: Should all flags be delimitated?
```
input := "l0 task logs "
input += "--tail 100 "
input += "--start start "
input += "--end end tsk_name"
```

cli/command/service_test.go
zpatrick: I was thinking about it, but I figured why not just take care of that whenever we do https://github.com/quintilesims/layer0/issues/505

cli/command/task_test.go
zpatrick: for like anytime there are flags? I could get behind that for consistency's sake

cli/command/service_test.go
diemonster: It works for this one but if the commands are really long I bet wrapping the line will be pretty weird looking.... 

I kind of like the other concatenated `input`s, it reminds me of something you'd do after programming in Go for 3+ years 😉

nehayward: I agree. I meant more just pulling that line out and doing an input var for all commands. I wasn’t saying do away with the concatenation.

cli/command/task_test.go
nehayward: Yup.

PR #542, opened 2018-02-05
530 validate

### LINKED ISSUE #530: Validate() on Request models should error when required fields have zero-values
Currently, if a required field in a request model has a default value, we allow that field to be a 0-value and use the default value where necessary. This isn't a great solution because it requires one to know exactly which fields are and aren't checked during the `RequestModel.Validate()` function. 
It would be better to require that all required fields in a request model be, well, required.

nehayward:
This will add validation on all required fields.

Tested with unit tests.

**Checklist**
- [x] Unit tests

closes #530


common/models/create_environment_request.go
```diff
@@ -41,6 +41,26 @@ func (r CreateEnvironmentRequest) Validate() error {
 		return fmt.Errorf("setting `Scale` is not valid for dynamic environments")
 	}
 
+	if r.EnvironmentType == "" {
+		return fmt.Errorf("EnvironmentType is required, please specify 'static' or 'dynamic'")
+	}
+
+	if r.InstanceType == "" {
```
nehayward: I think we should add a default value for instance type to be dynamic in environment_create.go

sesh-kebab: InstanceType and UserDataTemplate can be empty if the EnvironmentType is `dynamic`.

nehayward: Okay I was just looking into this. I'll remove that check.

sesh-kebab: Or you can make the check conditional.

nehayward: Something like this?
```
if strings.EqualFold(r.EnvironmentType, EnvironmentTypeStatic) && r.UserDataTemplate == ""
		... error
}
```

PR #543, opened 2018-02-06
Update request model validation

### LINKED ISSUE #530: Validate() on Request models should error when required fields have zero-values

zpatrick:
Improves request validation in the `models` pkg, moves `Validate()` calls into the `client` pkg, which allows them to be removed from `cli/command` and `plugins/terraform`. 
Doesn't allow empty values on required params. 

Unit Tests

**Checklist**
- [ ] Unit tests

closes #530 


client/environment_test.go
```diff
@@ -11,11 +11,11 @@ import (
 func TestCreateEnvironment(t *testing.T) {
 	req := models.CreateEnvironmentRequest{
 		EnvironmentName:  "env_name",
+		EnvironmentType:  models.EnvironmentTypeStatic,
```
nehayward: Should we test dynamic as well here?

plugins/terraform/layer0/resource_layer0_service_test.go
```diff
@@ -16,6 +17,8 @@ func TestResourceServiceCreateRead(t *testing.T) {
 
 	mockClient := mock_client.NewMockClient(ctrl)
 
+	defer client.SetTimeMultiplier(0)()
```
nehayward: Is this still needed?

client/environment_test.go
zpatrick: I'm not convinced that would add much value; I'd be more open to something like adding tests for the models' `Validate()` functions.

plugins/terraform/layer0/resource_layer0_service_test.go
zpatrick: Not sure what you mean because I just added it in

client/environment_test.go
nehayward: Okay I'm just wondering if it should be tested somewhere. So that would work.

plugins/terraform/layer0/resource_layer0_service_test.go
nehayward: I'm wondering what does this does when we only deal with mockclient in the test?

zpatrick: check out the code we are testing and you'll see where this comes into play

nehayward: Okay I checked it out. The variable name through me off a little.

diemonster:
@zpatrick can you talk a bit about the pros and cons of moving the `Validate()` functions out of `models`? I'm slightly confused on how request validation will work in Terraform going forward --- I'm assuming the plan is to rely on terraform errors instead?

zpatrick:
@diemonster the `Validate()` functions themselves aren't moved, it's where those functions are being called that's moved. This keeps the code DRY.

common/models/create_environment_request.go
```diff
@@ -22,23 +21,46 @@ func (r CreateEnvironmentRequest) Validate() error {
 		return fmt.Errorf("EnvironmentName is required")
 	}
 
-	if r.Scale < 0 {
-		return fmt.Errorf("Scale must be a positive integer")
+	switch r.EnvironmentType {
+	case EnvironmentTypeStatic, EnvironmentTypeDynamic:
```
nehayward: Should this return anything or have an info line?

common/models/validation_test.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	updateServiceRequest := func(fn func(*UpdateServiceRequest)) *UpdateServiceRequest {
+		req := &UpdateServiceRequest{}
+		fn(req)
+		return req
+	}
+
+	// todo: dynamic environment checks? may not be required depending on changes
+	cases := map[string]Validator{
```
nehayward: Nice! This looks rigorous.

common/models/create_environment_request.go
zpatrick: I don't think there is an need to do anything on success cases here

PR #544, opened 2018-02-06
541: Add AWS region output to -d flag for endpoint and allow spaces for inputs

### LINKED ISSUE #541: `l0-setup endpoint -d -i` does not export the `LAYER0_AWS_REGION` env var
### Expected behavior
Setting the environment variables from `l0-setup endpoint -d -i` should give a user(/developer) everything they need to run Layer0.

### Actual behavior
Attempting to run the API locally for a cluster not created in the default `us-west-2` fails, because `l0-setup` doesn't export the `LAYER0_AWS_REGION` variable, and it must be set manually for clusters in other regions (for example, when I'm trying to test Fargate things in `us-east-1`). It's an easy enough workaround to manually export the var, but still not desired behavior.

```
cd $layer0dir/api && go run -v main.go -debug

2018/02/05 13:08:12 [DEBUG] Adding API tags
2018/02/05 13:08:13 ClusterNotFoundException: Cluster not found.
        status code: 400, request id: 4496be12-0ab8-11e8-8961-4d5bc4863735
exit status 1
```

Try running the API locally using the vars from `l0-setup endpoint -d -i` with the current state of `api-refactor` (or `500-launch-type-for-services-and-tasks` in my case, which is based off of `api-refactor`).

nehayward:
Adds region to the outputs for -d flag

**Checklist**
- [x] Unit tests

closes #541

### DIFF
--- a/setup/instance/inputs.go
+++ b/setup/instance/inputs.go
@@ -194,7 +194,7 @@ func RequiredStringPrompter(m ModuleInput, current interface{}) (interface{}, er
 			scanner := bufio.NewScanner(os.Stdin)
 			scanner.Scan()
 			if err := scanner.Err(); err != nil {
-				return nil, fmt.Errorf("Error for '%s': %e", m.Name, err)
+				return nil, fmt.Errorf("Error for '%s': %s", m.Name, err.Error())
 			}
 
 			input := scanner.Text()
--- a/setup/instance/inputs.go
+++ b/setup/instance/inputs.go
@@ -1,8 +1,10 @@
 package instance
 
 import (
+	"bufio"
 	"fmt"
 	"log"
+	"os"
 
 	"github.com/quintilesims/layer0/common/config"
 )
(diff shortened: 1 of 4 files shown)


setup/instance/inputs.go
```diff
@@ -188,8 +190,10 @@ func RequiredStringPrompter(m ModuleInput, current interface{}) (interface{}, er
 		for i := 0; i < 3; i++ {
 			fmt.Printf("\tInput: ")
 
-			var input string
-			fmt.Scanln(&input)
+			// Allow spaces for input
+			reader := bufio.NewReader(os.Stdin)
```
zpatrick: where/why do we want spaces for inputs? This impl is swallowing errors.

nehayward: I want the ability to add file paths with spaces.

zpatrick: From golang's docs for [ReadLine](https://golang.org/pkg/bufio/#Reader.ReadLine), they suggest using a [Scanner](https://golang.org/pkg/bufio/#Scanner):

```go
scanner := bufio.NewScanner(os.Stdin)
scanner.Scan()
if err := scanner.Err(); err != nil {
    return err
}

input := scanner.Text()
```

jparsons04:
Also looks good to me, as I implemented your changes in my other PR to handle the tfstate push issue: https://github.com/quintilesims/layer0/pull/552

PR #545, opened 2018-02-07
Release v0.10.6

sesh-kebab:
Fixes issue where setting certificate on a load balancer can result in the load balancer getting recreated on every consecutive terraform apply.


api/backend/ecs/load_balancer_manager.go
```diff
@@ -220,6 +220,7 @@ func (e *ECSLoadBalancerManager) listenerToPort(listener *awselb.Listener) model
 	}
 
 	if listener.SSLCertificateId != nil {
+		port.CertificateARN = *listener.SSLCertificateId
 		port.CertificateName = id.CertificateARNToName(*listener.SSLCertificateId)
```
zpatrick: won't the diff fail on `port.CertificateName` unless suppressed in terraform?

sesh-kebab: `port.CertificateName` isn't exposed in Terraform. Only `Certificate` which uses `CertificateARN` is it isn't empty.

https://github.com/quintilesims/layer0/blob/develop/plugins/terraform/resource_layer0_load_balancer.go#L293

PR #547, opened 2018-02-09
Clean up makefiles and simplify CI

### LINKED ISSUE #74: Separate travis CI commands and local commands in Makefiles
Running certain `make` command in Travis CI require workarounds and nuances one wouldn't know about just looking at the Makefiles. It would be nice to separate out any Makefile target that needs special behavior in travis into it's own target, with a `travis-` prefix. For example, running smoketests locally call path is:
```
make smoketests 
	$(MAKE) -C tests/smoke test
```

Running this in travis requires extra calls:
```
make install-smoketests 
        $(MAKE) -C cli install-smoketest
	$(MAKE) -C setup install-smoketest
	$(MAKE) -C api deps
	$(MAKE) -C api release
	$(MAKE) -C runner release
make apply-smoketests
	$(MAKE) -C setup apply-smoketest
make destroy-smoketests
	$(MAKE) -C setup destroy-smoketest
```

We should make it clear which targets are meant for travis and have a consistent naming scheme:
```
make travis-deps
    $(MAKE) -C api travis-deps          # build&pushes api docker image
    $(MAKE) -C runner travis-deps       # build&pushes runner docker image
    $(MAKE) -C setup travis-deps        # installs l0-setup, runs l0-setup apply  
    $(MAKE) -C cli travis-deps          # installs l0 binary
    $(MAKE) -C test/smoke travis-deps   # installs bats

make smoketests 
	$(MAKE) -C tests/smoke test     # note this hasn't changed - all deps have been taken care of

make travis-teardown
    $(MAKE) -C setup travis-teardown    # runs l0-setup destroy 

```

nehayward:
This simplifies the CI and makefiles. It removes the old smoke test for the CI, this should be done manually now.

**Checklist**
- [x] Unit tests

closes #74


nehayward:
So I believe we still have some stuff in develop that's not in api-refactor that needs to be added in on the merge. It's passing right now so the coverage that we want is there now.

PR #552, opened 2018-02-14
Fix l0-setup apply issue for instances created in another region

### LINKED ISSUE #444: failure to push remote tf state when not using default region

jparsons04:
Issue #444 fixed an issue where terraform state would not be pushed at setup time if the layer0 instance was created in a different region. This change was not applied to the api-refactor branches, so this changes that.

Run `l0-setup init` and specify an instance in a region other than us-west-2, then `l0-setup apply`

**Checklist**
~- [ ] Unit tests~

closes #444

### DIFF
--- a/setup/command/pull.go
+++ b/setup/command/pull.go
@@ -2,7 +2,12 @@ package command
 
 import (
 	"fmt"
+	"strings"
 
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3iface"
+	common_aws "github.com/quintilesims/layer0/common/aws"
 	"github.com/urfave/cli"
 )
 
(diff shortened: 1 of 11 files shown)


setup/command/pull.go
```diff
@@ -33,3 +62,32 @@ func (f *CommandFactory) Pull() cli.Command {
...
+
+	for _, bucket := range listBucketsOutput.Buckets {
+		bucketName := aws.StringValue(bucket.Name)
+
+		if split := strings.Split(bucketName, "-"); len(split) == 3 && split[0] == "layer0" && split[1] == instanceName {
+			return bucketName, nil
+		}
+	}
+
+	return "", fmt.Errorf("No S3 bucket found for given instance name")
```
zpatrick: [nit] should probably show the instance name with some quotes here: `...No S3 bucket found for instance '%s'", instanceName)`

PR #555, opened 2018-02-21
Clean up smoke test and add build and install to make file

### LINKED ISSUE #74: Separate travis CI commands and local commands in Makefiles

nehayward:
This updates the ci-testing some changes still need to be brought in from the develop branch. This can be done now or when merged.

Tested through ci deployment.

closes #74


Makefile
```diff
@@ -1,6 +1,16 @@
...
 
+build: 
+	$(MAKE) -C api build        
+	$(MAKE) -C setup build       
+	$(MAKE) -C cli build      
+
+install:
+	$(MAKE) -C cli install
+	$(MAKE) -C setup install
+	$(MAKE) -C api release
```
zpatrick: Is the `install` target ever used? Since we aren't really doing CI at the moment, I can't think of a use for it. Also, I'm not sure it makes sense to have `api release` under this target?

zpatrick: well, we already have a `build` target - I don't think there is anything to install w the api.

zpatrick:
(fixed the merge conflict because it was only one line; figured it would be faster to just fix it while i was already looking at it)

Makefile
nehayward: Maybe make a target to run it on docker then. I just want a command to setup everything for a dev.

nehayward: Either way I can remove this. Should I merge the develop stuff here so we can close out this issues.

nehayward:
TY, I'll go ahead and merge it.

Issue #128, opened 2017-03-17
Request: Create packages for distribution


Cryowatt:
Why doesn't layer0 distribute itself as a package instead of a zip?

Windows - Chocolatey
Mac - Homebrew
Linux - RPM and/or DEB

diemonster:
Good idea, no reason other than we haven't done it yet. Would love to get your help on the Chocolatey distribution. 

http://i3.cpcache.com/product/696364158/lets_get_chocolatey_bumper_stickers.jpg

Cryowatt:
Ideally if you move to a packaged-based distribution method, you should add terraform as a package dependency instead of redistributing their binaries as part of the L0 release.

diemonster:
@Cryowatt pinging you for template that you've worked on already for this?

nehayward:
https://askubuntu.com/questions/16446/how-to-get-my-software-into-ubuntu

We are going to have to use https://launchpad.net to distribute on linux. I think this means a separate code base to maintain

PR #556, opened 2018-02-23
Adds homebrew formula for layer0

nehayward:
This adds the homebrew formula for installing layer0.

Once this is merged in master we can do.
`brew install quintilesims/layer0/layer0`

`(user/repo/formula)`


diemonster:
This partially addresses https://github.com/quintilesims/layer0/issues/128, sans Windows support (Chocolatey). Is there a better option than Chocolatey at this point? cc @Cryowatt

packages/homebrew/layer0.rb
```diff
@@ -0,0 +1,22 @@
+class Layer0 < Formula
+  desc "Framework that helps you deploy to the AWS with minimal fuss"
```
diemonster: [nit] typo: 'to the AWS' (unless you're going for an old man vibe)

packages/homebrew/layer0.rb
```diff
@@ -0,0 +1,22 @@
...
+  homepage "http://layer0.ims.io"
+  url "https://github.com/quintilesims/layer0/archive/v0.10.6.tar.gz"
```
diemonster: this is targeting v0.10.X releases, but the PR is targeting `api-refactor` --- we should either preemptively switch to a v0.11 release that doesn't exist yet, or target this PR for `develop`.

nehayward: Yup, I'll go ahead and change that.

nehayward:
Yeah I'm curious what the best package managers are for linux and windows.

packages/homebrew/layer0.rb
nehayward: Actually I think I'll leave it until 11 is read so we can at least test this.

packages/homebrew/layer0.rb
```diff
@@ -0,0 +1,22 @@
+class Layer0 < Formula
+  desc "Framework that helps you deploy to AWS with minimal fuss"
+  homepage "http://layer0.ims.io"
+  url "https://github.com/quintilesims/layer0/archive/v0.10.6.tar.gz"
```
diemonster: we should use latest version of Layer0

diemonster: btw, is there any way to have these scripts always grab `latest`, or do we always have to manually update them?

nehayward: It doesn't look like it. We would probably have to do a script
https://stackoverflow.com/questions/24987542/is-there-a-link-to-github-for-downloading-a-file-in-the-latest-release-of-a-repo

PR #557, opened 2018-02-23
553: Support ELB Idle Timeout in CLI and Terraform when creating or updating load balancer

### LINKED ISSUE #553: need to configure ELB Idle timeout
Currently there is no way to change the default idle timeout for a l0 loadbalancer. Unfortunately we need to change the default value, but there is no option other than manual updates to the ELB right now.

current l0 version: v0.10.6

jparsons04:
Adds CLI functionality for `l0 loadbalancer create` and a new subcommand `l0 loadbalancer idletimeout`. Create has a new optional flag `--idle-timeout` for create which can be used to specify the Idle Timeout of the load balancer at creation time. `l0 loadbalancer idletimeout` is an update subcommand that updates a given load balancer's idle timeout to a new value.

Adds Terraform support for defining a load balancer's Idle Timeout. A new key is introduced into the resource, `idle_timeout`.

Manually via the CLI: `l0 loadbalancer create env_a lb_a --idle-timeout 100` and `l0 loadbalancer idletimeout lb_a 200`

Via terraform:
```
provider "layer0" {
  skip_ssl_verify = true
}

  name = "env_a"
}

resource "layer0_load_balancer" "lb_a" {
  name        = "lb_a"

  port {
    host_port      = 80
    container_port = 80
    protocol       = "http"
  }

  idle_timeout = 75
}
```
Apply, then change idle_timeout and apply again.

Unit tests have been created and they should pass.  Documentation has been updated to reflect these changes.

**Checklist**
- [x] Unit tests

closes #553


docs-src/docs/reference/cli.md
```diff
@@ -405,6 +406,7 @@ l0 loadbalancer create [--port port ... | --certificate nameOrARN |
 * `--healthcheck-timeout timeout` - The length of time before the check times out (default: `5`).
 * `--healthcheck-healthy-threshold healthyThreshold` - The number of checks before the instance is declared healthy (default: `2`).
 * `--healthcheck-unhealthy-threshold unhealthyThreshold` - The number of checks before the instance is declared unhealthy (default: `2`).
+* `--idle-timeout idleTimeout` - The ELB Idle Timeout in seconds.
```
diemonster: I'm a little confused... is `--idle-timeout` a required or optional param?

docs-src/docs/reference/cli.md
```diff
@@ -512,6 +514,18 @@ l0 loadbalancer healthcheck [--set-target target | --set-interval interval |
 #### Additional information
 Calling the subcommand without flags will display the current configuration of the load balancer's health check. Setting any of the flags will update the corresponding field in the health check, and all omitted flags will leave the corresponding fields unchanged.
 
+### loadbalancer idletimeout
+Use the `idletimeout` subcommand to update the ELB Idle Timout of a load balancer.
```
diemonster: I assume it can be used to view the current idle timeout as well? Also [nit] `Timout`

cli/command/load_balancer_command.go
```diff
@@ -77,6 +77,11 @@ func (l *LoadBalancerCommand) GetCommand() cli.Command {
 						Value: 2,
 						Usage: "number of consecutive failures required to count as unhealthy",
 					},
+					cli.IntFlag{
+						Name:  "idle-timeout",
+						Value: 60,
+						Usage: "idle timeout of the load balancer in seconds",
```
diemonster: I think we should give users some more help on what this value is and how rare it would be to set it to anything other than default? 

I've personally never set this value to anything other than default...

docs-src/docs/reference/cli.md
diemonster: We generally avoid referring to 'ELB' in our docs

docs-src/docs/reference/cli.md
diemonster: ditto re: avoiding reference to `ELB`

docs-src/docs/reference/cli.md
```diff
@@ -512,6 +514,18 @@ l0 loadbalancer healthcheck [--set-target target | --set-interval interval |
...
+
+#### Usage
+```
+l0 loadbalancer idletimeout loadbalancerName idleTimeout
+```
+
+#### Required parameters
+* `loadBalancerName` - The name of the existing Layer0 load balancer you are modifying.
+* `idleTimeout` - The ELB Idle Timeout in seconds.
```
diemonster: ditto re: avoiding reference to `ELB`

docs-src/docs/reference/terraform-plugin.md
```diff
@@ -365,6 +367,7 @@ The following arguments are supported:
 * `private` - (Optional) If true, the load balancer will not be exposed to the public internet
 * `port` - (Optional, Default: 80:80/tcp) A list of port blocks. Ports documented below
 * `health_check` - (Optional, Default: `{"TCP:80" 30 5 2 2}`) A health_check block. Health check documented below
+* `idle_timeout` - (Optiona, Default: 60) The ELB Idle Timeout of the load balancer in seconds
```
diemonster: ditto re: avoiding reference to `ELB`

docs-src/docs/reference/cli.md
jparsons04: Optional, see here: https://github.com/quintilesims/layer0/pull/557/files/10b08bfdc37b962ea6383fbc772573a9b0fff523#diff-743886dffc7c416b92da823ddca0a301R395

docs-src/docs/reference/cli.md
jparsons04: `l0 loadbalancer get` has now been modified to show the idle timeout of the load balancer. The subcommand `idletimeout` is just to update the loadbalancer's idle timeout.

jparsons04:
Re: the comment about Swagger updates, it was updated to the extent that I could here: https://github.com/quintilesims/layer0/pull/557/files/10b08bfdc37b962ea6383fbc772573a9b0fff523#diff-3d7954424b8a45d741cc730ba1391e5eR140

docs-src/docs/reference/cli.md
jparsons04: Although it should be noted that if the load balancer's idle timeout is updated, then the printer will print the load balancer with the new idle timeout value.

docs-src/docs/reference/cli.md
diemonster: ok, you might wanna move this out of 'required parameters' then

docs-src/docs/reference/cli.md
diemonster: ok, i think that's different behavior from the `healthcheck` subcommand, for example. I'm not sure if i agree that `idletimeout` deserves to be part of `loadbalancer get`.

docs-src/docs/reference/cli.md
jparsons04: I think the truncated view of the code makes it look like it's under Required Parameters, but under the fold it should show that it is under the Optional Arguments section.

diemonster: ahhh opps, apologies then!

docs-src/docs/reference/cli.md
jparsons04: I think what I'm going to do is make it behave like `healthcheck` but I'd prefer leaving the `loadbalancer get` column that I added in.

cli/command/load_balancer_command.go
jparsons04: Should that explanation be here or in the documentation? It seems like I should put a more verbose description in the documentation rather than here in the CLI context.

jparsons04: I've opted to add a couple of sections to the CLI and Terraform Plugin reference pages with a more thorough explanation of what the idle timeout is.

api/backend/ecs/load_balancer_manager.go
```diff
@@ -71,7 +71,14 @@ func (e *ECSLoadBalancerManager) GetLoadBalancer(loadBalancerID string) (*models
 		return nil, err
 	}
 
-	return e.populateModel(loadBalancer), nil
+	lbAttributes, err := e.ELB.DescribeLoadBalancerAttributes((*loadBalancer.LoadBalancerName))
```
zpatrick: need to safely de-ref using `aws.StringValue(loadBalancer.LoadBalancerName)`

zpatrick: need to safely de-ref using `aws.StringValue(loadBalancer.LoadBalancerName)`

api/backend/ecs/load_balancer_manager.go
```diff
@@ -71,7 +71,14 @@ func (e *ECSLoadBalancerManager) GetLoadBalancer(loadBalancerID string) (*models
...
+	if err != nil {
+		if ContainsErrCode(err, "LoadBalancerAttributeNotFound") {
+			return nil, errors.New(errors.LoadBalancerAttributeNotFound, err)
```
zpatrick: we should only catch/handle errors if we can do something meaningful with it; this looks like we are just passing the error though here.

api/backend/ecs/load_balancer_manager.go
```diff
@@ -206,6 +213,7 @@ func (e *ECSLoadBalancerManager) populateModel(description *elb.LoadBalancerDesc
 		IsPublic:       stringOrEmpty(description.Scheme) == "internet-facing",
 		URL:            stringOrEmpty(description.DNSName),
 		HealthCheck:    healthCheck,
+		IdleTimeout:    int(*lbAttributes.ConnectionSettings.IdleTimeout),
```
zpatrick: needs safe-dereference

api/backend/ecs/load_balancer_manager.go
```diff
@@ -349,6 +364,15 @@ func (e *ECSLoadBalancerManager) UpdateLoadBalancerHealthCheck(loadBalancerID st
 	return e.GetLoadBalancer(loadBalancerID)
 }
 
+func (e *ECSLoadBalancerManager) UpdateLoadBalancerIdleTimeout(loadBalancerID string, idleTimeout int) (*models.LoadBalancer, error) {
```
zpatrick: Update should only return errors; adding `return e.GetLoadBalancer(loadBalancerID)` is coupling this function to `GetLoadBalancer`, which will make this much harder to test

api/backend/ecs/load_balancer_manager_test.go
```diff
@@ -221,11 +233,14 @@ func TestCreateLoadBalancer(t *testing.T) {
 				mockLB.ELB.EXPECT().
 					ConfigureHealthCheck(loadBalancerID.String(), elbHealthCheck)
 
+				mockLB.ELB.EXPECT().
+					SetIdleTimeout(loadBalancerID.String(), 60)
```
zpatrick: should specify what you are returning here

api/backend/ecs/load_balancer_manager_test.go
```diff
@@ -671,4 +702,47 @@ func TestUpdateLoadBalancerHealthCheck(t *testing.T) {
...
+					Return(loadBalancer, nil)
+
+				loadBalancerAttributes := elb.NewLoadBalancerAttributes()
+
+				mockLB.ELB.EXPECT().
+					DescribeLoadBalancerAttributes(gomock.Any()).
+					Return(loadBalancerAttributes, nil)
+
+				mockLB.ELB.EXPECT().
+					SetIdleTimeout(loadBalancerID.String(), 60)
```
zpatrick: ditto

cli/command/load_balancer_command.go
```diff
@@ -376,6 +388,37 @@ func (l *LoadBalancerCommand) HealthCheck(c *cli.Context) error {
...
+	if err != nil {
+		return err
+	}
+
+	loadBalancer, err := l.Client.GetLoadBalancer(id)
+	if err != nil {
+		return err
+	}
+
+	if timeout := args["TIMEOUT"]; timeout != "" {
```
zpatrick: this will never happen because `extractArgs` would error if it was not specified

cli/printer/test.go
```diff
@@ -20,6 +20,7 @@ func (t *TestPrinter) PrintJobs(...*models.Job) error
 func (t *TestPrinter) PrintLoadBalancers(...*models.LoadBalancer) error                { return nil }
 func (t *TestPrinter) PrintLoadBalancerSummaries(...*models.LoadBalancerSummary) error { return nil }
 func (t *TestPrinter) PrintLoadBalancerHealthCheck(*models.LoadBalancer) error         { return nil }
+func (t *TestPrinter) PrintLoadBalancerIdleTimeout(*models.LoadBalancer) error         { return nil }
```
zpatrick: where would this be used? i don't think we need this.

cli/printer/text.go
```diff
@@ -237,6 +238,28 @@ func (t *TextPrinter) PrintLoadBalancerHealthCheck(loadBalancer *models.LoadBala
 	return nil
 }
 
+func (t *TextPrinter) PrintLoadBalancerIdleTimeout(loadBalancer *models.LoadBalancer) error {
```
zpatrick: ditto - i don't see this being used anywhere.

api/backend/ecs/load_balancer_manager.go
jparsons04: Why does `UpdateLoadBalancerHealthCheck` use this pattern in that case? I modeled this behavior off of the `healthcheck` subcommand.

cli/printer/test.go
jparsons04: It is used in the `l0 loadbalancer idletimeout` subcommand. See here: https://github.com/quintilesims/layer0/blob/553-jparsons-elb-timeout/cli/command/load_balancer_command.go#L419

cli/printer/text.go
jparsons04: See above comment.

api/backend/ecs/load_balancer_manager.go
zpatrick: ah I see - was thinking about `api-refactor` patterns instead of the ones in `develop` - disregard!

cli/printer/test.go
zpatrick: looks like this UX is based off of the healthcheck pattern, which isn't very good IMO. I think it would be better to base this off of something like `l0 service scale`

PR #566, opened 2018-02-28
559: Add load balancer idle timeout support for api-refactor

### LINKED ISSUE #559: Configure ELB idle timeout in Layer0 v0.11.x
Related to https://github.com/quintilesims/layer0/issues/553 , we should add configuration of loadbalancer idle timeout in the v0.11.x code. Questions remain about how to best approach UX for this feature as similar commands have changed in the `api-refactor` branch.

jparsons04:
This adds idle timeout support for load balancers to the api-refactor branches.

* Adds an optional flag for `l0 loadbalancer create`, `--idle-timeout` to specify the idle timeout of the load balancer at creation time
* Adds loadbalancer subcommand `l0 loadbalancer idletimeout` which allows the user to specify the idle timeout of a given load balancer
* Adds terraform support for idle timeout by adding a new key to the load balancer resource `idle_timeout`

Unit tests have been added/modified.

Note: Other unit tests have been modified that are currently breaking in api-refactor.

Manually via the CLI: `l0 loadbalancer create env_a lb_a --idle-timeout 100` and then `l0 loadbalancer idletimeout lb_a 200`

Via terraform:

```
provider "layer0" {
  skip_ssl_verify = true
}

  name = "env_a"
  scale = 1
  environment_type = "static"
  instance_type = "t2.small"
}

resource "layer0_load_balancer" "lb_a" {
  name        = "lb_a"

  port {
    host_port      = 80
    container_port = 80
    protocol       = "http"
  }

  idle_timeout = 75
}
```

Apply, then change `idle_timeout` and apply again.

**Checklist**
- [x] Unit tests

closes #559

### DIFF
--- a/api/provider/aws/load_balancer_update.go
+++ b/api/provider/aws/load_balancer_update.go
@@ -82,6 +82,13 @@ func (l *LoadBalancerProvider) Update(loadBalancerID string, req models.UpdateLo
 		}
 	}
 
+	if req.IdleTimeout != 0 {
+		idleTimeout := req.IdleTimeout
+		if err := l.setIdleTimeout(loadBalancerName, idleTimeout); err != nil {
+			return err
+		}
+	}
+
 	return nil
 }
 
@@ -95,6 +102,28 @@ func (l *LoadBalancerProvider) updateHealthCheck(loadBalancerName string, health
 	}
 
 	if _, err := l.AWS.ELB.ConfigureHealthCheck(input); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (l *LoadBalancerProvider) setIdleTimeout(loadBalancerName string, idleTimeout int) error {
+	connectionSettings := &elb.ConnectionSettings{}
+	connectionSettings.SetIdleTimeout(int64(idleTimeout))
+
+	loadBalancerAttributes := &elb.LoadBalancerAttributes{}
+	loadBalancerAttributes.SetConnectionSettings(connectionSettings)
+
+	input := &elb.ModifyLoadBalancerAttributesInput{}
+	input.SetLoadBalancerName(loadBalancerName)
+	input.SetLoadBalancerAttributes(loadBalancerAttributes)
+
+	if err := input.Validate(); err != nil {
+		return err
+	}
+
+	if _, err := l.AWS.ELB.ModifyLoadBalancerAttributes(input); err != nil {
 		return err
 	}
 
--- a/api/provider/aws/load_balancer_update.go
+++ b/api/provider/aws/load_balancer_update.go
@@ -82,9 +82,9 @@ func (l *LoadBalancerProvider) Update(loadBalancerID string, req models.UpdateLo
 		}
 	}
 
-	if req.IdleTimeout != 0 {
+	if req.IdleTimeout != nil {
 		idleTimeout := req.IdleTimeout
-		if err := l.setIdleTimeout(loadBalancerName, idleTimeout); err != nil {
+		if err := l.setIdleTimeout(loadBalancerName, *idleTimeout); err != nil {
 			return err
 		}
 	}
--- a/api/provider/aws/test_aws/environment_create_test.go
+++ b/api/provider/aws/test_aws/environment_create_test.go
@@ -12,7 +12,6 @@ import (
 	provider "github.com/quintilesims/layer0/api/provider/aws"
 	"github.com/quintilesims/layer0/api/tag"
 	awsc "github.com/quintilesims/layer0/common/aws"
-	"github.com/quintilesims/layer0/common/config"
 	"github.com/quintilesims/layer0/common/config/mock_config"
 	"github.com/quintilesims/layer0/common/models"
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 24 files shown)


api/provider/aws/load_balancer_update.go
```diff
@@ -82,6 +82,13 @@ func (l *LoadBalancerProvider) Update(loadBalancerID string, req models.UpdateLo
 		}
 	}
 
+	if req.IdleTimeout != 0 {
```
zpatrick: update request fields should be pointers, so this will need to become `if req.IdleTimeout != nil`

api/provider/aws/test_aws/environment_create_test.go
```diff
@@ -261,30 +260,6 @@ func TestEnvironmentCreateDefaults(t *testing.T) {
 		AuthorizeSecurityGroupIngress(gomock.Any()).
 		Return(&ec2.AuthorizeSecurityGroupIngressOutput{}, nil)
 
-	// ensure we pass the default instance type, ami id, and user data to the launch configuration
```
zpatrick: why is this being removed?

cli/command/load_balancer.go
```diff
@@ -374,6 +386,38 @@ func (l *LoadBalancerCommand) healthcheck(c *cli.Context) error {
...
 
+func (l *LoadBalancerCommand) idletimeout(c *cli.Context) error {
+	args, err := extractArgs(c.Args(), "LOAD_BALANCER_NAME", "TIMEOUT")
+	if err != nil {
+		return err
+	}
+
+	idleTimeout, err := strconv.Atoi(args["TIMEOUT"])
+	if err != nil {
+		return fmt.Errorf("Failed to parse TIMEOUT argument: %v", args["TIMEOUT"])
```
zpatrick: `...TIMEOUT argument: %v", err)`

cli/command/load_balancer_test.go
```diff
@@ -181,6 +183,40 @@ func TestLoadBalancerDropPortInputErrors(t *testing.T) {
...
+	input := "l0 loadbalancer idletimeout lb_name 80"
+	if err := testutils.RunApp(command, input); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestLoadBalancerIdleTimeoutInputErrors(t *testing.T) {
+	testInputErrors(t, NewLoadBalancerCommand(nil).Command(), map[string]string{
+		"Missing LOAD_BALANCER_NAME arg": "l0 loadbalancer idletimeout",
+		"Missing TIMEOUT arg":            "l0 loadbalancer idletimeout name",
```
zpatrick: needs a case for a non-integer timeout

cli/printer/text.go
```diff
@@ -188,6 +188,28 @@ func (t *TextPrinter) PrintLoadBalancerHealthCheck(loadBalancer *models.LoadBala
 	return nil
 }
 
+func (t *TextPrinter) PrintLoadBalancerIdleTimeout(loadBalancer *models.LoadBalancer) error {
```
zpatrick: this is overkill imo; other update functions just use the default entity printing or just print some sort of a message.

common/errors/code.go
```diff
@@ -7,13 +7,14 @@ func (e ErrorCode) String() string {
...
-	ServiceDoesNotExist      ErrorCode = "ServiceDoesNotExist"
-	TaskDoesNotExist         ErrorCode = "TaskDoesNotExist"
-	UnexpectedError          ErrorCode = "UnexpectedError"
+	InvalidRequest                ErrorCode = "InvalidReqest"
+	DeployDoesNotExist            ErrorCode = "DeployDoesNotExist"
+	EnvironmentDoesNotExist       ErrorCode = "EnvironmentDoesNotExist"
+	DependencyError               ErrorCode = "DependencyError"
+	JobDoesNotExist               ErrorCode = "JobDoesNotExist"
+	LoadBalancerDoesNotExist      ErrorCode = "LoadBalancerDoesNotExist"
+	LoadBalancerAttributeNotFound ErrorCode = "LoadBalancerAttributeNotFound"
```
zpatrick: I don't think we need this error code; there is no actionable item for it, and it's also a leaky aws abstraction.

common/models/update_load_balancer_request.go
```diff
@@ -7,6 +7,7 @@ import (
 type UpdateLoadBalancerRequest struct {
 	Ports       *[]Port      `json:"ports"`
 	HealthCheck *HealthCheck `json:"health_check"`
+	IdleTimeout int          `json:"idle_timeout"`
```
zpatrick: needs to be a `*int`

api/provider/aws/test_aws/environment_create_test.go
jparsons04: Because of a failing unit test related to work being done in another branch off api-refactor. Specifically, the default behavior (in `api-refactor) of creating an environment (i.e. `l0 environment create env_a`) is creating a stateless environment. This means that Launch Configurations and Auto Scaling Groups are not created. I'd rather not have a commit that shows breaking unit tests and would rather fix the unit tests that are breaking given `api-refactor`'s current state.

cli/command/load_balancer_test.go
jparsons04: :white_check_mark: Added

cli/printer/text.go
jparsons04: This is not entirely true because (here)[https://github.com/quintilesims/layer0/blob/559-jparsons-lb-timeout/cli/command/load_balancer.go#L386] we see an example where the `l0 loadbalancer healthcheck` subcommand, one that updates the health check of a load balancer, has its own text printer and calls it when the update is done.

api/provider/aws/test_aws/environment_create_test.go
zpatrick: i see, make note that this will cause issue with https://github.com/quintilesims/layer0/pull/567, since it should exist in that pr

cli/printer/text.go
zpatrick: fair point, but I'd also argue that`l0 loadbalancer healthcheck` is poorly designed. 
Regardless, consider my comments on this a [nit]

PR #567, opened 2018-02-28
[WIP] 500: Set LaunchType for services and tasks; remove EnvironmentType from environments

### LINKED ISSUE #500: Set LaunchType when creating Services and Tasks

tlake:
Provides a solution for #500 that adds a `--stateful` boolean flag to the CLI's `service create` and `task create` to indicate a user's intent to launch on EC2 or Fargate instances. A similar flag is provided for the Terraform plugin. As a side effect, this changeset also removes the `EnvironmentType` (= "static" or "dynamic") field from the `Environment` model.

CLI and Terraform plugin read events have also been updated to understand the concept of a service or task's statefulness.

A user should specify `requiresCompatibilities` in the task definition file at the time of Deploy creation, and AWS will evaluate the task definition's compatibilities when it is registered in the course of Deploy creation. If a user fails to specify any `requiresCompatibilities`, Layer0 will supply both `"EC2"` and `"FARGATE"`. In order for a service or task to be considered FARGATE-compatible (or "stateless"), a few additional fields must be configured in the task definition - see the **Notes** below.

- Create deploys from task definitions that have various combinations of `"EC2"` and `"FARGATE"` in the top-level `requiresCompatibilities` field and verify that the resulting deploy prints out the correct `COMPATIBILITIES` (`"stateful"` for EC2, `"stateless"` for FARGATE). Then, create service and tasks from the new deploys and provide various valid and invalid declarations of the `--stateful` flag. It seems that AWS will always check for EC2 compatibility, but will only check for Fargate compatibility when explicitly told to do so. Since `"networkMode": "awsvpc"` logic has been decoupled from the launch type, I think that most any Fargate-compatible deploys should work with stateful services and tasks.
    - `l0 service create --stateful env-name svc-name ec2-dpl:latest` should succeed ✅
    - `l0 service create --stateful env-name svc-name fargate-dpl:latest` should succeed ✅
    - `l0 service create env-name svc-name ec2-dpl:latest` should fail ❌ 
    - `l0 service create env-name svc-name fargate-dpl:latest` should succeed ✅
- Verify that the Terraform plugin functions as expected
- Verify that unit tests pass

The smoketests have been updated for the new workflow, and updated task definitions have been provided. That's probably a good place to gain understanding of the sort of workflows to use in manual testing.

**Notes:**
- Fargate is only available in `us-east-1` right now, so make sure that's set in `l0-setup`.
- I've pushed an API binary to dockerhub: specify `tlaketest` in `l0-setup`.
- There are some requirements for a task definition file to be Fargate-compatible. Probably the most notable are the following four parameters (defined at the same level as `"containerDefinitions"`):
```
    "requiresCompatibilities": [ "FARGATE" ], 
    "networkMode": "awsvpc", 
    "cpu": "256", 
    "memory": "0.5GB",
```
More info at https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html - in particular, note that there are specific values allowed for `"cpu"` and `"memory"`.
- Effort has been made to be as hands-off as possible in error handling. Because Fargate is so new, it is likely pointless to anchor Layer0 to its current implementation, and so Layer0 will by and large do what the user tells it do and then pass along any AWS errors that arise. These errors seem to be relatively helpful; it would be a good idea to keep an eye out for any error messages which are confusing or opaque, and also to verify that expected AWS errors are arising.

**Checklist**
- [x] Unit tests
- ~[ ] Documentation (if applicable)~ (a separate issue exists for documentation, which will be undertaken once a path for launch types has been greenlit)

closes #500


api/provider/aws/task_create.go
```diff
@@ -5,29 +5,63 @@ import (
...
+	taskOverrides := convertContainerOverrides(req.ContainerOverrides)
 
 	deployName, deployVersion, err := lookupDeployNameAndVersion(t.TagStore, req.DeployID)
 	if err != nil {
 		return "", err
 	}
 
-	clusterName := fqEnvironmentID
-	startedBy := t.Config.Instance()
 	taskDefinitionFamily := addLayer0Prefix(t.Config.Instance(), deployName)
```
tlake: Is there a reason why I shouldn't change Task to use task definition ARNs (like Service does) instead of "name:version"?

api/provider/aws/deploy_create.go
```diff
@@ -47,6 +47,24 @@ func (d *DeployProvider) createTaskDefinition(taskDefinition *ecs.TaskDefinition
 		input.SetNetworkMode(nm)
 	}
 
+	// Because of the logic in renderTaskDefinition(), taskDefinition.RequiresCompatibilities
+	// should always contain one or both of "EC2" and "FARGATE"
+	input.SetRequiresCompatibilities(taskDefinition.RequiresCompatibilities)
+
+	if cpu := aws.StringValue(taskDefinition.Cpu); cpu != "" {
```
zpatrick: this (and memory below) is essentially saying "if thing is set, then set thing"; it's redundant.

api/provider/aws/service_create.go
```diff
@@ -5,22 +5,66 @@ import (
...
 func (s *ServiceProvider) Create(req models.CreateServiceRequest) (string, error) {
 	fqEnvironmentID := addLayer0Prefix(s.Config.Instance(), req.EnvironmentID)
 	cluster := fqEnvironmentID
-
 	serviceID := entityIDGenerator(req.ServiceName)
 	fqServiceID := addLayer0Prefix(s.Config.Instance(), serviceID)
 	serviceName := fqServiceID
 
+	scale := req.Scale
+	if req.Scale == 0 {
```
zpatrick: this should be handled by `models.CreateServiceRequest.Validate()`

nehayward: Should this be a less than or equal? Or should we assert this somewhere else, is it weird to have the default be 1?

api/provider/aws/service_read_test.go
```diff
@@ -85,4 +85,5 @@ func TestService_makeServiceModel(t *testing.T) {
 	assert.Equal(t, "lb_name", result.LoadBalancerName)
 	assert.Equal(t, "svc_id", result.ServiceID)
 	assert.Equal(t, "svc_name", result.ServiceName)
+	assert.Equal(t, "stateless", result.ServiceType)
```
zpatrick: [nit] should we use the `const` here?

api/provider/aws/task_create.go
```diff
@@ -5,29 +5,69 @@ import (
...
+	taskDefinition, err := describeTaskDefinition(t.AWS.ECS, taskDefinitionARN)
+	if err != nil {
+		return "", err
+	}
+
+	networkMode := aws.StringValue(taskDefinition.NetworkMode)
+
+	var securityGroupIDs []*string
+	var subnets []string
+	if networkMode == "awsvpc" {
```
zpatrick: ditto

api/provider/aws/service_create.go
```diff
@@ -76,18 +126,49 @@ func (s *ServiceProvider) Create(req models.CreateServiceRequest) (string, error
...
+
+	launchType := ecs.LaunchTypeEc2
+	if serviceType == models.DeployCompatibilityStateless {
+		launchType = ecs.LaunchTypeFargate
+		input.SetPlatformVersion(fargatePlatformVersion)
+	}
+
+	input.SetLaunchType(launchType)
+
+	if networkMode == "awsvpc" {
```
zpatrick: is there a `const` in `aws-sdk-go` we can user here?

api/provider/aws/task_create.go
```diff
@@ -65,14 +105,48 @@ func convertContainerOverrides(overrides []models.ContainerOverride) *ecs.TaskOv
...
-	input.SetTaskDefinition(taskFamilyRevision)
+	launchType := ecs.LaunchTypeEc2
+	if taskType == models.DeployCompatibilityStateless {
+		launchType = ecs.LaunchTypeFargate
+		input.SetPlatformVersion(fargatePlatformVersion)
+	}
+
+	input.SetLaunchType(launchType)
+
+	if networkMode == "awsvpc" {
```
zpatrick: ditto

api/provider/aws/service_create.go
tlake: I think this is because it doesn't make sense for a user to deploy a service with a scale of 0? I dunno, this was existing behavior that wasn't part of my changeset and that I just shifted up a couple dozen lines, so I didn't think too hard about it xD

api/provider/aws/test_aws/environment_create_test.go
```diff
@@ -251,6 +157,7 @@ func TestEnvironmentCreateDefaults(t *testing.T) {
 
 	req := models.CreateEnvironmentRequest{
 		EnvironmentName: "env_name",
+		Scale:           1,
```
nehayward: We should test with 0 as well.

api/provider/aws/service_read_test.go
zpatrick: I'm not a fan of `ServiceType` and `TaskType`; shouldn't it be `DeployCompatibility` ?

api/provider/aws/task_read.go
```diff
@@ -107,11 +112,12 @@ func (t *TaskProvider) lookupTaskARN(taskID string) (string, error) {
...
 
-func (t *TaskProvider) makeTaskModel(taskID, environmentID, deployID string) (*models.Task, error) {
+func (t *TaskProvider) makeTaskModel(taskID, environmentID, deployID, taskType string) (*models.Task, error) {
 	model := &models.Task{
-		TaskID:        taskID,
-		EnvironmentID: environmentID,
 		DeployID:      deployID,
+		EnvironmentID: environmentID,
+		TaskID:        taskID,
+		TaskType:      taskType,
```
zpatrick: ditto

api/provider/aws/task_read_test.go
```diff
@@ -60,4 +60,5 @@ func TestTask_makeTaskModel(t *testing.T) {
 	assert.Equal(t, "env_name", result.EnvironmentName)
 	assert.Equal(t, "dpl_name", result.DeployName)
 	assert.Equal(t, "dpl_version", result.DeployVersion)
+	assert.Equal(t, "stateless", result.TaskType)
```
zpatrick: [nit] ditto

cli/printer/text_test.go
```diff
@@ -9,15 +9,29 @@ import (
 func ExampleTextPrintDeploys() {
 	printer := &TextPrinter{}
 	deploys := []*models.Deploy{
-		{DeployID: "id1", DeployName: "name1", Version: "1"},
-		{DeployID: "id2", DeployName: "name2", Version: "2"},
+		{
+			Compatibilities: []string{models.DeployCompatibilityStateless},
+			DeployID:        "id1",
```
nehayward: [nit] 
```
dpl_id
dpl_name
```

cli/command/task_test.go
```diff
@@ -57,6 +57,54 @@ func TestCreateTask(t *testing.T) {
...
+		{ContainerName: "c1", EnvironmentOverrides: map[string]string{"k1": "v1"}},
+		{ContainerName: "c2", EnvironmentOverrides: map[string]string{"k2": "v2"}},
+	}
+
+	validateRequest := func(req models.CreateTaskRequest) {
+		assert.Equal(t, "tsk_name", req.TaskName)
+		assert.Equal(t, "env_id", req.EnvironmentID)
+		assert.Equal(t, "dpl_id", req.DeployID)
+
+		assert.Len(t, req.ContainerOverrides, len(overrides))
```
zpatrick: pretty sure you can just use [ElementsMatch](https://godoc.org/github.com/stretchr/testify/assert#ElementsMatch) `assert.ElementsMatch(t, req.ContainerOverrides, overrides)`

cli/printer/text_test.go
```diff
@@ -219,72 +230,77 @@ func ExampleTextPrintServices() {
...
 
 	printer.PrintServices(services...)
 	// Output:
-	// SERVICE ID  SERVICE NAME  ENVIRONMENT  LOADBALANCER  DEPLOYMENTS  SCALE
-	// id1         svc1          ename1       lname1        d1:1         1/1
-	// id2         svc2          eid2         lid2          d2:1         1/1
-	// id3         svc3          eid3                       d3:1*        0/1 (1)
-	// id4         svc4          eid4                       d4:1*        1/2 (1)
-	// id5         svc5          eid5                       d5:1*        2/1
+	// SERVICE ID  SERVICE NAME  ENVIRONMENT  LOADBALANCER  DEPLOYMENTS  SCALE    TYPE
```
zpatrick: [nit] this is a little hard to differentiate at a glance, perhaps the column is `STATEFUL` and the rows are `true/false`?

plugins/terraform/layer0/resource_layer0_service.go
```diff
@@ -44,19 +44,31 @@ func resourceLayer0Service() *schema.Resource {
 				Optional: true,
 				Default:  1,
 			},
+			"stateful": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				ForceNew: true,
+				Computed: true,
```
zpatrick: this is not a computed value

plugins/terraform/layer0/resource_layer0_service.go
```diff
@@ -44,19 +44,31 @@ func resourceLayer0Service() *schema.Resource {
...
+			},
 		},
 	}
 }
 
 func resourceLayer0ServiceCreate(d *schema.ResourceData, meta interface{}) error {
 	apiClient := meta.(client.Client)
 
+	serviceType := models.DeployCompatibilityStateless
+	if d.Get("stateful").(bool) == true {
```
zpatrick: [nit] `if d.Get("stateful").(bool) { ... }`

plugins/terraform/layer0/resource_layer0_service.go
```diff
@@ -92,6 +104,7 @@ func resourceLayer0ServiceRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("environment", service.EnvironmentID)
 	d.Set("load_balancer", service.LoadBalancerID)
 	d.Set("scale", service.DesiredCount)
+	d.Set("stateful", bool(service.ServiceType == models.DeployCompatibilityStateful))
```
zpatrick: [nit] ditto

plugins/terraform/layer0/data_source_layer0_service.go
```diff
@@ -65,6 +69,7 @@ func dataSourceLayer0ServiceRead(d *schema.ResourceData, meta interface{}) error
 	d.Set("environment_id", service.EnvironmentID)
 	d.Set("environment_name", service.EnvironmentName)
 	d.Set("scale", service.DesiredCount)
+	d.Set("stateful", bool(service.ServiceType == models.DeployCompatibilityStateful))
```
zpatrick: [nit] don't need to typecast here

plugins/terraform/layer0/resource_layer0_service_test.go
```diff
@@ -66,6 +69,68 @@ func TestResourceServiceCreateRead(t *testing.T) {
...
+		AnyTimes()
+
+	serviceResource := Provider().(*schema.Provider).ResourcesMap["layer0_service"]
+	d := schema.TestResourceDataRaw(t, serviceResource.Schema, map[string]interface{}{
+		"name":          "svc_name",
+		"environment":   "env_id",
+		"load_balancer": "lb_id",
+		"deploy":        "dpl_id",
+		"scale":         3,
+		"stateful":      true,
```
zpatrick: it seems a little odd that we mix bools and strings for the same field. perhaps instead of `ServiceType/TaskType string` fields in the models, they should just have `Stateful bool` fields?

zpatrick:
@tlake: don't forget to update swagger models!

tlake:
@zpatrick Too late, I already did! xD

Thanks for the reminder, I'll get on those.

api/provider/aws/service_create.go
tlake: Hot dang, there is! Great call. ✅

api/provider/aws/service_create.go
nehayward: Okay I wasn't sure either, just wanted to start a discussion. It's probably fine as is.

plugins/terraform/layer0/resource_layer0_service_test.go
tlake: You're right - takin' this train to Boolville. ✅

api/provider/aws/service_read_test.go
tlake: Replacing with `Stateful` boolean field as per https://github.com/quintilesims/layer0/pull/567#discussion_r171947782 ✅

cli/printer/text_test.go
tlake: ✅

api/provider/aws/service_read_test.go
tlake: Moot point after https://github.com/quintilesims/layer0/pull/567#discussion_r171947782 ✅

api/provider/aws/deploy_create.go
tlake: ✅

api/provider/aws/task_create.go

api/provider/aws/task_create.go

api/provider/aws/task_read.go
tlake: ✅

api/provider/aws/task_read_test.go
tlake: ✅

api/provider/aws/service_create.go
tlake: This logic will be removed in favor of the proper channel of `CreateServiceRequest.Validate()`. ✅

cli/command/task_test.go
tlake: Oh cool! TIL. ✅

api/provider/aws/deploy_create.go
tlake: @zpatrick I wonder if this is actually desired behavior. Do we expect the following two task definitions, once they're registered, to be functionally identical?
```
{
  ContainerDefinitions: [{ . . . }],
  RequiresCompatibilities: [ "EC2"],
}
```
vs
```
{
  ContainerDefinitions: [{ . . . }],
  Cpu: "",
  Memory: "",
  RequiresCompatibilities: [ "EC2" ],
}
```

plugins/terraform/layer0/resource_layer0_service.go
tlake: ✅

plugins/terraform/layer0/resource_layer0_service.go
tlake: Moot point after replacing `serviceType` with boolean `stateful`. ✅

plugins/terraform/layer0/resource_layer0_service.go
tlake: Is now `d.Set("stateful", service.Stateful)`. ✅

plugins/terraform/layer0/data_source_layer0_service.go
tlake: ✅

api/provider/aws/task_create.go
```diff
@@ -5,29 +5,69 @@ import (
...
+
+		securityGroupIDs = append(securityGroupIDs, environmentSecurityGroup.GroupId)
+
+		subnets = t.Config.PrivateSubnets()
+	}
 
-	task, err := t.runTask(clusterName, startedBy, taskDefinitionFamily, taskDefinitionVersion, taskOverrides)
+	task, err := t.runTask(
+		clusterName,
+		req.TaskType,
```
sesh-kebab: Same things as the service. You're passing in 2 arguments TaskType and fargatePlatformVersion, the latter of which is derived from the former. But fargatePlatformVersion isn't used in the caller's scope.

tlake:
@diemonster I'm happy to touch on smoketests in this PR. I haven't devoted any time to them yet because I wanted to not waste energy on them if this turned out to be the wrong path to follow. However, I know we've been wanting to get this into `api-refactor` ASAP, so messing with smoketests would definitely delay that.

Another option might be to merge this once code review has passed so that other devs can incorporate these changes in their own work, and then I can do smoketests in the same PR in which I add documentation for this new launch type support.

api/provider/aws/test_aws/environment_create_test.go
sesh-kebab: Why set the Scale when it isn't relevant to the test?

api/provider/aws/test_aws/service_create_test.go
```diff
@@ -90,8 +116,11 @@ func TestServiceCreate(t *testing.T) {
 		containerDefinition,
 	}
 
+	networkMode := "awsvpc"
```
sesh-kebab: 👍

api/provider/aws/test_aws/service_create_test.go
```diff
@@ -187,26 +233,220 @@ func TestServiceCreate_defaults(t *testing.T) {
...
+		},
+		{
+			EntityID:   "svc_id",
+			EntityType: "service",
+			Key:        "environment_id",
+			Value:      "env_id",
+		},
+	}
+
+	for _, tag := range expectedTags {
```
sesh-kebab: Same as the comment above. It would be good if we start checking the count for the tagstore values to match the expected value of 2 in this case.

api/provider/aws/test_aws/environment_create_test.go
tlake: ✅

api/provider/aws/deploy_create.go
zpatrick: as sorry, didn't see that `input` and `taskDefinition` were different variables

tlake: ✅

api/provider/aws/deploy_create.go
```diff
@@ -85,6 +103,16 @@ func (d *DeployProvider) renderTaskDefinition(body []byte, familyName string) (*
 		}
 	}
 
+	// If a user does not specify any requiresCompatibilities in the task definition, we default
```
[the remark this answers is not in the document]
tlake: @sesh-kebab Yeah! So, when you register a task definition, AWS will run through some logic to evaluate the task definition and check its launch type compatibilities; the task definition that AWS returns after registration will contain a new `Compatibilities` field.

It seems that AWS will always check for EC2 compatibility no matter what. If a user specifies `"FARGATE"` in the task definition's `RequiresCompatibilities` field, AWS will also check for Fargate compatibility.

The idea behind this bit of logic is to keep our users informed in spite of themselves. If a user doesn't put anything in `RequiresCompatibilities`, AWS will never check for Fargate compatibility. This bit of logic just says, _"hey, the user might not know that this field exists, because they didn't supply it. let's just go ahead and gather all the compatibility information we can for the user."_ It's just a nicety.

api/provider/aws/deploy_create_test.go
```diff
@@ -72,6 +74,7 @@ func TestDeploy_renderTaskDefinition(t *testing.T) {
 	taskDef.SetContainerDefinitions(containers)
 	taskDef.SetFamily("test_family")
 	taskDef.SetNetworkMode("bridge")
+	taskDef.SetRequiresCompatibilities([]*string{aws.String("EC2")})
```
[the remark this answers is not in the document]
tlake: ✅

api/provider/aws/deploy_create_test.go
```diff
@@ -86,6 +89,59 @@ func TestDeploy_renderTaskDefinition(t *testing.T) {
...
+	expectedContainer.SetName("test_name")
+	expectedContainer.SetLogConfiguration(expectedLogConfig)
+
+	expectedContainers := []*ecs.ContainerDefinition{}
+	expectedContainers = append(expectedContainers, expectedContainer)
+
+	expected := &ecs.TaskDefinition{}
+	expected.SetContainerDefinitions(expectedContainers)
+	expected.SetFamily("test_family")
+	expected.SetRequiresCompatibilities([]*string{aws.String("EC2"), aws.String("FARGATE")})
```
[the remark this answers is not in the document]
tlake: ✅

api/provider/aws/environment_create_test.go
```diff
@@ -13,7 +13,7 @@ func TestEnvironment_createTags(t *testing.T) {
 	tagStore := tag.NewMemoryStore()
 	environment := NewEnvironmentProvider(nil, tagStore, nil)
 
-	if err := environment.createTags("env_id", "env_name", "static", "env_os"); err != nil {
+	if err := environment.createTags("env_id", "env_name", "env_os"); err != nil {
 		t.Fatal(err)
 	}
```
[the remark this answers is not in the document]
tlake: @sesh-kebab There's a pattern in `provider/aws/service_create_test.go` that I think will merge the current `for _, tag := range expectedTags {}` with your `len()` logic:
```
assert.Equal(t, expectedTags, tagStore.Tags())
```

api/provider/aws/deploy_create.go
```diff
@@ -47,6 +47,24 @@ func (d *DeployProvider) createTaskDefinition(taskDefinition *ecs.TaskDefinition
...
+		input.SetCpu(cpu)
+	}
+
+	if mem := aws.StringValue(taskDefinition.Memory); mem != "" {
+		input.SetMemory(mem)
+	}
+
+	// Hard-coding an ARN isn't usually the best thing, but we know exactly what this ARN will look like
+	// and the only variable in it is the account ID, which Layer0 already knows about.
+	accountID := d.Config.AccountID()
```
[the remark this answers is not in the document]
diemonster: It's created as part of `l0-setup`

api/provider/aws/service_create.go
```diff
@@ -5,22 +5,66 @@ import (
...
+// found in its helper functions as well as errors returned by AWS.
 func (s *ServiceProvider) Create(req models.CreateServiceRequest) (string, error) {
 	fqEnvironmentID := addLayer0Prefix(s.Config.Instance(), req.EnvironmentID)
 	cluster := fqEnvironmentID
-
 	serviceID := entityIDGenerator(req.ServiceName)
 	fqServiceID := addLayer0Prefix(s.Config.Instance(), serviceID)
 	serviceName := fqServiceID
 
+	scale := req.Scale
```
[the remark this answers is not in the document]
tlake: This is solved via https://github.com/quintilesims/layer0/pull/567#discussion_r171942118 ✅

api/provider/aws/service_create.go
```diff
@@ -53,18 +102,19 @@ func (s *ServiceProvider) Create(req models.CreateServiceRequest) (string, error
...
 	if err := s.createService(
 		cluster,
+		req.ServiceType,
 		serviceName,
 		taskDefinitionARN,
-		scale,
 		loadBalancerRole,
-		loadBalancer); err != nil {
+		networkMode,
+		fargatePlatformVersion,
```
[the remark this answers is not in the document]
tlake: Good call. ✅

api/provider/aws/task_create.go

api/provider/aws/test_aws/environment_create_test.go
tlake: ✅

api/provider/aws/test_aws/service_create_test.go
tlake: ✅

api/provider/aws/deploy_create.go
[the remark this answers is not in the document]
diemonster: To clarify my answer a bit more, this execution role is something we were already creating as part of `l0-setup`; it's needed elsewhere by ECS. You can see this role being explicitly called out in https://github.com/quintilesims/layer0/pull/567/commits/a73f23457940ce0ede50c1128d9b4a46252217c7 for instance. Why does it need to be called out here? It's a Fargate launch type requirement.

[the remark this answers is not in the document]
tlake: @sesh-kebab It's set here because it's a requirement to execute Fargate services and tasks. Maybe it ought to be moved into a fargate-conditional? It hasn't hurt anything by being out in the open, though. :p

Now that you bring it up, though, I don't _know_ if we can guarantee that this role exists. The docs say that it's created as part of the console first-run walkthrough, but that's probably not guaranteed to have occurred for every user. I think @diemonster made the discovery (and implementation) about needing this role for Fargate; maybe he has additional insight about whether or not we can guarantee it, and if not, what the best solution to making that guarantee would be.

api/provider/aws/service_read.go
```diff
@@ -24,6 +25,8 @@ func (s *ServiceProvider) Read(serviceID string) (*models.Service, error) {
 		return nil, err
 	}
 
+	stateful := bool(aws.StringValue(ecsService.LaunchType) == ecs.LaunchTypeEc2)
```
zpatrick: [nit] don't need a typecast here

tlake: ✅

cli/printer/text_test.go
tlake: I'm tempted to leave these as-is, instead of changing the entity IDs for every single test case in this file. It seems like a lot of busywork for barely any benefit.

nehayward:
Could we list whether the env is fargate or ec2 in the ‘l0 environment list command’

PR #568, opened 2018-02-28
424: Add support for sending L0 api logs to cloudwatch

### LINKED ISSUE #424: Enable Docker debug output for ECS container instances and the Docker daemon

jparsons04:
Modifies the L0 API module to send ECS agent, Docker, and other Linux system logs to the CloudWatch log group created when the instance is created.

Create a new instance with `l0-setup` and note the new log streams created in the log group of your instance.

**Checklist**
~- [ ] Unit tests~

closes #424

### DIFF
--- a/setup/module/api/user_data.sh
+++ b/setup/module/api/user_data.sh
@@ -1,3 +1,7 @@
+# The majority of this script was largely taken from AWS's documentation
+# on using CloudWatch logs with container instances
+# https://docs.aws.amazon.com/AmazonECS/latest/developerguide/using_cloudwatch_logs.html
+
 Content-Type: multipart/mixed; boundary="==BOUNDARY=="
 MIME-Version: 1.0
 
--- a/setup/module/api/user_data.sh
+++ b/setup/module/api/user_data.sh
@@ -1,9 +1,99 @@
+Content-Type: multipart/mixed; boundary="==BOUNDARY=="
+MIME-Version: 1.0
+
+--==BOUNDARY==
 #!/bin/bash
+exec > >(tee /var/log/user-data.log|logger -t user-data -s 2>/dev/console) 2>&1
 echo ECS_CLUSTER=${cluster_id} >> /etc/ecs/ecs.config
 echo ECS_ENGINE_AUTH_TYPE=dockercfg >> /etc/ecs/ecs.config
-yum install -y aws-cli awslogs
+echo ECS_LOGLEVEL=info >> /etc/ecs/ecs.config
+
+# Known issue where packages will sometimes not install on the first try
+n=0
+until [ $n -ge 5 ]
+do
+    yum install -y aws-cli awslogs jq && break
+    n=$[$n+1]
+    sleep 5
+done
+
+# Inject the CloudWatch Logs configuration file contents
+cat > /etc/awslogs/awslogs.conf <<- EOF
+[general]
+state_file = /var/lib/awslogs/agent-state        
+ 
+[/var/log/docker]
+file = /var/log/docker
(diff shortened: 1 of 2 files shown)


setup/module/api/user_data.sh
```diff
@@ -1,9 +1,99 @@
...
+
+--==BOUNDARY==
 #!/bin/bash
+exec > >(tee /var/log/user-data.log|logger -t user-data -s 2>/dev/console) 2>&1
 echo ECS_CLUSTER=${cluster_id} >> /etc/ecs/ecs.config
 echo ECS_ENGINE_AUTH_TYPE=dockercfg >> /etc/ecs/ecs.config
-yum install -y aws-cli awslogs
+echo ECS_LOGLEVEL=info >> /etc/ecs/ecs.config
+
+# Known issue where packages will sometimes not install on the first try
```
diemonster: After briefly reading the following doc https://docs.aws.amazon.com/AmazonECS/latest/developerguide/docker-debug-mode.html , I'm curious how much of this script needs to be included when we're already using the AWS ECS optimized AMI. 

Also curious where the source of much of this script is located (unless you wrote the majority of it?)

jparsons04: The script largely came from this page: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/using_cloudwatch_logs.html

jparsons04: The problem with just enabling debug mode in docker is that the logs aren't being shipped anywhere for review. The page I linked talks about how you have to leverage the awslogs Log Driver to actually send logs to CloudWatch, so I modified it a bit to suit our needs.

diemonster: Couple thoughts:

- We should reference in a comment where the source of that script came from. 
- I wonder if we're doubling up commands given that we're already using the ECS Optimized AMI? It's hard to tell whether or not the entirety of that script is required.

jparsons04: I'll add a comment for where I got the script from.

As for the contents, in the documentation, it actually makes a few explicit references that the configurations in the script are to be used in ECS-optimized instances:
```
You can configure your container instances to send log information to CloudWatch Logs. This enables you to view different logs from your container instances in one convenient location. This topic helps you get started using CloudWatch Logs on your container instances that were launched with the Amazon ECS-optimized AMI.
```

```
The CloudWatch Logs agent configuration file (/etc/awslogs/awslogs.conf) describes the log files to send to CloudWatch Logs. The agent configuration file's [general] section defines common configurations that apply to all log streams, and you can add individual log stream sections for each file on your container instances that you want to monitor. For more information, see CloudWatch Logs Agent Reference in the Amazon CloudWatch User Guide.

The example configuration file below is configured for the Amazon ECS-optimized AMI, and it provides log streams for several common log files:
```

So it looks to me like you are supposed to include these configurations and things when launching an ECS-optimized AMI.

diemonster:
@jparsons04 Should `debug` mode be a configurable option? How noisy are the additional logs?

jparsons04:
DEBUG level is pretty noisy, so where applicable, I made the log level INFO

PR #570, opened 2018-03-07
569 WIP: Retry logic for load balancer. 

### LINKED ISSUE #569: Add retry logic for eventually consistent resources
When creating and deleting resources in Layer0, the AWS API often quickly returns a response designating that they have received the request, but that request is not necessarily reflective of reality. AWS terms this as 'eventual consistency', and we must build retry logic in order to account for it.

ELBs and Security Groups are the biggest offenders here --- but other AWS resources have the potential to be eventually consistent as well. As such, we should build middleware in the API's `client` package that checks if a resource has been created or deleted before returning.

* [x] LB Delete #584
* [x] Env Delete #583
* [x] Service Create #582
* [x] API client retry when hit specific dependency errors #585

nehayward:
I'm looking for feedback on how the retry logic should be designed.

**Checklist**
- [x] Unit tests

closes #569

### DIFF
--- a/api/provider/aws/load_balancer_delete.go
+++ b/api/provider/aws/load_balancer_delete.go
@@ -1,13 +1,14 @@
 package aws
 
 import (
+	"fmt"
 	"strings"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/elb"
 	"github.com/aws/aws-sdk-go/service/iam"
-	"github.com/quintilesims/layer0/common/config"
 	"github.com/quintilesims/layer0/common/errors"
 )
 
@@ -46,49 +47,70 @@ func (l *LoadBalancerProvider) Delete(loadBalancerID string) error {
 		}
 	}
 
-	// Confirm Delete
-	consistencyCheckFN := func() error {
+	getRoleFN := func() error {
 		// Check IAM Role
-		getRoleInput := &iam.GetRoleInput{}
-		getRoleInput.SetRoleName(roleName)
+		input := &iam.GetRoleInput{}
+		input.SetRoleName(roleName)
 
-		if _, err := l.AWS.IAM.GetRole(getRoleInput); err != nil {
+		if _, err := l.AWS.IAM.GetRole(input); err != nil {
 			if err, ok := err.(awserr.Error); ok && err.Code() == "NoSuchEntity" {
 				return nil
 			}
-
 			return err
 		}
-
-		// Check Role Policy
-		getRolePolicyInput := &iam.GetRolePolicyInput{}
-		getRolePolicyInput.SetRoleName(roleName)
-		getRolePolicyInput.SetPolicyName(policyName)
-
-		if _, err := l.AWS.IAM.GetRolePolicy(getRolePolicyInput); err != nil {
-			if err, ok := err.(awserr.Error); ok && err.Code() == "NoSuchEntity" {
-				return nil
-			}
-
-			return err
-		}
-
-		// Check Security Group
-		_, err := readSG(l.AWS.EC2, securityGroupName)
-		if err != nil && !strings.Contains(err.Error(), "does not exist") {
-			return err
-		}
-
-		_, err = describeLoadBalancer(l.AWS.ELB, fqLoadBalancerID)
-		if serverError, ok := err.(*errors.ServerError); ok && serverError.Code == "LoadBalancerDoesNotExist" {
-			return nil
-		}
-
-		return err
+		return nil
 	}
 
-	if err := retry(config.DefaultRetryTimeout, config.DefaultRetryTick, consistencyCheckFN); err != nil {
-		return err
+	getRolePolicyeFN := func() error {
+		// Check Role Policy
+		input := &iam.GetRolePolicyInput{}
+		input.SetRoleName(roleName)
+		input.SetPolicyName(policyName)
+
+		if _, err := l.AWS.IAM.GetRolePolicy(input); err != nil {
+			if err, ok := err.(awserr.Error); ok && err.Code() == "NoSuchEntity" {
+				return nil
+			}
+			return err
+		}
+		return nil
+	}
+
+	readSecurityGroupFN := func() error {
+		// Check Security Group
+		if _, err := readSG(l.AWS.EC2, securityGroupName); err != nil && !strings.Contains(err.Error(), "does not exist") {
+			return err
+		}
+		return nil
+	}
+
+	desribeLoadBalancerFN := func() error {
+		if _, err = describeLoadBalancer(l.AWS.ELB, fqLoadBalancerID); err != nil {
+			if serverError, ok := err.(*errors.ServerError); ok && serverError.Code == "LoadBalancerDoesNotExist" {
+				return nil
+			}
+			return err
+		}
+		return nil
+	}
+
+	ch := make(chan error, 4)
+	defer close(ch)
+
+	go retry(10*time.Second, 2*time.Second, ch, getRoleFN)
+	go retry(10*time.Second, 2*time.Second, ch, getRolePolicyeFN)
+	go retry(30*time.Second, 5*time.Second, ch, readSecurityGroupFN)
+	go retry(10*time.Second, time.Second, ch, desribeLoadBalancerFN)
+
+	var errstrings []string
+	for i := 0; i < 4; i++ {
+		if err := <-ch; err != nil {
+			errstrings = append(errstrings, err.Error())
+		}
+	}
+
+	if len(errstrings) > 0 {
+		return errors.New(errors.FailedRequestTimeout, fmt.Errorf(strings.Join(errstrings, "\n")))
 	}
 
 	if err := l.deleteTags(loadBalancerID); err != nil {
--- a/api/provider/aws/load_balancer_delete.go
+++ b/api/provider/aws/load_balancer_delete.go
@@ -7,6 +7,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/elb"
 	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/quintilesims/layer0/common/errors"
 )
 
 // Delete is used to delete an Elastic Load Balancer using the specified loadBalancerID.
@@ -42,6 +43,19 @@ func (l *LoadBalancerProvider) Delete(loadBalancerID string) error {
 		if err := deleteSG(l.AWS.EC2, groupID); err != nil {
 			return err
 		}
+	}
+
+	// Confirm Delete
+	describeFN := func(id string) error {
+		_, err := describeLoadBalancer(l.AWS.ELB, id)
+		if serverError, ok := err.(*errors.ServerError); ok && serverError.Code == "LoadBalancerDoesNotExist" {
+			return nil
+		}
+		return err
+	}
+
+	if err := Retry(fqLoadBalancerID, describeFN); err != nil {
+		return err
 	}
 
 	if err := l.deleteTags(loadBalancerID); err != nil {
--- a/api/provider/aws/load_balancer_delete.go
+++ b/api/provider/aws/load_balancer_delete.go
@@ -7,6 +7,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/elb"
 	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/quintilesims/layer0/common/config"
 	"github.com/quintilesims/layer0/common/errors"
 )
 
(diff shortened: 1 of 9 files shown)


api/provider/aws/load_balancer_delete.go
```diff
@@ -44,6 +45,19 @@ func (l *LoadBalancerProvider) Delete(loadBalancerID string) error {
 		}
 	}
 
+	// Confirm Delete
+	describeFN := func(id string) error {
+		_, err := describeLoadBalancer(l.AWS.ELB, id)
+		if serverError, ok := err.(*errors.ServerError); ok && serverError.Code == "LoadBalancerDoesNotExist" {
```
zpatrick: We need to operate on a lower level: the problem here is the loadbalancer's security group, so we need to make calls to the aws api to verify that the security group is deleted

api/provider/aws/retry.go
```diff
@@ -0,0 +1,24 @@
+package aws
+
+import (
+	"time"
+
+	"github.com/quintilesims/layer0/common/config"
+	"github.com/quintilesims/layer0/common/errors"
+)
+
+func Retry(id string, describeFN func(id string) error) error {
```
zpatrick: I'd expect a generic retry function to have a signature like:
```
func retry(timeout, delay time.Duration, fn func() error) error   
-- or possibly --
func retry(timeout, delay time.Duration, fn func() error) (bool, error)
```

The `id string` pass through in this current impl doesn't make a lot of sense to me, and the lack of ability to change the timeout and delay params seems like this would be difficult to test (unless maybe you used a `timeMultiplier`?). We will also need a clear way to determine when to retry and when to error out (hence the two signatures) above).

api/provider/aws/retry.go
```diff
@@ -0,0 +1,24 @@
...
+	timeout := time.After(config.DefaultRetryTimeOut)
+	tick := time.Tick(config.DefaultRetryWaitTime)
+
+	for {
+		select {
+		case <-timeout:
```
zpatrick: [nit] can just use `<-time.After(timeout)` here. Also, in the examples I've seen for golang timeouts, the timeout case is at the bottom of the select statement - similar to how the `default` is usually at the bottom as well.

common/config/defaults.go
```diff
@@ -15,6 +15,8 @@ const (
 	DefaultEnvironmentType         = models.EnvironmentTypeDynamic
 	DefaultEnvironmentOS           = "linux"
 	DefaultServiceScale            = 1
+	DefaultRetryTimeOut            = time.Second * 60
```
zpatrick: `TimeOut` should be `Timeout`

common/errors/code.go
```diff
@@ -16,4 +16,5 @@ const (
 	ServiceDoesNotExist      ErrorCode = "ServiceDoesNotExist"
 	TaskDoesNotExist         ErrorCode = "TaskDoesNotExist"
 	UnexpectedError          ErrorCode = "UnexpectedError"
+	TimeOut                  ErrorCode = "TimeOut"
```
zpatrick: [nit] I'm a little bit on the fence here; a `Timeout` error code does accurately reflect what went wrong, but, given we are going to have the `client` retry any timeout it encounters, I'm worried someone may use this error code on a function that should not be retried? Maybe I'm being overly cautious? Curious what you think.

api/provider/aws/load_balancer_delete.go
jparsons04: I would add that additional items need to be checked. Keep in mind that when you create a load balancer you also create an IAM Role and a Role Policy. We should probably check that these items are removed as well.

api/provider/aws/load_balancer_delete.go
```diff
@@ -44,6 +46,51 @@ func (l *LoadBalancerProvider) Delete(loadBalancerID string) error {
...
 
+	// Confirm Delete
+	consistencyCheckFN := func() error {
+		// Check IAM Role
+		getRoleInput := &iam.GetRoleInput{}
+		getRoleInput.SetRoleName(roleName)
+
+		if _, err := l.AWS.IAM.GetRole(getRoleInput); err != nil {
+			if err, ok := err.(awserr.Error); ok && err.Code() == "NoSuchEntity" {
+				return nil
```
zpatrick: this is returning too early; we want all checks to pass

api/provider/aws/load_balancer_delete.go
```diff
@@ -44,6 +46,51 @@ func (l *LoadBalancerProvider) Delete(loadBalancerID string) error {
 		}
 	}
 
+	// Confirm Delete
+	consistencyCheckFN := func() error {
+		// Check IAM Role
+		getRoleInput := &iam.GetRoleInput{}
+		getRoleInput.SetRoleName(roleName)
+
+		if _, err := l.AWS.IAM.GetRole(getRoleInput); err != nil {
```
zpatrick: we should follow existing patterns for aws api calls

api/provider/aws/load_balancer_delete.go
```diff
@@ -44,6 +46,51 @@ func (l *LoadBalancerProvider) Delete(loadBalancerID string) error {
...
+		}
+
+		// Check Role Policy
+		getRolePolicyInput := &iam.GetRolePolicyInput{}
+		getRolePolicyInput.SetRoleName(roleName)
+		getRolePolicyInput.SetPolicyName(policyName)
+
+		if _, err := l.AWS.IAM.GetRolePolicy(getRolePolicyInput); err != nil {
+			if err, ok := err.(awserr.Error); ok && err.Code() == "NoSuchEntity" {
+				return nil
```
zpatrick: ditto

api/provider/aws/load_balancer_delete.go
```diff
@@ -44,6 +46,51 @@ func (l *LoadBalancerProvider) Delete(loadBalancerID string) error {
...
+			}
+
+			return err
+		}
+
+		// Check Security Group
+		_, err := readSG(l.AWS.EC2, securityGroupName)
```
zpatrick: `if _, err := ...`

api/provider/aws/load_balancer_delete.go
```diff
@@ -44,6 +46,51 @@ func (l *LoadBalancerProvider) Delete(loadBalancerID string) error {
...
+		if err != nil && !strings.Contains(err.Error(), "does not exist") {
+			return err
+		}
+
+		_, err = describeLoadBalancer(l.AWS.ELB, fqLoadBalancerID)
```
zpatrick: `if _, err := ...`

api/provider/aws/load_balancer_delete.go
```diff
@@ -44,6 +46,51 @@ func (l *LoadBalancerProvider) Delete(loadBalancerID string) error {
...
+		if serverError, ok := err.(*errors.ServerError); ok && serverError.Code == "LoadBalancerDoesNotExist" {
+			return nil
+		}
+
+		return err
+	}
+
+	if err := retry(config.DefaultRetryTimeout, config.DefaultRetryTick, consistencyCheckFN); err != nil {
```
zpatrick: how do we know when to retry and when there is a fatal error?

api/provider/aws/retry.go
```diff
@@ -0,0 +1,22 @@
+package aws
+
+import (
+	"time"
+
+	"github.com/quintilesims/layer0/common/errors"
+)
+
+func retry(timeout, tick time.Duration, fn func() error) error {
+	after := time.After(timeout)
```
zpatrick: this is a memory leak; there's no mechanism to stop the ticker's goroutine.

api/provider/aws/retry.go
```diff
@@ -0,0 +1,22 @@
...
+
+	for {
+		select {
+		case <-time.Tick(tick):
+			if err := fn(); err == nil {
```
zpatrick: this just swallows errors

common/config/defaults.go
```diff
@@ -15,6 +15,8 @@ const (
 	DefaultEnvironmentType         = models.EnvironmentTypeDynamic
 	DefaultEnvironmentOS           = "linux"
 	DefaultServiceScale            = 1
+	DefaultRetryTimeout            = time.Second * 60
```
zpatrick: this probably isn't required; the different api calls should each use retry and sleep parameters suited for that particular call; e.g. the calls for determining if the security group is deleted should be much slower than something like checking the loadbalancer is deleted.

api/provider/aws/load_balancer_delete.go
nehayward: What determines a fatal error? Should it be with 1 or more errors?

api/provider/aws/retry.go
nehayward: Can you elaborate on this or point me to somewhere I can better understand.

api/provider/aws/load_balancer_delete.go
zpatrick: fatal probably isn't the best word: how do we determine when to retry and when to not retry in the face of errors?

api/provider/aws/retry.go
zpatrick: (I put my first comment in the wrong spot, this is should be on the line `case <-time.Tick(tick)`)

There's a couple things wrong. First, you aren't using `time.Tick` correctly here. You're re-creating a new `time.Ticker` instance each time we go through the `for` loop, when you really only need one.
Second, each `time.Tick` spawns its own goroutine, and you haven't provided a way to terminate those goroutines, hence the memory leak.

zpatrick: Instead of trying to plug these memory leaks, I'd recommend trying a different approach to this problem. We've already established a simple pattern with the `testutils.WaitFor` helper function. I'd recommend starting there.

api/provider/aws/retry_test.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestRetry_Timeout(t *testing.T) {
+	fn := func() error {
+		return fmt.Errorf("Tick")
+	}
+
+	if err := retry(20*time.Millisecond, 5*time.Millisecond, fn); err != nil {
+		assert.Equal(t, err, errors.New(errors.FailedRequestTimeout, nil))
```
zpatrick: why do we assume we should get a timeout in this test? There's nothing blocking in `fn` that would cause a timeout to occur, so it seems like this is either passing because:
a) it's relying on the time it takes for goroutines to spin-up and `fn` to execute to be less than 20ms, which is a very brittle assumption. 
b) or, the logic in `retry` is broken and returns timeout errors when it actually shouldn't.

api/provider/aws/retry_test.go
```diff
@@ -0,0 +1,30 @@
...
+	}
+}
+
+func TestRetry_NoTimeout(t *testing.T) {
+	fn := func() error {
+		return nil
+	}
+
+	if err := retry(20*time.Millisecond, 5*time.Millisecond, fn); err != nil {
```
zpatrick: these no assertion that `fn` was every actually called

api/provider/aws/retry_test.go
```diff
@@ -0,0 +1,30 @@
...
+		t.Fatal(err)
+	}
```
zpatrick: we need a test that ensures retry behavior is handled correctly

api/provider/aws/load_balancer_delete.go
```diff
@@ -44,6 +47,72 @@ func (l *LoadBalancerProvider) Delete(loadBalancerID string) error {
 		}
 	}
```
nehayward: Maybe we need some sort of helper function here.

api/provider/aws/load_balancer_delete.go
```diff
@@ -44,6 +47,72 @@ func (l *LoadBalancerProvider) Delete(loadBalancerID string) error {
...
+			}
+			return err
+		}
+		return nil
+	}
+
+	ch := make(chan error, 4)
+	defer close(ch)
+
+	go retry(10*time.Second, 2*time.Second, ch, getRoleFN)
```
nehayward: Would love advice for cleaning this section up. Is it possible to do an array of functions? Or some helper function.

api/provider/aws/load_balancer_delete.go
```diff
@@ -44,6 +47,72 @@ func (l *LoadBalancerProvider) Delete(loadBalancerID string) error {
...
+
+	var errstrings []string
+	for i := 0; i < 4; i++ {
+		if err := <-ch; err != nil {
+			errstrings = append(errstrings, err.Error())
+		}
+	}
+
+	if len(errstrings) > 0 {
+		return errors.New(errors.FailedRequestTimeout, fmt.Errorf(strings.Join(errstrings, "\n")))
```
nehayward: This is one approach, it might be too verbose with all the errors.

api/provider/aws/retry.go
```diff
@@ -0,0 +1,29 @@
...
+	"time"
+
+	"github.com/quintilesims/layer0/common/errors"
+)
+
+func retry(timeout, tick time.Duration, ch chan<- error, fn func() error) {
+	var lastError error
+	after := time.After(timeout)
+	ticker := time.NewTicker(tick)
+	defer ticker.Stop()
```
nehayward: I believe this solves the memory leak.

api/provider/aws/retry_test.go
```diff
@@ -0,0 +1,60 @@
...
+		return nil
+	}
+
+	go retry(20*time.Millisecond, 5*time.Millisecond, ch, fn)
+	if err := <-ch; err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestRetry_GoRoutines(t *testing.T) {
```
nehayward: Not sure if this is needed.

client/load_balancer.go
```diff
@@ -21,8 +22,15 @@ func (c *APIClient) CreateLoadBalancer(req models.CreateLoadBalancerRequest) (st
 
 func (c *APIClient) DeleteLoadBalancer(loadBalancerID string) error {
 	path := fmt.Sprintf("/loadbalancer/%s", loadBalancerID)
-	if err := c.client.Delete(path, nil, nil); err != nil {
-		return err
+	// Retry if Timeout
+	for i := 0; i < 2; i++ {
```
nehayward: Open to suggestion here too. I'm not sure if retry should be a constant here or not.

api/provider/aws/retry_test.go
nehayward: Hmm, my intended goal was to have it retry when there is an error. Since FN only returns an error the function times out before getting a success or nil return.

zpatrick:
closing so we can spend some time working on design

PR #571, opened 2018-03-09
WIP - Adds rudimentary support for Application Load Balancers

sesh-kebab:
Adds a new type of Load Balancer: Application Load Balancer. Default behavior has been kept the same but you can now optionally specify `--type elb|alb` flag (default is elb).

An application load balancer has health check that needs to be specified differently to classic ELB's health check target. So a new property `Path` has been added to HealthCheck model which is used for an ALB type Load Balancer.

When creating a fargate service, you will need to specify a alb load balancer as a classic ELB isn't supported.

* Create a new Applicaiton Load Balancer using `--type alb` flag
* Specify the load balancer when creating a Fargate service
* Ensure that the service is accessible and Target Group health check recognizes a healthy service

**Todo**

- [x] Read SSL certificate information from Target Group Targets (how?)
- [x] Add wait logic when creating/deleting load balancers
- [ ] Update Load Balancer and Service unit tests for ALB cases
- [x] Update Service Read to check if the service is behind an ALB and resolve LB correctly


diemonster:
> When creating a fargate service, you will need to specify a alb load balancer as a classic ELB isn't supported.

Is this set as a default then, or can you still specify a "classic" elb for a fargate service and receive an AWS error?

api/provider/aws/generic_load_balancer.go
```diff
@@ -0,0 +1,33 @@
+package aws
+
+import (
+	"github.com/aws/aws-sdk-go/service/elb"
+	alb "github.com/aws/aws-sdk-go/service/elbv2"
+)
+
+type genericLoadBalancer struct {
```
diemonster: Could this object be simplified? Seems like you could just use typecasting in `GetScheme()`

api/provider/aws/load_balancer_create.go
```diff
@@ -9,14 +9,15 @@ import (
...
+	alb "github.com/aws/aws-sdk-go/service/elbv2"
 	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/quintilesims/layer0/common/config"
 	"github.com/quintilesims/layer0/common/models"
 )
 
-// Create is used to create an Elastic Load Balancer using the specified Create
-// Load Balancer Request. The Create Load Balancer Request contains the name of
-// the Load Balancer, the Environment ID in which to create the Load Balancer,
+// Create is used to create an Elastic or an Application Load Balancer using the
```
diemonster: We should be consistent with calling our previous implementation of ELBs "classic" load balancers, mostly because it get confusing w `elb` versus `elbv2` and all the other name confusion.

api/provider/aws/load_balancer_update.go
```diff
@@ -1,94 +1,135 @@
...
 	"github.com/quintilesims/layer0/common/models"
 )
 
-// Update is used to update an Elastic Load Balancer using the specified Update
-// Load Balancer Request. The Update Load Balancer Request contains the Load
-// Balancer ID, a list of ports to configure as the listeners, and a Health
-// Check to determine if attached EC2 instances are in service or not. If ports
-// are included in the Update Load Balancer Request, all existing listeners and
-// EC2 Security Group ingress rules are removed first and then new listeners and
+// Update is used to update an Classic and Application Load Balancers using the
```
diemonster: [nit] typo: 'an'

setup/module/api/policies/elb_group_policy.json
```diff
@@ -4,7 +4,30 @@
 	{
             "Effect": "Allow",
             "Action": [
-                "elasticloadbalancing:Describe*"
+                "elasticloadbalancing:ModifyListener",
+                "elasticloadbalancing:CreateLoadBalancer",
+                "elasticloadbalancing:RegisterTargets",
+                "elasticloadbalancing:CreateTargetGroup",
+                "elasticloadbalancing:*",
```
diemonster: Do we need `*` when we already reference all these other actions?

cli/command/load_balancer.go
```diff
@@ -84,6 +89,11 @@ func (l *LoadBalancerCommand) Command() cli.Command {
 						Value: dhc.UnhealthyThreshold,
 						Usage: "Number of consecutive failures required to count as unhealthy",
 					},
+					cli.StringFlag{
+						Name:  "type",
+						Value: config.DefaultLoadBalancerType,
+						Usage: "Type of load balancer, either 'ELB' or 'ALB' for a classic elastic or application load balancer (default is 'ELB')",
```
diemonster: either 'ELB' for a classic load balancer or....

api/provider/aws/service_create.go
```diff
@@ -75,21 +75,53 @@ func (s *ServiceProvider) Create(req models.CreateServiceRequest) (string, error
...
-							ContainerPort:    portMapping.ContainerPort,
-							LoadBalancerName: loadBalancerDescription.LoadBalancerName,
+		var assignLoadBalancer = func() error {
+			for _, containerDefinition := range taskDefinition.ContainerDefinitions {
+				for _, portMapping := range containerDefinition.PortMappings {
+					var loadBalancerName *string
+					var targetGroupArn *string
+
+					// todo: check if this verifying this is actually needed
+					// why do we need to verify at least one container exposes a port that the elb has a listener for?
```
diemonster: Because if there are no exposed ports that tie to a listener, then the loadbalancer can't loadbalance anything. I think this was us trying to 'help'.

api/provider/aws/common.go
```diff
@@ -12,29 +12,68 @@ import (
...
-func describeLoadBalancer(elbapi elbiface.ELBAPI, loadBalancerName string) (*elb.LoadBalancerDescription, error) {
-	input := &elb.DescribeLoadBalancersInput{}
-	input.SetLoadBalancerNames([]*string{aws.String(loadBalancerName)})
-	input.SetPageSize(1)
+// searches for the load balancer name as both classic and application load balancers and returns
+// the first found result or an error if the neither classic or application lb could be found for
+// the given lb name
+func describeLoadBalancer(elbapi elbiface.ELBAPI, albapi albiface.ELBV2API, loadBalancerName string) (*genericLoadBalancer, error) {
+	// search classic elastic load balancers
+	searchELB := func() (*genericLoadBalancer, error) {
```
tlake: What do we gain by bundling the ELB search logic into an interior function? I don't think it's used more than once, and I'm not seeing any logic that would be particularly simplified. Maybe the `if err and err != LoadBalancerNotFound` bit, but ALB search just below has that logic as well, and isn't bundled into a function. It just seems strange to me that ELB search and ALB search are so incredibly similar, but are treated pretty differently.

api/provider/aws/generic_load_balancer.go
sesh-kebab: Not sure what you mean? Can you please provide an example?

api/provider/aws/load_balancer_create.go
sesh-kebab: ✅

api/provider/aws/load_balancer_update.go
sesh-kebab: ✅

setup/module/api/policies/elb_group_policy.json
sesh-kebab: ✅

cli/command/load_balancer.go
sesh-kebab: I changed it to "classic or application, load balancer" to be a bit more concise, but if you feel strongly it should be "classic load balancer or application load balancer" I will update it.
✅

api/provider/aws/service_create.go
sesh-kebab: Would you say this level of help is actually needed? Vanilla ECS doesn't enforce this and it is not like we were returning a useful error message when this happened. It looks like we would just silently fail.

api/provider/aws/load_balancer_create.go
```diff
@@ -211,6 +283,41 @@ func (l *LoadBalancerProvider) createLoadBalancer(
...
 
+func (l *LoadBalancerProvider) createApplicationLoadBalancer(
+	loadBalancerName string,
+	scheme string,
+	securityGroupIDs []string,
+	subnetIDs []string,
+) (*alb.LoadBalancer, error) {
+	securityGroups := make([]*string, len(securityGroupIDs))
+	for i, securityGroupID := range securityGroupIDs {
+		securityGroups[i] = aws.String(securityGroupID)
```
tlake: I think you're running into a `range` quirk here that I ran into a couple of weeks ago, where `securityGroups` will end up just being a slice of identical pointers to the final element of `securityGroupIDs`. The value pointer doesn't track all the way through the loop; it just catches the last value.

Example: https://play.golang.org/p/o0F_7HLWf0p

api/provider/aws/load_balancer_create.go
```diff
@@ -211,6 +283,41 @@ func (l *LoadBalancerProvider) createLoadBalancer(
...
+	}
+
+	subnets := make([]*string, len(subnetIDs))
+	for i, subnetID := range subnetIDs {
+		subnets[i] = aws.String(subnetID)
```
tlake: `range` quirk here, too.

api/provider/aws/generic_load_balancer.go
diemonster: My typecasting statement was written in haste, although i'm sure there is a way you could take that approach here as well.

Anyway, it seems redundant to have bools for `isELB` or `isALB` when you could just return an ELB or ALB as part of `describeLoadBalancer`. Then in `GetScheme()`, instead of checking `isELB` or `isALB`, you could just check for `nil`.

sesh-kebab:
@diemonster sorry I missed replying to the below question. You can still specify any load balancer. If you try to specify a classic load balancer when creating a fargate/stateless service, you will get an error stating fargate only supports application load balancers which is surfaced up to the client.

>Is this set as a default then, or can you still specify a "classic" elb for a fargate service and receive an AWS error?

api/provider/aws/generic_load_balancer.go
sesh-kebab: Oh right - yes you're correct. I was in fact doing exactly what you described. But the calling code which looked checked for ELB or an ALB looked cleaner with the bools:

```go
loadBalancer := DescribeLoadBalancer(loadBalancerID)

// without the bools
if loadbalancer.ELB != nil { ... }

// but having isELB & isALB property, it looks more like (which seemed subjectively cleaner)
if loadbalancer.isELB { ... }
```

tlake:
`service get` commands don't populate the LOADBALANCER column for a fargate service deployed behind an ALB. I don't think service read logic has been updated yet.

sesh-kebab:
@tlake right on! will update the todo list in summary.

PR #576, opened 2018-03-14
573: Modify l0-setup to create CloudTrail trail and send events to CloudWatch

### LINKED ISSUE #573: Report CloudTrail errors via Layer0
Layer0's API does a good job of reporting errors that occur due to direct action on the part of the Layer0 API, but it doesn't necessarily capture errors that spawn from those actions. For example, we swallow errors relating to:

1. AWS Account Service Limits (i.e. max ecs clusters reached, max elbs reached)
2. Docker private repo authentication issues

These events _are_ captured in a given AWS account's CloudTrail stream, so we should make those log events viewable by Layer0. Later on, (and out of scope for this issue) we'll make those events actionable. 

We will need to, at minimum:
- via l0-setup, create a mapping between CloudTrail and Cloudwatch so that we can access the log events
- apply filtering on said log events, so that we are pulling only events that are specific to a given Layer0 instance
- allow for those events to be viewable as part of standard layer0 CLI and API actions. 

@jparsons04 if you want to outline your design for the feature in this issue, feel free (especially if you want to discuss). Otherwise, put it in the final PR destined for `api-refactor` 

Tasks:

- [x] `l0-setup` will create a CloudTrail Trail that streams into a CloudWatch log stream within the instance's CloudWatch Log Group
- [x] `l0 environment logs` CLI subcommand - Will display CloudTrail logs for a given Environment
- [x] `l0 admin instancelogs` CLI subcommand - Will display CloudTrail logs for the entire Layer0 instance
- [x] Unit tests for new subcommands

Not in scope and will be designed and worked on in a future issue:
- Passive AWS service limit checks
- Wrap Layer0 actions around actively checking for certain types of events and either erroring or taking corrective action

jparsons04:
Modifies the Terraform structure of l0-setup, specifically the api module, to create the following:

* A CloudTrail trail that will send events to the CloudWatch log group setup creates
* An IAM Role and an inline policy that allows CloudTrail to create log streams and send logs to CloudWatch
* An S3 bucket policy on the bucket that setup creates that lets CloudWatch also send events there

Create a new l0 instance with l0-setup. A CloudTrail trail with the FQDN should be created, and it should send events to a stream in the CloudWatch log group created when the instance is created.

**Checklist**
~- [ ] Unit tests~

closes #573

### DIFF
--- a/setup/module/api/core.tf
+++ b/setup/module/api/core.tf
@@ -22,6 +22,19 @@ resource "aws_s3_bucket" "mod" {
   force_destroy = true
 }
 
+resource "aws_s3_bucket_policy" "cloudtrail" {
+  bucket = "${aws_s3_bucket.mod.id}"
+  policy = "${data.template_file.s3_cloudtrail_bucket_policy.rendered}"
+}
+
+data "template_file" "s3_cloudtrail_bucket_policy" {
+  template = "${file("${path.module}/policies/s3_cloudtrail_bucket_policy.json")}"
+
+  vars {
+    s3_bucket = "${aws_s3_bucket.mod.id}"
+  }
+}
+
 resource "aws_s3_bucket_object" "dockercfg" {
   bucket  = "${aws_s3_bucket.mod.id}"
   key     = "bootstrap/dockercfg"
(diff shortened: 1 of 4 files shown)


sesh-kebab:
I don't quite understand why the S3 portion of this is necessary? Given

> A CloudTrail trail that will send events to the CloudWatch log group setup creates

This seems straight forward.

But re:

> An S3 bucket policy on the bucket that setup creates that lets CloudWatch also send events there

How does S3 come in to play?

jparsons04:
Because an S3 bucket is a required field when creating the CloudTrail trail. See: https://www.terraform.io/docs/providers/aws/r/cloudtrail.html

The S3 bucket is required. Because the S3 bucket is required, a bucket policy has to be attached to the bucket to allow CloudTrail to interact with it.

jparsons04:
An example bucket policy is offered on the same web page in the aws_s3_bucket resource. I found that If I didn't include this bucket policy, terraform apply would error.

jparsons04:
I'm closing this until I get something more feature complete to review.

PR #578, opened 2018-03-20
577: api-refactor, Fix for issue preventing L0 API instances from registering to ECS cluster

### LINKED ISSUE #577: api-refactor: API service of newly created layer0 instances will not register to ECS cluster
### Background

When `l0-setup` is used to create a new Layer0 instance, the API service creates two ECS-optimized EC2 instances that assume an IAM Role that allows the ECS agent on the instance to register itself to the ECS cluster that is created for the API service. The EC2 instance registers itself by way of a [user data script](https://github.com/quintilesims/layer0/blob/api-refactor/setup/module/api/user_data.sh) that is executed once when the instance is launched.

The comments at the very top of the script actually prevent the script from executing in full when the EC2 instances are launched, as was made apparent from ECS agent and cloud-init logs recorded on the EC2 instance post-launch.

### Expected behavior

When creating a new layer0 instance from `api-refactor`, the API service of the instance should come up gracefully -- two EC2 instances should be launched, they should register with the ECS cluster created for the API service, and l0-setup should complete its process without failing.

### Actual behavior

The ECS-optimized EC2 instances that are created for the API service never register with the ECS cluster created.

This is happening because of a formatting issue with the `user_data.sh` script that executes when the EC2 instances are launched.

`l0-setup init` then `l0-setup apply` using `api-refactor` or a child branch as your working branch.

Observe that the EC2 instances are created but never register themselves to your ECS cluster.

jparsons04:

This PR resolves an issue in the `api-refactor` (and children branches) with the ECS container's user data script where the Layer0 API service's EC2 instances would not register with the ECS cluster created for the API service.

Create a new layer0 instance through `l0-setup`. `l0-setup` must complete without failure. The API service's EC2 instances should register correctly with the ECS cluster and the service's task should run without issue. Confirm that the EC2 instances are registering correctly by reviewing the `ECS Instances` tab for the cluster called `l0-<prefix>` where prefix is the name of the layer0 instance specified at `l0-setup` time.

**Checklist**
~- [ ] Unit tests~

closes #577

### DIFF
--- a/setup/module/api/user_data.sh
+++ b/setup/module/api/user_data.sh
@@ -1,12 +1,11 @@
-# The majority of this script was largely taken from AWS's documentation
-# on using CloudWatch logs with container instances
-# https://docs.aws.amazon.com/AmazonECS/latest/developerguide/using_cloudwatch_logs.html
-
 Content-Type: multipart/mixed; boundary="==BOUNDARY=="
 MIME-Version: 1.0
 
 --==BOUNDARY==
 #!/bin/bash
+# The majority of this script was largely taken from AWS's documentation
+# on using CloudWatch logs with container instances
+# https://docs.aws.amazon.com/AmazonECS/latest/developerguide/using_cloudwatch_logs.html
 exec > >(tee /var/log/user-data.log|logger -t user-data -s 2>/dev/console) 2>&1
 echo ECS_CLUSTER=${cluster_id} >> /etc/ecs/ecs.config
 echo ECS_ENGINE_AUTH_TYPE=dockercfg >> /etc/ecs/ecs.config


setup/module/api/user_data.sh
```diff
@@ -1,12 +1,11 @@
-# The majority of this script was largely taken from AWS's documentation
-# on using CloudWatch logs with container instances
-# https://docs.aws.amazon.com/AmazonECS/latest/developerguide/using_cloudwatch_logs.html
-
 Content-Type: multipart/mixed; boundary="==BOUNDARY=="
 MIME-Version: 1.0
 
 --==BOUNDARY==
 #!/bin/bash
+# The majority of this script was largely taken from AWS's documentation
```
zpatrick: Seems like you forgot to add the actual changes to the script(s)?

jparsons04: The change was to move this commented section below the declaration of the multipart/mixed Content-Type. I suppose I could move it above the `--==BOUNDARY==` but the comment should still be relevant for the bash script itself, and thus why I have opted to comment there.

It appeared in my testing that having this commented section above the Content-Type line caused an issue where the script would not execute and would not be able to register itself as an ECS cluster instance.

PR #579, opened 2018-03-23
Add common retry 

nehayward:
This adds a common **retry function** in the common repo within the new **utils** package. This will be used to help with eventual consistency in the future, but also is designed to be used in other scenarios. 

**Detals**
- We chose to use channels, so that we can be as true as possible to the desired timeout param. This is represented in the unit tests. 
- Channels can add unwanted complexity and our goal was to keep it simple and readable, please point out anything that doesn't make sense.

- This can be tested [here](https://play.golang.org/p/7u9lyI0K33j), since there is no current implementation in code.

**Added**
- [x] Unit tests
- [x] Retry function
- [x] Timeout error coded

### DIFF
--- a/common/utils/retry.go
+++ b/common/utils/retry.go
@@ -6,8 +6,8 @@ import (
 	"github.com/quintilesims/layer0/common/errors"
 )
 
-func Retry(timeout, delay time.Duration, fn func() (bool, error)) error {
-	ticker := time.NewTicker(delay)
+func Retry(timeout, interval time.Duration, fn func() (bool, error)) error {
+	ticker := time.NewTicker(interval)
 	errc := make(chan error)
 	defer func() {
 		ticker.Stop()
--- a/common/utils/retry.go
+++ b/common/utils/retry.go
@@ -1,0 +1,40 @@
+package utils
+
+import (
+	"time"
+
+	"github.com/quintilesims/layer0/common/errors"
+)
+
+func Retry(timeout, delay time.Duration, fn func() (bool, error)) error {
+	ticker := time.NewTicker(delay)
+	errc := make(chan error)
+	defer func() {
+		ticker.Stop()
+		close(errc)
+	}()
+
+	go func() { errc <- retry(ticker, fn) }()
+
+	select {
+	case err := <-errc:
+		return err
+	case <-time.After(timeout):
+		return errors.New(errors.FailedRequestTimeout, nil)
+	}
+}
+
+func retry(ticker *time.Ticker, fn func() (bool, error)) error {
+	for ; true; <-ticker.C {
+		shouldContinue, err := fn()
+		if err != nil {
+			return err
+		}
+
+		if !shouldContinue {
+			return err
+		}
+	}
+
+	return nil
+}
--- a/common/utils/retry_test.go
+++ b/common/utils/retry_test.go
@@ -1,0 +1,93 @@
+package utils
+
+import (
+	"math/rand"
+	"testing"
+	"time"
+
+	"github.com/quintilesims/layer0/common/errors"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestRetryTimeout(t *testing.T) {
+	start := time.Now()
+	fn := func() (bool, error) {
+		time.Sleep((time.Duration(rand.Intn(2))) * time.Millisecond)
+		return true, nil
+	}
+
+	if err := Retry(time.Millisecond*1, time.Millisecond*1, fn); err != nil {
+		assert.WithinDuration(t, start.Add(time.Millisecond*1), time.Now(), time.Millisecond)
+	}
+}
+
+func TestRetryNoTimeout(t *testing.T) {
+	retries := 0
+	fn := func() (bool, error) {
+		retries++
+		time.Sleep(1 * time.Millisecond)
+
+		if retries == 5 {
+			return false, nil
+		}
+
+		return true, nil
+	}
+
+	if err := Retry(time.Millisecond*100, time.Millisecond*1, fn); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestRetryFuncTimeRandom(t *testing.T) {
+	start := time.Now()
+	fn := func() (bool, error) {
+		time.Sleep((time.Duration(rand.Intn(2))) * time.Millisecond)
+		return true, nil
+	}
+
+	if err := Retry(time.Millisecond*5, time.Millisecond*1, fn); err != nil {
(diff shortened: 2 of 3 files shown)


common/utils/retry.go
```diff
@@ -0,0 +1,40 @@
+package utils
+
+import (
+	"time"
+
+	"github.com/quintilesims/layer0/common/errors"
+)
+
+func Retry(timeout, delay time.Duration, fn func() (bool, error)) error {
```
diemonster: I like this, very straightforward. Slight nit but how do ppl feel about renaming `delay` to `delayInterval`? Otherwise I read 'delay' as the delay before `Retry()` initially executes, which doesn't imply looping (to me, at least).

nehayward:
@diemonster I have a lot of existing work on AWS api that I can add in, @zpatrick suggested we get this merged by end of the day and then we can reevaluate on Monday?

common/utils/retry.go
diemonster: yeah `interval` works for me

common/utils/retry.go
```diff
@@ -0,0 +1,40 @@
...
+	select {
+	case err := <-errc:
+		return err
+	case <-time.After(timeout):
+		return errors.New(errors.FailedRequestTimeout, nil)
+	}
+}
+
+func retry(ticker *time.Ticker, fn func() (bool, error)) error {
+	for ; true; <-ticker.C {
```
zpatrick: I believe this is a memory leak. I think <-ticker.C will hang forever after the channel is closed.

common/utils/retry_test.go
```diff
@@ -0,0 +1,93 @@
...
+	"testing"
+	"time"
+
+	"github.com/quintilesims/layer0/common/errors"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestRetryTimeout(t *testing.T) {
+	start := time.Now()
+	fn := func() (bool, error) {
```
zpatrick: No need for randomness here; just keep it simple and have this function sleep longer than the timeout

common/utils/retry_test.go
```diff
@@ -0,0 +1,93 @@
...
+		time.Sleep((time.Duration(rand.Intn(2))) * time.Millisecond)
+		return true, nil
+	}
+
+	if err := Retry(time.Millisecond*1, time.Millisecond*1, fn); err != nil {
+		assert.WithinDuration(t, start.Add(time.Millisecond*1), time.Now(), time.Millisecond)
```
zpatrick: this test will falsely succeed if err == nil

common/utils/retry_test.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+}
+
+func TestRetryNoTimeout(t *testing.T) {
```
zpatrick: This behavior doesn't make much sense. Looking at the test itself, it looks like you're actually testing that fn is repeatedly called while you return true, nil

common/utils/retry_test.go
```diff
@@ -0,0 +1,93 @@
...
+	retries := 0
+	fn := func() (bool, error) {
+		retries++
+		time.Sleep(1 * time.Millisecond)
```
zpatrick: not sure why you need to sleep here

common/utils/retry_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+func TestRetryFuncTimeRandom(t *testing.T) {
+	start := time.Now()
+	fn := func() (bool, error) {
+		time.Sleep((time.Duration(rand.Intn(2))) * time.Millisecond)
+		return true, nil
+	}
+
+	if err := Retry(time.Millisecond*5, time.Millisecond*1, fn); err != nil {
+		assert.WithinDuration(t, time.Now(), start.Add(time.Millisecond*5), time.Millisecond)
```
zpatrick: this doesn't execute if err == nil, and this assertion actually isn't affected at all by the rest of the test

common/utils/retry_test.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+}
+
+func TestRetryWaitTimeLongerThanTimeout(t *testing.T) {
```
zpatrick: This is testing the same behavior as TestRetryTimeout

common/utils/retry_test.go
```diff
@@ -0,0 +1,93 @@
...
+		assert.Equal(t, err, errors.New(errors.FailedRequestTimeout, nil))
+		assert.WithinDuration(t, time.Now(), start.Add(time.Millisecond*5), time.Millisecond)
+	}
+}
+
+func TestRetryCount(t *testing.T) {
+	retries := 0
+	fn := func() (bool, error) {
+		retries++
+		time.Sleep(1 * time.Millisecond)
```
zpatrick: why sleep here?

common/utils/retry_test.go
```diff
@@ -0,0 +1,93 @@
...
+		return true, nil
+	}
+
+	if err := Retry(time.Millisecond*5, time.Millisecond*1, fn); err != nil {
```
zpatrick: again, doesn't execute if err == nil, and this test is doing assertions for both the timeout and retry behavior; I'm not sure why we need both

common/utils/retry_test.go
```diff
@@ -0,0 +1,93 @@
...
+		assert.Equal(t, retries, 5)
+		assert.Equal(t, err, errors.New(errors.FailedRequestTimeout, nil))
+	}
+}
+
+func TestRetryFNRanBeforeDelay(t *testing.T) {
+	start := time.Now()
+	fn := func() (bool, error) {
+		assert.WithinDuration(t, time.Now(), start, time.Millisecond)
```
zpatrick: good!

common/utils/retry_test.go
```diff
@@ -0,0 +1,93 @@
...
+		time.Sleep(1 * time.Second)
```
zpatrick: why sleep?

common/utils/retry_test.go
```diff
@@ -0,0 +1,93 @@
...
+		return true, nil
+	}
+
+	Retry(time.Millisecond*5, time.Millisecond*1, fn)
```
zpatrick: [nit] you don't need to multiple by 1

common/utils/retry_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+		return true, nil
+	}
+
+	if err := Retry(time.Millisecond*100, time.Millisecond*1, fn); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestRetryFuncTimeRandom(t *testing.T) {
```
zpatrick: I'm not sure what behavior you're trying to test here?

PR #580, opened 2018-03-23
Adds Application Load Balancer support for stateless Services

### LINKED ISSUE #500: Set LaunchType when creating Services and Tasks

sesh-kebab:
### What does this pull request do?

Stateless services or services that use a fargate launch type do not support Classic Load Balancers. So support for Application LBs is required to host stateless services behind a load balancer.

Extends the existing Layer0 Load Balancer by adding a new type: Application Load Balancer. Default behavior has been changed to create an application LB instead of a classic LB, as we now create a stateless service by default. A classic LB can be created using an optional flag `--type clb|alb` flag (default is alb).

### How should this be tested?

**Prequisites:**
* Fargate compatible layer0 deploy. The deploy sepcfically will require: 
   * Need to set "FARGATE" as a Compatibility in `requiresCompatibilities:['FARGATE']`
   * Need to set the `networkMode:'awsvpc'`
* The docker image needs to be hosted on a public repository as fargate does not currently support private docker repository authentication. Recommend using `quintilesims/guestbook`.

**Default Path:**
1. Create a new Applicaiton Load Balancer, which is the default type of load balancer that is created. You can also use the optional `--type alb(default) | clb` flag.
2. Specify the load balancer when creating a Stateless service (also the default)
3. Ensure that the service is accessible and Target Group health check recognizes a healthy service

```bash
l0 environment create demo
l0 loadbalancer create demo alb1
l0 deploy create ./dockerrun.json guestbook-dpl
l0 service create demo guestbook guestbook-dpl --loadbalancer alb1 
```

**Terraform:**
Create a configuration that matches the above CLI commands apply the configuraiton:
```yaml
# data resources needed for existing layer0 environment & deploy

# loadbalancer type is alb by default
resource "layer0_load_balancer" "alb" {
  name        = "tftestalb"
  environment = "${data.layer0_environment.env.id}"
  private     = false

  port {
    host_port      = 80
    container_port = 80
    protocol       = "HTTP"
  }

  health_check {
    path                = "/"
    interval            = 30
    timeout             = 5
    healthy_threshold   = 2
    unhealthy_threshold = 2
  }
}

# service is stateless (fargate) by default
resource "layer0_service" "guestbook" {
  name          = "guestbook-test"
  environment   = "${data.layer0_environment.env.id}"
  deploy        = "${data.layer0_deploy.dpl.id}"
  load_balancer = "${layer0_load_balancer.alb.id}"
  scale         = 3
}

output "load_balancer_url" {
  value = "${layer0_load_balancer.alb.url}"
}
```

**Test Existing Path for Stateful services:**
Create a stateful service behind a classic load balancer to ensure there are no regressions.

**Test subtle differences in the UX:**
Ensure Load Balancer type is correctly output when queried from the L0 cli.

Test differences between ALB and CLB listeners and health checks.

* Note that an ALB uses `Path=/` where as a CLB uses a `Target=TCP:80`.
* LoadBalancer Add/Drop ports works different with Classic vs Application LBs. Classic has a HostPort and a Container Port where as Application _only_ has a Port which is essentially the HostPort.

Please feel free to propose changes to the UX for adding and removing ALB listeners.

**Checklist**
- [x] Unit tests

partially closes #500


tlake:
@seshi I'm getting an error when trying to create a classic loadbalancer:

```
l0 loadbalancer create --type clb demo clb
2018/03/26 13:00:39 classic is not a supported load balancer type
```

sesh-kebab:
@tlake fixed the issue in the latest. Type flag validation was case sensitive `--type CLB`.

plugins/terraform/layer0/resource_layer0_load_balancer.go
```diff
@@ -34,6 +34,12 @@ func resourceLayer0LoadBalancer() *schema.Resource {
 				Required: true,
 				ForceNew: true,
 			},
+			"load_balancer_type": {
```
tlake: [nit] I wonder if "type" is better than the verbose "load_balancer_type" - since we declare it in a `layer0_load_balancer` already, it seems a little much to have to restate that part.

sesh-kebab: I agree that it is verbose. It somewhat stemmed from my implementation of environment types, where type was already taken so I had to create an `environment_type` property. I don't feel strongly either way on this to change it though, so let me know if you consider this is something I definitely need to change and I will update it accordingly.

zpatrick:
@seshi: this work item needs unit, smoke, system tests and documentation updates

api/provider/aws/generic_load_balancer.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+type genericLoadBalancer struct {
+	CLB   *elb.LoadBalancerDescription
+	ALB   *alb.LoadBalancer
+	isALB bool
+	isCLB bool
+}
+
+func (c genericLoadBalancer) Scheme() *string {
```
zpatrick: why deal with pointers here? Seems to make life more difficult for the callers

api/provider/aws/generic_load_balancer.go
```diff
@@ -0,0 +1,51 @@
...
+		return &genericLoadBalancer{
+			ALB:   ALB,
+			isALB: true,
+		}
+	}
+
+	return nil
+}
+
+type genericLoadBalancer struct {
```
zpatrick: [optional] This might be a more scalable approach:
```
type LoadBalancer interface {
    Scheme() string
    DNSName() string
}

type ALB alb.LoadBalancer 

func (a ALB) Scheme() string {
    // this is a name collision, but you get the idea
    return a.Scheme
}

func (a ALB) DNSName() string {
    return a.DNSName
}

type CLB elb.LoadBalancerDescription
...
```

api/provider/aws/load_balancer_create.go
```diff
@@ -90,41 +95,116 @@ func (l *LoadBalancerProvider) Create(req models.CreateLoadBalancerRequest) (str
...
+	if err != nil {
+		return nil, err
 	}
 
-	if err := l.createTags(loadBalancerID, req.LoadBalancerName, req.EnvironmentID); err != nil {
-		return "", err
+	return output.TargetGroups[0], nil
+}
+
+func (l *LoadBalancerProvider) createListener(loadBalancerArn, targetGroupArn *string, ports []models.Port) error {
```
zpatrick: abbreviations are all uppers in go: `Arn -> ARN`

sesh-kebab:
@zpatrick:
 - unit tests have been updated
 - smoke & system tests won't be possible without tlake's changes and the retry logic changes
 - I can update the documentation but my preference was to combine the effort with the fargate service changes' documentation updtes.

api/provider/aws/load_balancer_delete.go
```diff
@@ -44,25 +50,89 @@ func (l *LoadBalancerProvider) Delete(loadBalancerID string) error {
...
+	lb, err := describeLoadBalancer(l.AWS.ELB, l.AWS.ALB, loadBalancerName)
+	if err != nil {
+		if err, ok := err.(awserr.Error); ok && err.Code() == alb.ErrCodeLoadBalancerNotFoundException {
+			return nil
+		}
+
 		return err
 	}
 
+	if lb.isCLB {
```
zpatrick: all of these switches seem to indicate a problem with the initial design; I think a better approach here is to have both ALBs and CLBs implement a common interface, and have unique types for both ALB and CLB that know how to Read(), Delete(), Create(), etc. themselves.

api/provider/aws/service_create.go
```diff
@@ -60,12 +60,18 @@ func (s *ServiceProvider) Create(req models.CreateServiceRequest) (string, error
...
 		fqLoadBalancerID := addLayer0Prefix(s.Config.Instance(), req.LoadBalancerID)
-		loadBalancerDescription, err := describeLoadBalancer(s.AWS.ELB, fqLoadBalancerID)
+		lb, err := describeLoadBalancer(s.AWS.ELB, s.AWS.ALB, fqLoadBalancerID)
 		if err != nil {
 			return "", err
 		}
 
-		if aws.StringValue(loadBalancerDescription.Scheme) == "internet-facing" {
+		if lb.isALB && req.Stateful {
+			errorMsg := "application load balancer with a stateful service is not currently" +
```
zpatrick: [nit] the pattern in this codebase is to use `text` and multi-line assignment, e.g.:
```
text := "application ... "
text += "supported; please ...."
return fmt.Errorf(text)
```

api/provider/aws/service_read.go
```diff
@@ -53,9 +54,31 @@ func (s *ServiceProvider) Read(serviceID string) (*models.Service, error) {
...
+
+		// classic load balancer
+		if loadBalancer.LoadBalancerName != nil {
+			fqLoadBalancerID := aws.StringValue(loadBalancer.LoadBalancerName)
+			loadBalancerID = delLayer0Prefix(s.Config.Instance(), fqLoadBalancerID)
+		}
+
+		// application load balancer
+		if loadBalancerID == "" && loadBalancer.TargetGroupArn != nil {
+			readTargetGroup(s.AWS.ALB, nil, loadBalancer.TargetGroupArn)
```
zpatrick: does this return an error? if so, can't ignore it here.

cli/command/load_balancer.go
```diff
@@ -84,6 +89,11 @@ func (l *LoadBalancerCommand) Command() cli.Command {
 						Value: dhc.UnhealthyThreshold,
 						Usage: "Number of consecutive failures required to count as unhealthy",
 					},
+					cli.StringFlag{
+						Name:  "type",
+						Value: string(config.DefaultLoadBalancerType),
```
zpatrick: Why don't just have `config.DefaultLoadBalancerType` be a `const string`?

cli/command/load_balancer.go
```diff
@@ -223,12 +232,17 @@ func (l *LoadBalancerCommand) create(c *cli.Context) error {
...
 	req := models.CreateLoadBalancerRequest{
 		LoadBalancerName: args["LOAD_BALANCER_NAME"],
+		LoadBalancerType: models.LoadBalancerType(c.String("type")),
 		EnvironmentID:    environmentID,
 		IsPublic:         !c.Bool("private"),
 		Ports:            ports,
 		HealthCheck:      healthCheck,
 	}
 
+	if err := req.Validate(); err != nil {
```
zpatrick: don't do that here; this already happens in the `client` package.

cli/command/load_balancer.go
```diff
@@ -362,6 +377,10 @@ func (l *LoadBalancerCommand) healthcheck(c *cli.Context) error {
 		HealthCheck: &healthCheck,
 	}
 
+	if err := req.Validate(); err != nil {
```
zpatrick: ditto

cli/command/load_balancer_test.go
```diff
@@ -89,6 +92,7 @@ func TestCreateLoadBalancer(t *testing.T) {
 
 	input := "l0 loadbalancer create "
 	input += "--private "
+	input += "--type " + string(config.DefaultLoadBalancerType) + " "
```
zpatrick: It seems like using `config.DefaultLoadBalancerType`  just makes things less readable; why not use string literals for the test here? I don't see a reason to _have_ to use the default loadbalancer type variable?

cli/command/load_balancer_test.go
```diff
@@ -309,8 +313,15 @@ func TestValidateTarget(t *testing.T) {
 	}
 
 	for _, target := range cases {
+		hc := models.HealthCheck{
+			Target:             target,
```
zpatrick: [nit] this is actually an invalid healthcheck by aws's standards.

common/config/api.go
```diff
@@ -143,6 +143,8 @@ func NewContextAPIConfig(c *cli.Context) *ContextAPIConfig {
 }
 
 func (c *ContextAPIConfig) Validate() error {
+	flags := APIFlags()
+
 	requiredVars := []string{
```
zpatrick: why not just convert this from `[]string` to `map[string]string` where the val is the corresponding environment variable?

common/models/load_balancer_type.go
```diff
@@ -0,0 +1,41 @@
+package models
+
+import (
+	"strings"
+)
+
+type LoadBalancerType string
+
+const (
```
zpatrick: This seems like overkill; we've been using `const string` (or const int) for everything else like this up to this point and has seemed to work great.

setup/module/api/policies/elb_group_policy.json
```diff
@@ -4,7 +4,29 @@
 	{
             "Effect": "Allow",
             "Action": [
-                "elasticloadbalancing:Describe*"
+                "elasticloadbalancing:ModifyListener",
```
zpatrick: We should probably just do `"elasticloadbalancing:*"` if it's getting this large.

sesh-kebab: Agreed ✅

cli/command/load_balancer.go
sesh-kebab: ✅

cli/command/load_balancer.go
sesh-kebab: ✅

api/provider/aws/service_read.go
sesh-kebab: ✅ this call isn't needed.

api/provider/aws/service_create.go
sesh-kebab: ✅

api/provider/aws/load_balancer_create.go
sesh-kebab: ✅ 

as a side note - I got a warning about this from the IDE but hesitantly decided to stay consistent with what is aws go sdk provides (Arn)

api/provider/aws/generic_load_balancer.go
sesh-kebab: ✅

common/config/api.go
sesh-kebab: ✅

cli/command/load_balancer_test.go
sesh-kebab: You definitely don't _have_ to use it. But all the usual pros about strongly typed apply here. It makes it harder for LoadBalancerType to be a random string.

cli/command/load_balancer.go
sesh-kebab: LoadBalancerType is
```go
type LoadBalancerType string
```

common/models/load_balancer_type.go
sesh-kebab: It may seem that way but it certainly helps keep the load balancer type logic encapsulated in one place. The only other thing that is similar that I can think of is `EnvironmentOS`. 

The user can specify either `linux` or `windows`. But load balancer type can be `clb` or `alb` and these values are also translated to `classic` and `application` for the cli. So it makes sense to me to encapsulate this in a type rather than a const string and have supporting helper methods scattered elsewhere.

api/provider/aws/load_balancer_delete.go
sesh-kebab: I would've liked to have done that, but a common interface wouldn't be possible due to the types being in different packages. 

My initial design embedded an ALB & CLB in a common load balancer. But this gets tricky when accessing properties; as you don't know which property belongs to an ALB or a CLB.

sesh-kebab: I'm open to a different approach though for sure.

cli/command/load_balancer.go
```diff
@@ -62,7 +62,12 @@ func (l *LoadBalancerCommand) Command() cli.Command {
 					cli.StringFlag{
 						Name:  "healthcheck-target",
 						Value: dhc.Target,
-						Usage: "Health check target in format 'PROTOCOL:PORT' or 'PROTOCOL:PORT/WITH/PATH'",
+						Usage: "Health check target in format 'PROTOCOL:PORT' or 'PROTOCOL:PORT/WITH/PATH', used for load balancer type 'ELB'",
```
jparsons04: I believe you intended to say CLB instead of ELB here.

setup/module/api/policies/elb_group_policy.json
```diff
@@ -1,21 +1,10 @@
...
         {
             "Effect": "Allow",
-            "Action": [
-                "elasticloadbalancing:*"
-            ],
-            "Resource": [
-		"arn:aws:elasticloadbalancing:${region}:${account_id}:loadbalancer/l0-${name}-*"
-	    ]
+            "Action": "elasticloadbalancing:*",
+            "Resource": "*"
```
jparsons04: I'm a little concerned about this change because it opens up the door for the IAM user to manipulate other load balancers in the AWS account when it shouldn't have permission to do so.

cli/command/load_balancer.go
sesh-kebab: ✅

setup/module/api/policies/elb_group_policy.json
sesh-kebab: I do share your concern but the IAM User needs almost all elasticlaodbalancing permissions anyhow where resource=`*`. I'm not really sure what we're trying to prevent here.

jparsons04: Was there something specific that was breaking with the way it was before? It looks like it had star privileges on a narrow band of load balancer resources, but is there something fundamentally different about the ARN with respect to application load balancers?

sesh-kebab: There was nothing breaking with the way it was before. But in response to a review comment, https://github.com/quintilesims/layer0/pull/580/files/b651a4094bf146d4957fa3ff2b84b11cde2f941c#diff-e52a5f1d1c193669b1a7570d8eab8485

I gave all elasticloadbalancing permissions. So it didn't make sense to have 2 statements with the same permissions, one with resource="*" and the other with resource="arn:aws:elasticloadbalancing:${region}:${account_id}:loadbalancer/l0-${name}-*"

jparsons04: I believe the permissions for `resource="*"` was `Describe*` and the permission for the longer ARN was star permissions.

sesh-kebab: It was `Describe*` prior to this pr. After this PR, it was initially a list of 20+ permissions which is now `elasticloadbalancing:*`.

jparsons04: I see. I think overall my comment might be for a discussion for another issue. I think the permissions can be designed in a way that can balance readability and security, but for now I'll just :heavy_check_mark: this.

api/provider/aws/generic_load_balancer.go
sesh-kebab: I still need access to the underlying ELB & ALB types which map to elb.LoadBalancerDescription & alb.LoadBalancer respectively. I suppose I could have something like this if I wanted to go via the interface route but not sure how it would be beneficial

```go
type LoadBalancer interface {
	Scheme() string
	DNSName() string
	ClassicLoadBalancer
	ApplicationLoadBalancer
}

type ClassicLoadBalancer interface {
	ELB() *elb.LoadBalancerDescription
}

type ApplicationLoadBalancer interface {
	ALB() *alb.LoadBalancer
}

```

common/models/load_balancer_type.go
sesh-kebab: Actually I do see your point about this being overkill. I'll remove the concept of `CLB` and `ALB` and just have `classic` and `application` as load balancer types. This would allow us to stay somewhat consistent.

api/provider/aws/generic_load_balancer.go
zpatrick: > I still need access to the underlying ELB & ALB types 

Why? Isn't just for conversion to/from layer0 models/aws models? If so, why not have the conversion(s) in the interface?

cli/command/load_balancer_test.go
zpatrick: > It makes it harder for LoadBalancerType to be a random string

How could it be random if we are the ones setting it in the test?

Either way, consider this a [nit]

PR #586, opened 2018-03-26
Add CloudTrail functionality to Layer0

jparsons04:
This PR implements the solution outlined in issue #573.

`l0-setup` changes
- Creating a new Layer0 instance will now create a Layer0 CloudTrail Trail
- The Trail will be configured to create new LogStreams in the user's CloudWatch Log Group

CLI changes
- New subcommand `l0 environment logs`, read from the CloudTrail CloudWatch log streams to pull events for a given environment. Designed in a similar way to `l0 service logs`
- New subcommand `l0 admin instancelogs`, read from the CloudTrail CloudWatch log streams to pull all events generated by the Layer0 instance. Designed in a similar way to `l0 service logs`

Testing:
Create a new Layer0 instance through `l0-setup`
Create any valid environment and a deploy
Deploy a service using the deploy to the environment

Run the following commands:
- [x] `l0 environment logs [--start YYYY-MM-DD HH:MM | --end YYYY-MM-DD HH:MM | --tail n]`
- [x] `l0 admin instancelogs [--start YYYY-MM-DD HH:MM | --end YYYY-MM-DD HH:MM | --tail n]`

Each command should return JSON blobs directly from the CloudWatch streams pulling events from CloudTrail.

- [x] All unit tests should pass.


api/provider/aws/logging.go
```diff
@@ -16,6 +16,73 @@ const (
 	MAX_DESCRIBE_STREAMS_COUNT = 1000
```
sesh-kebab: I realize this isn't part of your change but can you please update this to use `mixedCasing` if it isn't used outside of this package or `MixedCasing` if it is.

api/provider/aws/logging.go
```diff
@@ -16,6 +16,73 @@ const (
...
+	filterPattern string,
+) ([]models.LogFile, error) {
+	logStreams, err := describeLogStreams(cloudWatchLogsAPI, logGroupName)
+	if err != nil {
+		return nil, err
+	}
+
+	logFiles := []models.LogFile{}
+	for _, logStream := range logStreams {
+		if strings.Contains(aws.StringValue(logStream.LogStreamName), "_CloudTrail_") {
```
sesh-kebab: [nit] Instead of nesting a lot of code in a if block, I think it would be better to have a continue statement and the rest outside the if block. For example:

```go
for _, logStream := range logStreams [
    if !strings.Contains(...) {
        continue
    }
    
    // write the rest of your code here
}
```

api/provider/aws/logging.go
```diff
@@ -16,6 +16,73 @@ const (
...
+				startMS := timeToMilliseconds(start)
+				input.SetStartTime(startMS)
+			}
+
+			if !end.IsZero() {
+				endMS := timeToMilliseconds(end)
+				input.SetEndTime(endMS)
+			}
+
+			var previousToken string
```
sesh-kebab: I would expect `output.NextToken` to be nil and the lastPage to be true if there are no further results. Wouldn't checking for this condition be sufficient so you don't have to keep track of the previousToken?

For example:
```go
eventsFN := func(...) {
    // append code

    return output.NextToken != nil or return lastPage
}
```

setup/module/api/policies/cloudtrail_role_policy.json
```diff
@@ -0,0 +1,15 @@
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Effect": "Allow",
```
sesh-kebab: Why two statements here? Your Resource looks the same so couldn't you just update the `Action` property to an array which has both `logs:CreateLogStream` and `logs:PutLogEvents`?

api/controllers/environment_test.go
```diff
@@ -137,6 +139,53 @@ func TestReadEnvironment(t *testing.T) {
...
+func TestReadEnvironmentLogs(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockEnvironmentProvider := mock_provider.NewMockEnvironmentProvider(ctrl)
+	controller := NewEnvironmentController(mockEnvironmentProvider)
+
+	expected := []models.LogFile{
+		{
+			ContainerName: "alping",
```
sesh-kebab: not that it really matters but I'm guessing you meant have `alpine`

cli/command/environment.go
```diff
@@ -223,6 +243,27 @@ func (e *EnvironmentCommand) list(c *cli.Context) error {
...
+	if err != nil {
+		return err
+	}
+
+	environmentID, err := e.resolveSingleEntityIDHelper("environment", args["ENVIRONMENT_NAME"])
+	if err != nil {
+		return err
+	}
+
+	query := buildLogQueryHelper(c.String("start"), c.String("end"), c.Int("tail"))
```
sesh-kebab: Would it make sense here to potentially save a round trip by validating the `start` & `end` if they were specified?

setup/module/api/policies/cloudtrail_role_policy.json
jparsons04: You're right, I believe this can be condensed to one Allow statement.

api/controllers/environment_test.go
jparsons04: Yep, I will fix this.

api/controllers/admin.go
```diff
@@ -26,6 +30,12 @@ func (a *AdminController) Routes() []*fireball.Route {
 				"GET": a.GetConfig,
 			},
 		},
+		{
+			Path: "/admin/instancelogs",
+			Handlers: fireball.Handlers{
+				"GET": a.readInstanceLogs,
```
zpatrick: Let's keep things consistent and just use `logs` instead of `instanceLogs` everywhere that applies.

api/provider/aws/logging.go
```diff
@@ -16,6 +16,73 @@ const (
 	MAX_DESCRIBE_STREAMS_COUNT = 1000
 )
 
+func GetLogsFromCloudTrail(
```
zpatrick: Keep things DRY: you should wrap/update/integrate with [GetLogsFromTaskARNs](https://github.com/quintilesims/layer0/blob/api-refactor/api/provider/aws/logging.go#L19) instead of using copy+paste for so much of the implementation.

cli/command/admin.go
```diff
@@ -25,6 +25,26 @@ func (a *AdminCommand) Command() cli.Command {
 				Action:    a.debug,
 				ArgsUsage: " ",
 			},
+			{
+				Name:      "instancelogs",
```
zpatrick: ditto, just use `logs`

cli/command/admin.go
```diff
@@ -25,6 +25,26 @@ func (a *AdminCommand) Command() cli.Command {
...
+				Usage:     "get all CloudTrail logs for a Layer0 instance",
+				Action:    a.instancelogs,
+				ArgsUsage: " ",
+				Flags: []cli.Flag{
+					cli.IntFlag{
+						Name:  "tail",
+						Usage: "number of lines from the end to return (default: 0)",
```
zpatrick: The CLI framework already displays defaults, no need to do that here.

cli/command/admin_test.go
```diff
@@ -21,3 +22,28 @@ func TestDebugAdmin(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestReadLayer0InstanceLogs(t *testing.T) {
```
zpatrick: This should follow our existing naming patterns, e.g. with the instanceLog->log conversion, this would be `TestAdminLogs`

cli/command/environment.go
zpatrick: If we implement this, it should be done in the  `client` package in the same way we do request model validation there.

client/admin.go
```diff
@@ -12,3 +15,12 @@ func (c *APIClient) ReadConfig() (*models.APIConfig, error) {
 
 	return config, nil
 }
+
+func (c *APIClient) ReadLayer0InstanceLogs(query url.Values) ([]models.LogFile, error) {
```
zpatrick: Sticking with the existing naming patterns, this would be `ReadAdminLogs()`

setup/module/api/policies/cloudtrail_role_policy.json
jparsons04: :heavy_check_mark:

cli/command/environment.go
jparsons04: I have decided to implement a similar validation solution in the `client` package by using the `controllers.parseLoggingQuery` helper function. I've made it exportable and call it within the `client` package validate.

api/controllers/admin_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"testing"
+	"time"
+
+	"github.com/golang/mock/gomock"
+	"github.com/quintilesims/layer0/api/provider/mock_provider"
+	"github.com/quintilesims/layer0/common/models"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestReadInstanceLogs(t *testing.T) {
```
zpatrick: `TestReadAdminLogs`

client/admin.go
```diff
@@ -12,3 +16,16 @@ func (c *APIClient) ReadConfig() (*models.APIConfig, error) {
 
 	return config, nil
 }
+
+func (c *APIClient) ReadAdminLogs(query url.Values) ([]models.LogFile, error) {
+	if _, _, _, err := controllers.ParseLoggingQuery(query); err != nil {
```
zpatrick: This introduces a dependency between the `client` package on the `controller` package, which should at least be the other way around (or not exist at all). As a quick fix, `ParseLoggingQuery` should exist in this package.

jparsons04: Should we move the function from the client package and then make calls in `controller` thus use `client.ParseLoggingQuery`?

zpatrick: yep!

sesh-kebab:
[nits] don't look like they were addressed but they don't have to be. Will leave it up to you.

jparsons04:
@sesh-kebab: I changed the camel casing of the constant, the `if` logic nit sort of got blended with changes I made off @zpatrick's suggestion on the merged function `GetLogsFromCloudWatch`, but I did incorporate that idea into the function, I did end up changing the logic of the `FilterLogEventsPages` function, and I renamed the test container name to `alpine`. I'm not sure how they are not being reflected in code though.

PR #591, opened 2018-03-29
Add retry logic for lb delete

### LINKED ISSUE #584: Retry Logic for LB Delete
Read operation to verify, instead of delete for error (for verify).  Pay attention to hard dependency, no point threaded implementation.

* Verify LB deleted
* Verify Security Group is deleted
* retry if either fails

nehayward:
Adds retry logic to lb-delete to handle eventual consistency errors.

### Features
- Retry on loadbalancer describe to verify delete
- Retry on security groups list to verify delete

**Checklist**
- [x] Unit tests

closes #584

### DIFF
--- a/api/provider/aws/common.go
+++ b/api/provider/aws/common.go
@@ -280,7 +280,7 @@ func readService(ecsapi ecsiface.ECSAPI, clusterName, serviceID string) (*ecs.Se
 	return output.Services[0], nil
 }
 
-func waitTillSGDeletedFN(ec2api ec2iface.EC2API, securityGroupName string) func() (shouldRetry bool, err error) {
+func waitUntilSGDeletedFN(ec2api ec2iface.EC2API, securityGroupName string) func() (shouldRetry bool, err error) {
 	return func() (shouldRetry bool, err error) {
 		filter := &ec2.Filter{}
 		filter.SetName("group-name")
--- a/api/provider/aws/common.go
+++ b/api/provider/aws/common.go
@@ -280,26 +280,31 @@ func readService(ecsapi ecsiface.ECSAPI, clusterName, serviceID string) (*ecs.Se
 	return output.Services[0], nil
 }
 
-// Check for eventually consistency
-func waitTillSGDeleted(ec2api ec2iface.EC2API, securityGroupName string) (shouldRetry bool, err error) {
-	filter := &ec2.Filter{}
-	filter.SetName("group-name")
-	filter.SetValues([]*string{aws.String(securityGroupName)})
+func waitTillSGDeletedFN(ec2api ec2iface.EC2API, securityGroupName string) func() (shouldRetry bool, err error) {
+	return func() (shouldRetry bool, err error) {
+		filter := &ec2.Filter{}
+		filter.SetName("group-name")
+		filter.SetValues([]*string{aws.String(securityGroupName)})
 
-	input := &ec2.DescribeSecurityGroupsInput{}
-	input.SetFilters([]*ec2.Filter{filter})
+		input := &ec2.DescribeSecurityGroupsInput{}
+		input.SetFilters([]*ec2.Filter{filter})
 
-	output, err := ec2api.DescribeSecurityGroups(input)
-	if err != nil && !strings.Contains(err.Error(), "does not exist") {
-		return false, err
-	}
+		output, err := ec2api.DescribeSecurityGroups(input)
+		if err != nil {
+			if strings.Contains(err.Error(), "does not exist") {
+				return false, nil
+			}
 
-	for _, group := range output.SecurityGroups {
-		if aws.StringValue(group.GroupName) == securityGroupName {
-			log.Printf("[DEBUG] Security Group not deleted, will retry lookup")
-			return true, nil
+			return false, err
 		}
-	}
 
-	return false, nil
+		for _, group := range output.SecurityGroups {
+			if aws.StringValue(group.GroupName) == securityGroupName {
+				log.Printf("[DEBUG] Security Group %s not deleted, will retry lookup", securityGroupName)
+				return true, nil
+			}
+		}
+
+		return false, nil
+	}
 }
--- a/api/provider/aws/common.go
+++ b/api/provider/aws/common.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"fmt"
+	"log"
 	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -277,4 +278,28 @@ func readService(ecsapi ecsiface.ECSAPI, clusterName, serviceID string) (*ecs.Se
 	}
 
 	return output.Services[0], nil
+}
+
+// Check for eventually consistency
+func waitTillSGDeleted(ec2api ec2iface.EC2API, securityGroupName string) (shouldRetry bool, err error) {
+	filter := &ec2.Filter{}
+	filter.SetName("group-name")
+	filter.SetValues([]*string{aws.String(securityGroupName)})
+
+	input := &ec2.DescribeSecurityGroupsInput{}
+	input.SetFilters([]*ec2.Filter{filter})
+
+	output, err := ec2api.DescribeSecurityGroups(input)
+	if err != nil && !strings.Contains(err.Error(), "does not exist") {
+		return false, err
+	}
+
+	for _, group := range output.SecurityGroups {
+		if aws.StringValue(group.GroupName) == securityGroupName {
+			log.Printf("[DEBUG] Security Group not deleted, will retry lookup")
+			return true, nil
+		}
+	}
+
+	return false, nil
 }
--- a/api/provider/aws/load_balancer_delete.go
+++ b/api/provider/aws/load_balancer_delete.go
@@ -1,6 +1,7 @@
 package aws
 
 import (
+	"log"
 	"strings"
 	"time"
 
@@ -45,6 +46,7 @@ func (l *LoadBalancerProvider) Delete(loadBalancerID string) error {
 			return false, nil
 		}
 
+		log.Printf("[DEBUG] Load Balancer not deleted, will retry lookup")
 		return true, nil
 	}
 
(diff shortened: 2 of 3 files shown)


api/provider/aws/load_balancer_delete.go
```diff
@@ -21,6 +26,32 @@ func (l *LoadBalancerProvider) Delete(loadBalancerID string) error {
...
+			}
+
+			return false, err
+		}
+
+		if len(output.LoadBalancerDescriptions) != 1 {
+			return false, nil
+		}
+
+		return true, nil
```
zpatrick: let's add a `log.Printf("[DEBUG] ...` before we do a retry

api/provider/aws/load_balancer_delete.go
```diff
@@ -44,6 +75,33 @@ func (l *LoadBalancerProvider) Delete(loadBalancerID string) error {
 		}
 	}
 
+	// Check for eventually consistency
+	fnReadSG := func() (shouldRetry bool, err error) {
```
zpatrick: we may want to put this into a helper function since environment delete will want to check that its security group got deleted as well.

api/provider/aws/test_aws/load_balancer_delete_test.go
```diff
@@ -106,8 +118,96 @@ func TestLoadBalancerDeleteIdempotence(t *testing.T) {
...
+
+	mockAWS.ELB.EXPECT().
+		DeleteLoadBalancer(deleteLBInput).
+		Return(&elb.DeleteLoadBalancerOutput{}, nil)
+
+	mockAWS.ELB.EXPECT().
+		DescribeLoadBalancers(gomock.Any()).
+		Return(&elb.DescribeLoadBalancersOutput{}, nil)
+
+	deleteRolePolicyInput := &iam.DeleteRolePolicyInput{}
```
zpatrick: We don't care about inputs in this test; just making one comment here even though it applies elsewhere in this test.

api/provider/aws/test_aws/load_balancer_delete_test.go
```diff
@@ -106,8 +118,96 @@ func TestLoadBalancerDeleteIdempotence(t *testing.T) {
...
+
+	deleteLBInput := &elb.DeleteLoadBalancerInput{}
+	deleteLBInput.SetLoadBalancerName("l0-test-lb_id")
+
+	mockAWS.ELB.EXPECT().
+		DeleteLoadBalancer(deleteLBInput).
+		Return(&elb.DeleteLoadBalancerOutput{}, nil)
+
+	mockAWS.ELB.EXPECT().
+		DescribeLoadBalancers(gomock.Any()).
```
zpatrick: this doesn't execute the retry; this should have at least 2 calls wrapped in `gomock.InOrder(...)`

api/provider/aws/test_aws/load_balancer_delete_test.go
```diff
@@ -106,8 +118,96 @@ func TestLoadBalancerDeleteIdempotence(t *testing.T) {
...
+			DescribeSecurityGroups(gomock.Any()).
+			Return(&ec2.DescribeSecurityGroupsOutput{}, nil),
+	)
+
 	target := provider.NewLoadBalancerProvider(mockAWS.Client(), tagStore, mockConfig)
 	if err := target.Delete("lb_id"); err != nil {
 		t.Fatal(err)
 	}
+
+	assert.Len(t, tagStore.Tags(), 0)
```
zpatrick: This is not the behavior we are trying to verify in this test, so this should be removed.

api/provider/aws/load_balancer_delete.go
```diff
@@ -44,6 +77,33 @@ func (l *LoadBalancerProvider) Delete(loadBalancerID string) error {
...
+		input.SetFilters([]*ec2.Filter{filter})
+
+		output, err := l.AWS.EC2.DescribeSecurityGroups(input)
+		if err != nil && !strings.Contains(err.Error(), "does not exist") {
+			return false, err
+		}
+
+		for _, group := range output.SecurityGroups {
+			if aws.StringValue(group.GroupName) == securityGroupName {
+				return true, nil
```
zpatrick: need a log statement here as well

api/provider/aws/test_aws/load_balancer_delete_test.go
```diff
@@ -106,6 +118,95 @@ func TestLoadBalancerDeleteIdempotence(t *testing.T) {
...
+		if err := tagStore.Insert(tag); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	deleteLBInput := &elb.DeleteLoadBalancerInput{}
+	deleteLBInput.SetLoadBalancerName("l0-test-lb_id")
+
+	mockAWS.ELB.EXPECT().
+		DeleteLoadBalancer(deleteLBInput).
```
zpatrick: don't care about input in this test

nehayward:
@sesh-kebab something like that?

api/provider/aws/common.go
```diff
@@ -278,3 +279,27 @@ func readService(ecsapi ecsiface.ECSAPI, clusterName, serviceID string) (*ecs.Se
...
+	input.SetFilters([]*ec2.Filter{filter})
+
+	output, err := ec2api.DescribeSecurityGroups(input)
+	if err != nil && !strings.Contains(err.Error(), "does not exist") {
+		return false, err
+	}
+
+	for _, group := range output.SecurityGroups {
+		if aws.StringValue(group.GroupName) == securityGroupName {
+			log.Printf("[DEBUG] Security Group not deleted, will retry lookup")
```
zpatrick: [nit] display `securityGroupName` here?

api/provider/aws/common.go
```diff
@@ -278,3 +279,27 @@ func readService(ecsapi ecsiface.ECSAPI, clusterName, serviceID string) (*ecs.Se
...
+func waitTillSGDeleted(ec2api ec2iface.EC2API, securityGroupName string) (shouldRetry bool, err error) {
+	filter := &ec2.Filter{}
+	filter.SetName("group-name")
+	filter.SetValues([]*string{aws.String(securityGroupName)})
+
+	input := &ec2.DescribeSecurityGroupsInput{}
+	input.SetFilters([]*ec2.Filter{filter})
+
+	output, err := ec2api.DescribeSecurityGroups(input)
+	if err != nil && !strings.Contains(err.Error(), "does not exist") {
```
zpatrick: I'm worried that in the "pass-through" case  (err == some error with "does not exist" message), `output` will be `nil`. If that happens, line 297 `range output.SecurityGroups` will cause a panic. I think the safer approach here is:
```go
output, err := ec2api.DescribeSecurityGroups(input)
if err != nil {
    if strings.Contains(err.Error(), "does not exist") {
        return false, nil
    }

    return false, err
}
```

api/provider/aws/load_balancer_delete.go
```diff
@@ -21,6 +26,33 @@ func (l *LoadBalancerProvider) Delete(loadBalancerID string) error {
...
+		output, err := l.AWS.ELB.DescribeLoadBalancers(input)
+		if err != nil {
+			if err, ok := err.(awserr.Error); ok && err.Code() == "LoadBalancerNotFound" {
+				return false, nil
+			}
+
+			return false, err
+		}
+
+		if len(output.LoadBalancerDescriptions) != 1 {
```
zpatrick: I think this should be `== 0`; it's unlikely, but possible that the len > 1

api/provider/aws/load_balancer_delete.go
```diff
@@ -44,6 +76,15 @@ func (l *LoadBalancerProvider) Delete(loadBalancerID string) error {
 		}
 	}
 
+	// Check for eventually consistency
+	fnWaitTillSGDeleted := func() (shouldRetry bool, err error) {
```
zpatrick: I think it would be better if `waitTillSGDeleted` actually returned a `func() (shouldRetry bool, err error)`; that would make the callers' lives a little easier, e.g.
```
// in common.go
func waitTillSGDeletedFN(ec2api ec2iface.EC2API, securityGroupName string) func() (shouldRetry bool, err error) {
    return func() (shouldRetry bool, err error) {
        filter := &ec2.Filter{}
        ...
    }
}

// and in this file:
fn := waitTillSGDeletedFN(l.AWS.EC2, securityGroupName)
if err := retry.Retry(fn, retry.WithTimeout(time.Second*30), retry.WithDelay(time.Second)); err != nil {
...
```

api/provider/aws/load_balancer_delete.go
sesh-kebab: This code path can never be reached as you're checking for a `LoadBalancerNotFound` error and returning in line 37. Also the pagesize is set to 1 so len should never in theory be greater than 1. But in practice it'll only ever be 1 if it gets to here, if i'm reading this code correctly so you shouldn't need this.

api/provider/aws/load_balancer_delete.go
```diff
@@ -21,6 +26,29 @@ func (l *LoadBalancerProvider) Delete(loadBalancerID string) error {
 		return err
 	}
 
+	// Check for eventually consistency
+	waitTillLBDeletedFN := func() (shouldRetry bool, err error) {
+		input := &elb.DescribeLoadBalancersInput{}
+		input.SetLoadBalancerNames([]*string{aws.String(fqLoadBalancerID)})
+		input.SetPageSize(1)
+
+		_, err = l.AWS.ELB.DescribeLoadBalancers(input)
```
zpatrick:
```go
if _, err := ...; err != nil {
    ...
}
```

nehayward: my b.

api/provider/aws/common.go
```diff
@@ -278,3 +279,32 @@ func readService(ecsapi ecsiface.ECSAPI, clusterName, serviceID string) (*ecs.Se
 
 	return output.Services[0], nil
 }
+
+func waitTillSGDeletedFN(ec2api ec2iface.EC2API, securityGroupName string) func() (shouldRetry bool, err error) {
```
diemonster: is there a reason we're using "Till" instead of "Until" for this function name?

nehayward: Nope, I can fix that

PR #592, opened 2018-04-02
500b launch types via task definition

tlake:

**Adds support for launch types:**
- Logic is added for Layer0 Deploys to determine whether a task definition is Fargate and/or EC2 compatible.
- The new default path for Layer0 Deploys, Services, and Tasks is a "stateless" (=Fargate) approach.
  - To create a "stateful" Deploy, the task definition / dockerrun file must specify `"requiresCompatibilities":["EC2"],` at the same top level as `"containerDefinitions"`. A task definition which omits the `"requiresCompatibilities"` field entirely will result in Layer0 automatically inserting that field and populating it with `["EC2","FARGATE"]`, in keeping with a "stateless" default deployment path. This may result in an error at the time of Deploy creation if the task definition is otherwise incompatible with Fargate.
  - In order to deploy a "stateful" (=EC2) Service or Task, the user must pass the `--stateful` flag at the time of Service or Task creation.

**Adds ALB support:**
- The new default Layer0 Load Balancer is ALB, in keeping with the new "stateless" default path for Services and Tasks. In order to create a CLB for "stateful" services, the user must pass `--type classic` at the time of Load Balancer creation.

**Adds partial testing coverage:**
- Unit tests have been implemented for launch types and ALBs.
- System tests and smoke tests will be created/updated in a future PR.

# How should this be tested?
**Prequisites:**
- Fargate-compatible Layer0 instance. Some key points for `l0-setup init`:
  - Fargate does not currently support private docker auth, so don't try to set up a private docker repo.
  - **Source:** Use a local path. I use `/Users/tanner/go/src/github.com/quintilesims/layer0//setup/module`; modify it to suit your filesystem.
  - **Version:** I've built and pushed a `tlake500` image that you're welcome to use, or you can build and push your own.
  - **AWS Region:** `us-east-1`
- Fargate compatible Layer0 Deploy. The Deploy specfically will require: 
  - Need to set "FARGATE" as a Compatibility in `requiresCompatibilities: [ "FARGATE" ]`
  - Need to set the `networkMode: "awsvpc"`
- The docker image for a Deploy needs to be hosted on a public repository as fargate does not currently support private docker repository authentication. Recommend using `quintilesims/guestbook`.

**Default Path:**
1. Create a new Application Load Balancer, which is the default type of load balancer that is created. You can also use the optional `--type application (default) | classic` flag.
2. Specify the load balancer when creating a Stateless service (also the default)
3. Ensure that the service is accessible and Target Group health check recognizes a healthy service

```bash
l0 environment create demo
l0 loadbalancer create demo alb1
l0 deploy create ./dockerrun.json guestbook-dpl
l0 service create --loadbalancer alb1 demo guestbook guestbook-dpl 
```

**Terraform:**
Create a configuration that matches the above CLI commands apply the configuraiton:
```yaml
# data resources needed for existing layer0 environment & deploy

# loadbalancer type is application by default
resource "layer0_load_balancer" "alb" {
  name        = "tftestalb"
  environment = "${data.layer0_environment.env.id}"
  private     = false

  port {
    host_port      = 80
    container_port = 80
    protocol       = "HTTP"
  }

  health_check {
    path                = "/"
    interval            = 30
    timeout             = 5
    healthy_threshold   = 2
    unhealthy_threshold = 2
  }
}

# service is stateless (fargate) by default
resource "layer0_service" "guestbook" {
  name          = "guestbook-test"
  environment   = "${data.layer0_environment.env.id}"
  deploy        = "${data.layer0_deploy.dpl.id}"
  load_balancer = "${layer0_load_balancer.alb.id}"
  scale         = 3
}

output "load_balancer_url" {
  value = "${layer0_load_balancer.alb.url}"
}
```

**Test Existing Path for Stateful services:**
Create a stateful service behind a classic load balancer to ensure there are no regressions. You will need to pass `--type classic` for the load balancer and `--stateful` for the service.

**Test subtle differences in the UX:**
Ensure Load Balancer type is correctly output when queried from the L0 cli.

Test differences between ALB and CLB listeners and health checks.

- Note that an ALB uses `Path=/` where as a CLB uses a `Target=TCP:80`.
- LoadBalancer Add/Drop ports works different with Classic vs Application LBs. Classic has a HostPort and a Container Port where as Application _only_ has a Port which is essentially the HostPort.

**Test for appropriate Service/Task creation errors**
This PR introduces some proactive error-handling during Service and Task creation by comparing certain aspects of Deploys, Load Balancers, Services, and Tasks. Here are a couple of tables for the expected results of various combinations:

| SVC Type | LB Type | DPL Type | Expectation |
|---|---|---|---|
| stateless | alb | fargate | 👍 |
| stateless | alb | ec2 | :x: unsupported stateless+ec2 |
| stateless | clb | fargate | :x: unsupported stateless+clb |
| stateless | clb | ec2 | :x: unsupported stateless+clb |
| stateless | --- | fargate | 👍 |
| stateless | --- | ec2 | :x: unsupported stateless+ec2 |
| stateful | alb | fargate | :x: unsupported stateful+alb |
| stateful | alb | ec2 | :x: unsupported stateful+alb |
| stateful | clb | fargate | :x: unsupported clb+fargate |
| stateful | clb | ec2 | 👍 |
| stateful | --- | fargate | 👍 |
| stateful | --- | ec2 | 👍 |

| Task Type | DPL Type | Expectation |
|---|---|---|
| stateless | fargate | 👍 |
| stateless | ec2 | :x: unsupported stateless+ec2 |
| stateful | fargate | 👍 |
| stateful | ec2 | 👍 |

**Checklist**
- [x] Unit tests

This PR contains work towards #500; outstanding work items should just be non-unit tests and documentation, which will be addressed in a subsequent PR.


common/models/create_task_request.go
```diff
@@ -39,11 +40,12 @@ func (c CreateTaskRequest) Definition() swagger.Definition {
 	return swagger.Definition{
 		Type: "object",
 		Properties: map[string]swagger.Property{
-			"task_name":      swagger.NewStringProperty(),
-			"environment_id": swagger.NewStringProperty(),
-			"deploy_id":      swagger.NewStringProperty(),
 			// TODO: Change to Map defintion
```
sesh-kebab: Would it make sense to address this Todo be addressed?

common/config/defaults.go
```diff
@@ -12,25 +12,30 @@ const (
...
 	DefaultEnvironmentInstanceType = "t2.small"
-	DefaultEnvironmentType         = models.EnvironmentTypeDynamic
 	DefaultEnvironmentOS           = "linux"
 	DefaultServiceScale            = 1
+	DefaultLoadBalancerType        = models.ApplicationLoadBalancerType
+	DefaultTargetGroupPort         = 80
+	DefaultTargetGroupProtocol     = "HTTP"
+
+	// https://docs.aws.amazon.com/AmazonECS/latest/developerguide/platform_versions.html
+	DefaultFargatePlatformVersion = "1.0.0"
```
sesh-kebab: Should we be targeting the 1.1.0? It looks like it has nice things that we can make use of. 

```
Fargate Platform Version‐1.1.0
Added support for Amazon ECS Task Metadata Endpoint.
Added support for Health Check.
Added support for Service Discovery
```

I would make a case for using the Latest but this sentence from AWS docs doesn't make any sense
> If you specify LATEST, your tasks use the most current platform version available, which may not be the most recent platform version.

What is the most 'current' vs the 'recent'?

api/provider/aws/common.go
```diff
@@ -13,32 +13,56 @@ import (
...
+		return nil, err
+	}
+
+	elbExists := true
+	elbOutput, err := elbapi.DescribeLoadBalancers(elbInput)
 	if err != nil {
-		if err, ok := err.(awserr.Error); ok && err.Code() == "LoadBalancerNotFound" {
-			return nil, errors.Newf(errors.LoadBalancerDoesNotExist, "LoadBalancer '%s' does not exist", loadBalancerName)
+		if err, ok := err.(awserr.Error); !ok || err.Code() != "LoadBalancerNotFound" {
+			return nil, err
```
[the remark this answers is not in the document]
sesh-kebab: Nice catch. The logic for describe load balancers is it tries to find a CLB and then an ALB (in that order). If a CLB can't be found, we ignore that error and search for an ALB with the same name. If the ALB doesn't exist, we should be passing back the `LoadBalancerDoesNotExist` which we currently aren't.

@tlake as this originally from my pr, i'll fix this issue and push an update to your branch directly.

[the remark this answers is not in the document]
sesh-kebab: ✅

api/provider/aws/common.go
```diff
@@ -13,32 +14,61 @@ import (
...
 
-	if len(output.LoadBalancerDescriptions) != 1 {
-		return nil, errors.Newf(errors.LoadBalancerDoesNotExist, "LoadBalancer '%s' does not exist", loadBalancerName)
+	albOutput, err := albapi.DescribeLoadBalancers(albInput)
+	if err != nil {
+		if err, ok := err.(awserr.Error); !ok || err.Code() == alb.ErrCodeLoadBalancerNotFoundException {
+			return nil, errors.Newf(errors.LoadBalancerDoesNotExist, "LoadBalancer '%s' does not exist", loadBalancerName)
+		}
+
+		return nil, err
```
jparsons04: I understand the flow of the code here, but it seems awkward to execute it in this way. Would it be wrong to make calls to `elbapi.DescribeLoadBalancers(elbInput)` immediately followed by `albapi.DescribeLoadBalancers(albInput)` and then evaluate each of their respective `err`s after they return? 

It seems like you could cut down on the length of this logic, and you only need to check for LoadBalanacerDoesNotExist within one if block.

You would have something like 
```
if albErr != nil || elbErr != nil {
    <evaluate albErr.(awserr.Error) and elbErr.(awserr.Error)>

    return nil, errors.Newf(errors.LoadBalancerDoesNotExist, "LoadBalancer '%s' does not exist", loadBalancerName)
}
```

api/provider/aws/service_create.go
```diff
@@ -8,69 +8,140 @@ import (
...
 	"github.com/quintilesims/layer0/common/models"
 	"github.com/quintilesims/layer0/common/retry"
 )
 
+// Create runs an ECS Service using the specified CreateServiceRequest.
+// The CreateServiceRequest contains the DeployID, the EnvironmentID,
+// the LoadBalancerID (optional), the Scale, the ServiceName, and the ServiceType.
+//
+// The DeployID is used to look up the ARN of the ECS TaskDefinition to run. If a
+// LoadbalancerID is supplied, it will be used in conjunction with the TaskDefinition
```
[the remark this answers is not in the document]

api/provider/aws/task_create.go
```diff
@@ -5,29 +5,63 @@ import (
...
+// Create runs an ECS Task using the specified CreateTaskRequest.
+// The CreateTaskRequest contains the ContainerOverrides, the DeployID, the
+// EnvironmentID, the TaskName, and the TaskType.
+//
+// The Deploy ID is used to look up the ECS TaskDefinition Family and Version
+// of the Task to run. The TaskType parameter is one of "stateful" or "stateless"
+// and indicates which ECS LaunchType the user wishes to use ("EC2" or "FARGATE"
+// respectively).
+//
+// Create does not generate any custom errors of its own, but will bubble up errors
```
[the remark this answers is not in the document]
tlake: It looks like the service create retry logic is swallowing all errors right now:
```
	fn := func() (shouldRetry bool, err error) {
		if err := s.createService(
			cluster,
			serviceName,
			taskDefinitionARN,
			loadBalancerRole,
			networkMode,
			req.Stateful,
			scale,
			subnets,
			securityGroupIDs,
			loadBalancer,
		); err != nil {
			if strings.Contains(err.Error(), "Unable to assume role") {
				log.Printf("[DEBUG] Failed service create, will retry (%v)", err)
				return true, nil
			}

			return false, err
		}

		return false, nil
	}

	if err := retry.Retry(fn, retry.WithTimeout(time.Second*30), retry.WithDelay(time.Second)); err != nil {
		return "", errors.New(errors.EventualConsistencyError, err)
	}
```

In a situation where `fn()` returns `false` (for instance, if we obtain an AWS error because the user is trying to create a stateless service using a stateful deploy), the call is made three times and then an EventualConsistencyError is returned.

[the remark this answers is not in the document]
tlake: | SVC Type | LB Type | DPL Type | Result |
|---|---|---|---|
| stateless | alb | fargate | ✅ |
| stateless | alb | ec2 | AWS `InvalidParameterException`: task def does not support fargate |
| stateless | clb | fargate | Layer0 error: unsupported stateless+clb |
| stateless | clb | ec2 | Layer0 error: unsupported stateless+clb |
| stateless | --- | fargate | ✅ |
| stateless | --- | ec2 |  |
| stateful | alb | fargate | Layer0 error: unsupported stateful+alb |
| stateful | alb | ec2 | Layer0 error: unsupported stateful+alb |
| stateful | clb | fargate | AWS `InvalidParamaterException`: task def networkMode awsvpc incompatible with clb |
| stateful | clb | ec2 | ✅ |
| stateful | --- | fargate |  |
| stateful | --- | ec2 | ✅ |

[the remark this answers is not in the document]

common/models/create_task_request.go
tlake: I'm not sure about this comment; I haven't found `Definitions` that deviate fro this pattern elsewhere in our code, and as far as I can tell from the Swagger docs, there isn't a `"map"` `Type`.  I think it's just string, number, integer, boolean, array and object ([source](https://swagger.io/docs/specification/data-models/data-types/)). My best guess right now is that this TODO has already been accomplished, since `Properties` is already a `map[string]swagger.Property` and was forgotten to be removed.

common/config/defaults.go
tlake: That _is_ confusing. I bet we could update to use `1.1.0`; I had originally been using `LATEST`, but @diemonster cautioned against it because of how volatile the Fargate launch type will likely be.

api/provider/aws/common.go
tlake: I see where you're coming from, @jparsons04, and for my part, I think that here I have a slight preference for finding exit points sooner than later in favor of trying to make fewer API calls. It's a minor thing probably, but your proposal would have us _always_ making two `DescribeLoadBalancers()` calls; the current flow at least gives us the opportunity to sometimes only call it once if the load balancer happens to be a classic one.

I think this was @sesh-kebab's work - maybe he has a preference/reason he could weigh in with?

sesh-kebab: > Would it be wrong to make calls to elbapi.DescribeLoadBalancers(elbInput) immediately followed by albapi.DescribeLoadBalancers(albInput) and then evaluate each of their respective errs after they return?

Short answer would be yes (or what tanner said). We don't want to make an additional api request if we don't have to.

sesh-kebab: ✅

jparsons04: :heavy_check_mark:

tests/smoke/service.bats
```diff
@@ -2,16 +2,17 @@
 
 @test "create" {
   l0 environment create env_name
-  l0 deploy create ./common/Service.Dockerrun.aws.json dpl_name1
-  l0 deploy create ./common/Service.Dockerrun.aws.json dpl_name2
+  l0 deploy create ./common/Service_stateful.Dockerrun.aws.json dpl_name_stateful1
+  l0 deploy create ./common/Service_stateful.Dockerrun.aws.json dpl_name_stateful2
+  l0 deploy create ./common/Service_stateless.Dockerrun.aws.json dpl_name_stateless
   l0 loadbalancer create --port 80:80/http env_name lb_name1
```
sesh-kebab: You'll need the `--type classic` as this will create an `application` type load balancer.

tests/smoke/service.bats
```diff
@@ -2,16 +2,17 @@
...
-  l0 service create env_name svc_name1 dpl_name1
-  l0 service create --loadbalancer lb_name1 --scale 2 env_name svc_name2 dpl_name1
+  l0 service create --stateful --loadbalancer lb_name1 --scale 2 env_name svc_name_stateful dpl_name_stateful1
+  l0 service create env_name svc_name_stateless dpl_name_stateless
```
sesh-kebab: Recommend adding the `:latest` suffix as having more than one version of this deploy will lead to an error.

tests/smoke/service.bats
tlake: ✅

tests/smoke/service.bats
tlake: ✅

tests/smoke/service.bats
```diff
@@ -20,23 +21,24 @@
...
 
 @test "delete" {
-  l0 service delete svc_name1
-  l0 service delete svc_name2
+  l0 service delete svc_name_stateful
+  l0 service delete svc_name_stateless
   l0 loadbalancer delete lb_name1
-  l0 deploy delete dpl_name1
-  l0 deploy delete dpl_name2
+  l0 deploy delete dpl_name_stateful1
```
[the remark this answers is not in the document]
tlake: ✅

PR #593, opened 2018-04-04
Application Load Balancer - smoke tests

sesh-kebab:
Updates smoke tests to test Application & Classic Load Balancers. To get ALB smoke tests to consistently pass, retry logic also had to be introduced around security groups & target groups when creating and deleting load balancers.

Navigate to `./tests/smoke/`
Execute `bats load_balancer.bats`

**Checklist**
- ~[ ] Unit tests~

Addresses the ALB smoke tests part of #500

### DIFF
--- a/api/provider/aws/common.go
+++ b/api/provider/aws/common.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"strings"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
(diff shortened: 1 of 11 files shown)


jparsons04:
I just reproduced this exact same behavior on a different machine to test to see if this was perhaps something I was seeing locally the first time.

For reference, here is the Terraform main.tf file I'm using: https://gist.github.com/jparsons04/2e8ee723c5ebd92a4334581ad3a4a0a4

Here are the Dockerrun.aws.json files I'm using for each Deploy:
Stateless: https://gist.github.com/jparsons04/4288b78b73955d5ddfdeb3c9fc360ab6
Stateful: https://gist.github.com/jparsons04/e7f4e379da8a9c11e16acdb93f989edd

sesh-kebab:
@jparsons04 the above issue should be fixed. The issue was that we no longer had a default for `environment.instance_type` so on each plan, tf thought it needed to set it to an empty string which forces a recreate. Environment being recreated would also force a recreate on all the dependent resources.

jparsons04:
Just tested the new commit and confirm from my end the behavior works correctly now.

PR #603, opened 2018-04-13
- added helper method to smoke tests to wait for env scale

### LINKED ISSUE #602: Task smoke tests require wait logic for the environment scale
### Expected behavior
from tests/smoke running `bats tasks.bats` should pass.

### Actual behavior
It fails because a create environment command is immediately followed by create task command. As we no longer have a scaler, the create command fails due to insufficient cluster capacity to create the task.

run `bats tasks.bata` from the tests/smoke directory.

sesh-kebab:
Updates Test smoke tests to use helper method so that tasks are created after the cluster has the capacity to create tasks

Run the following commands:
`bats tasks.bats`
`bats service.bats`

**Checklist**
- ~[ ] Unit tests~

fixes #602

### DIFF
--- a/tests/smoke/service.bats
+++ b/tests/smoke/service.bats
@@ -27,7 +27,7 @@
 }
 
 @test "update" {
-  l0 service update svc_name_stateful dpl_name_stateful2
+  l0 service update svc_name_stateful dpl_name_stateful2:latest
 }
 
 @test "logs" {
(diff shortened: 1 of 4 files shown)


tests/smoke/service.bats
```diff
@@ -27,7 +27,7 @@
 }
 
 @test "update" {
-  l0 service update svc_name_stateful dpl_name_stateful2
+  l0 service update svc_name_stateful dpl_name_stateful2:latest
```
tlake: Ah, dang - I missed a `:latest`! Thanks for the catch.

✅

PR #613, opened 2018-05-01
612 remove windows support

### LINKED ISSUE #612: Remove Windows support
As part of the `api-refactor` work, it's been discovered that Windows containers no longer work as they were originally written. Since nobody is currently making use of Layer0's Windows support, we'll remove it for now instead of burning person-hours on it. That way, if it ever becomes a desired feature again, we can just undo this one changeset and get back to figuring out how to make them work again.

tlake:
Removes Windows containers support, since it was built upon a beta AWS featureset which has since been greatly changed. It will require some further looking into in order to re-implement the feature, so until this is a high enough priority item to occupy a dev or two, it's quicker to remove the feature here so that Layer0 doesn't contain broken code.

Closes #612


tlake:
With @diemonster's blessing, I'm merging this so we can get to a state where we can shelve the api-refactor work and prevent it from blocking other work.

PR #618, opened 2018-05-11
Add docker repo override feature to layer0

### LINKED ISSUE #540: Private Docker Repo auth when using Fargate launch type
See discussion below:
https://github.com/quintilesims/layer0/pull/525#discussion_r165734205

sesh-kebab:
Adds a new configuration called `docker-repo-override` to l0-setup. This allows the user to specify a mapping of docker image prefixes. For example, the value can be set to: `d.ims.io:12345.dkr.ecr.us-west-2.amazonaws.com`. 

One use case this allows is to allow images to be pulled from an ECR repository without specifying docker credentials.

1. Create a new l0 instance in a repository other than the carbon account (like the chunnel) with the mapping `d.ims.io:<carbon_aws_account_Id>.dkr.ecr.us-west-2.amazonaws.com`
  (Or you can re-use my instance `seshichunnels`)
2. Create a guestbook deploy
  `l0 deploy create <task definition> guestbook-dpl`
  a. Confirm the deploy correctly replaces the image prefix correctly
  `go run main.go -o json deploy get guestbook-dpl | jq '.[0].dockerrun' -r | base64 -d | jq .`
3. Create a new environment and load balancer 
  `l0 create environment demo-env && l0 loadbalancer create --port 80:80/http demo-env guestbook-lb`
4. Create a service that uses the deploy and load balancer created in the previous steps
  `l0 environment create demo-env`
5. Confirm the service is up and running
  `l0 service create --loadbalancer demo-env:guestbook-lb demo-env guestbook-svc guestbook-dpl:latest`

If the above steps worked, it would've meant that instead of pulling the image via a proxy docker repo frontend (like d.ims.io), the image was pulled directly from ECR by the ecs-agent authenticating using the IAM Role permissions. The image `d.ims.io/quintilesims/guestbook` is also hosted on the carbon account's ECR. So the test confirms that cross-account access will also work.

Notes: An image called `d.ims.io/quintilesims/guestbook:latest` already exists.

You can use the below task definition for step 2:
```
    "AWSEBDockerrunVersion": 2,
    "requiresCompatibilities": [ "EC2" ],
    "containerDefinitions": [
        {
            "name": "guestbook",
            "image": "d.ims.io/quintiles/guestbook",
            "essential": true,
            "memory": 128,
            "portMappings": [
                {
                    "hostPort": 80,
                    "containerPort": 80
                }
            ]
        }
    ]
}
```

**Checklist**
- [x] Unit tests

closes #540 
links https://github.com/quintilesims/d.ims.io/pull/58

### DIFF
--- a/common/config/config.go
+++ b/common/config/config.go
@@ -39,6 +39,7 @@ const (
 	TEST_AWS_TAG_DYNAMO_TABLE = "LAYER0_TEST_AWS_TAG_DYNAMO_TABLE"
 	TEST_AWS_JOB_DYNAMO_TABLE = "LAYER0_TEST_AWS_JOB_DYNAMO_TABLE"
 	AWS_TIME_BETWEEN_REQUESTS = "LAYER0_AWS_TIME_BETWEEN_REQUESTS"
+	DOCKER_REPO_OVERRIDE      = "LAYER0_DOCKER_REPO_OVERRIDE"
 )
 
 // defaults
@@ -253,4 +254,25 @@ func ShouldVerifyVersion() bool {
 	}
 
 	return true
+}
+
+func DockerRepoOverride(image string) string {
+	v := get(DOCKER_REPO_OVERRIDE)
+	overrideMap := map[string]string{}
+	for _, s := range strings.Split(v, ",") {
+		if mapping := strings.Split(s, ":"); len(mapping) == 2 {
+			overrideMap[mapping[0]] = mapping[1]
+		}
+	}
+
+	imageParts := strings.Split(image, "/")
+	if len(imageParts) < 2 {
+		return image
+	}
+
+	if override, ok := overrideMap[imageParts[0]]; ok {
+		return override + "/" + strings.Join(imageParts[1:], "/")
+	}
+
+	return image
 }
--- a/setup/command/init.go
+++ b/setup/command/init.go
@@ -19,6 +19,10 @@ func (f *CommandFactory) Init() cli.Command {
 				Usage: "path to docker config.json file",
 			},
 			cli.StringFlag{
+				Name:  "docker-repo-override",
+				Usage: "map to an ecr backed docker repository. (e.g. d.ims.io:<account_id>.dkr.ecr.us-west-2.amazonaws.com)",
+			},
+			cli.StringFlag{
 				Name:  "module-source",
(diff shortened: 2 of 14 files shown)


common/config/config.go
```diff
@@ -254,3 +255,24 @@ func ShouldVerifyVersion() bool {
...
+	overrideMap := map[string]string{}
+	for _, s := range strings.Split(v, ",") {
+		if mapping := strings.Split(s, ":"); len(mapping) == 2 {
+			overrideMap[mapping[0]] = mapping[1]
+		}
+	}
+
+	imageParts := strings.Split(image, "/")
+	if len(imageParts) < 2 {
+		return image
```
tlake: [Nit] We could fail even earlier and avoid a config lookup by moving this block up to the top

zpatrick:
Why wouldn't users just use `12345.dkr.ecr.us-west-2.amazonaws.com` instead of doing the mapping? That just seems like an unnecessary extra step.

sesh-kebab:
The mapping argument is an alternative to using docker config when creating a layer0 instance. So it isn't about a necessary or unecessary step, but giving the user an option in how they want the container instances to pull the images.

common/config/config.go
sesh-kebab: Do you mean the image parts check? I agree. It isn't a case I would expect to fail normally though. More of a mandatory check everytime we use split.

zpatrick:
They can already do that right now, correct? If they use images with the ecr tag in the deploy file?

sesh-kebab:
If they are hosting in carbon, yes.

setup/command/init.go
```diff
@@ -72,6 +76,10 @@ func (f *CommandFactory) Init() cli.Command {
 				overrides[instance.INPUT_AWS_SSH_KEY_PAIR] = v
 			}
 
+			if v := c.String("docker-repo-override"); v != "" {
```
nehayward: I wish this just avoided empty strings.

sesh-kebab:
closing due to inactivity and priority. might be worthwhile re-visting at some point.

diemonster:
I think we should consider merging this provided we have some end user documentation

diemonster:
Old PR, closing for now.

PR #621, opened 2018-06-19
Fix CPU leak caused by `time.Tick`

### LINKED ISSUE #620: API cpu utilization pegged
### Expected behavior

the api service should be performant and not consume all available resources.

### Actual behavior

the l0 api service consumes 100% of available cpu shortly after being started. reboots result in the same behavior. changing the api instance type - from t2.small to t2.xlarge - is helping, but it looks like the same trend will be occurring. 

after we have max cpu usage, the api seems far more likely to return the infamous `EOF` error. resulting in failed environments with incorrect state.

let the api run for a day.

v0.10.8 currently. but we've seen since at least v0.10.5

tlake:

`layer0/common/aws/provider/connection.go` contained code for a ticker intended to provide a rate limit for the `l0-api`. However, this resulted in infinitely-growing CPU usage on the AWS instance because:
- A new ticker would be created every time `connection.go::getConfig()` was called
- Using `time.Tick()` instead of `time.NewTicker()` yielded only the ticker - it did not yield a mechanism by which to terminate the ticker.

This PR replaces `time.Tick()` with `time.NewTicker()` in order to obtain the mechanism by which a ticker can be stopped. The ticker - henceforth referred to as the rate limiter - is initialized with a default `time.Duration` value, and control of the rate limiter is passed to outside packages by way of public functions. `api/main.go` and `runner/main.go` now handle the call to `time.ParseDuration()` so that they can return usefully (instead of causing a panic down the line if `time.NewTicker()` is called with `nil`), and they also contain `defer provider.StopRateLimiter()` to make sure that they don't leave dangling resources.

# How should this be tested?

## 1. Deploy
Deploy a new new Layer0 instance with `l0-setup`.
- Source: `github.com/quintilesims/layer0//setup/module?ref=620-api-cpu-utilization`
- Version: `tlake620`
- Key Pair: `xfra-dev`, or whatever key pair you use (necessary if you want to monitor the instance with SSH and htop)

## 2. Monitor
Use one or both methods below to monitor CPU usage of the instance. You can try hammering it with a lot of tasks (`for n in {1..999} ; do echo iter ${n}: $(l0 task create ENVIRONMENT task${n} DEPLOY) ; done`) or services or something. In any event, the CPU should respond accordingly, but - and here's the key - it should always return to consuming barely anything when idle.

If you were to deploy an API of v0.10.4-v0.10.8 and monitored it the same way, you would see the CPU usage floor of an idle API gradually increasing with time; it would grow faster the more AWS calls it had to make.

### AWS Console
- Find the Layer0 API cluster in ECS
- Metrics tab
- CPUUtilization

### SSH and htop
- Modify the Load Balancer for the Layer0 API Service
    - Add a 22:22/TCP listener
    - Add a rule to the Load Balancer's Security Group which allows SSH (port 22) traffic from anywhere
- Run `ssh -i path/to/key-pair.pem ec2-user@${LAYER0_API_ENDPOINT} -o serveraliveinterval=30`
- Run `sudo yum install -y htop`
- Run `htop`
    - Press `F4` and enter `l0` to filter out processes which aren't things like `l0-api` and `l0-runner`

## Some charts
Three-day CPU usage of Layer0 API in v0.10.8:

Three-day CPU usage of Layer0 API in v0.10.8

Three-day CPU usage of Layer0 API with these changes:

Three-day CPU usage of Layer0 API with these changes

# Checklist

- ~[ ] Unit tests~

_I'm actually not sure how best to test this particular behavior, if there's a good way to test it at all._

# Issues

Closes #620.

### DIFF
--- a/common/aws/provider/connection.go
+++ b/common/aws/provider/connection.go
@@ -2,8 +2,10 @@ package provider
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
 	"github.com/aws/aws-sdk-go/service/cloudformation"
@@ -53,8 +55,20 @@ var regionIsValid = func(region string) (isValid bool) {
 	return
 }
 
+var ticker *time.Ticker
+
+func init() {
+	delay, err := time.ParseDuration(config.AWSTimeBetweenRequests())
+	if err != nil {
+		return
+	}
+
+	ticker = time.NewTicker(delay)
+}
+
 var getConfig = func(credProvider CredProvider, region string) (sess *session.Session, err error) {
 	fmt.Printf("getConfig called with %v and %v\n", credProvider, region)
+	fmt.Printf("ticker: %v\n", ticker)
 	if !regionIsValid(region) {
 		err = fmt.Errorf("Region '%s' is not a valid region!", region)
 		return
@@ -72,6 +86,9 @@ var getConfig = func(credProvider CredProvider, region string) (sess *session.Se
 
 	creds := credentials.NewStaticCredentials(access_key, secret_key, "")
 	sess = session.New(config.GetAWSConfig(creds, config.AWSRegion()))
+	sess.Handlers.Send.PushBack(func(r *request.Request) {
+		<-ticker.C
+	})
 
 	return
 }
--- a/common/aws/provider/connection.go
+++ b/common/aws/provider/connection.go
@@ -2,10 +2,8 @@ package provider
 
 import (
 	"fmt"
-	"time"
 
 	"github.com/aws/aws-sdk-go/aws/credentials"
-	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
 	"github.com/aws/aws-sdk-go/service/cloudformation"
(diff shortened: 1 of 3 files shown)


common/aws/provider/connection.go
```diff
@@ -55,6 +55,17 @@ var regionIsValid = func(region string) (isValid bool) {
 	return
 }
 
+var Ticker *time.Ticker
+
+func init() {
```
diemonster: I'm kind of surprised at how infrequently we use this pattern. This seems like a perfect use case.

common/aws/provider/connection.go
```diff
@@ -55,6 +55,17 @@ var regionIsValid = func(region string) (isValid bool) {
...
+	delay, err := time.ParseDuration(config.AWSTimeBetweenRequests())
```
zpatrick: It would be better to fail in main and give a good error than to return early here and have the program crash later. 

Let's add some methods for other packages to manage this variable, but initialize it with a default.
I also think it is better to use variables describing what this object _does_ as opposed to what this object _is_, e.g.
```go
const DefaultRateLimit = time.XXX

var rateLimiter = time.NewTicker(DefaultRateLimit)

func ResetRateLimiter(rate time.Duration) {
    rateLimiter.Stop()
    rateLimiter = time.NewTicker(rate)
}

func StopRateLimiter() {
    rateLimiter.Stop()
}
```

zpatrick: FWIW, these suggestions are essentially all hacks. The real problem is that each aws api call uses `getConnection` each time, which is totally unnecessary. The "good" solution here is to remove that from each api call, and then allow `main` to pass in a `time.Ticker` as the rate limiter when constructing the aws api provider. However, that's a lot more work for a tiny win.

tlake: If we're giving the ticker a default and managing it from outside, I think I can take the `init()` function back out.

common/aws/provider/connection.go
tlake: It does, although zpatrick's recent suggestions have made `init()` unnecessary again. xD

I'm still glad that I've learned about it, though.

diemonster:
Current error I'm getting while testing:
`ServerError (code=21) Failed to start task: RESOURCE:MEMORY`

This is after about 55 tasks have been submitted in ~3 mins

PR #629, opened 2018-07-31
use home dir for terraform plugin

qayshp:
`~/.terraform.d/plugins` rather than arbitrary install directory

Update the docs page to help better with the terraform plugin installation.

Try the instructions now :)

**Checklist**
- [ ] ~Unit tests~
- [X] Documentation (if applicable)

closes #
No related issue.

### DIFF
--- a/docs-src/docs/reference/terraform-plugin.md
+++ b/docs-src/docs/reference/terraform-plugin.md
@@ -12,7 +12,7 @@ This provider allows users to create, manage, and update Layer0 entities using T
 ## Install
 Download a Layer0 v0.8.4+ [release](/releases).
 The Terraform plugin binary is located in the release zip file as `terraform-provider-layer0`.
-Copy this `terraform-provider-layer0` binary into the same directory as your Terraform binary - and you're done!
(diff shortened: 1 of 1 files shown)


qayshp:
Not sure if this really matters anymore, but, ¯\_(ツ)_/¯

docs-src/docs/reference/terraform-plugin.md
```diff
@@ -12,7 +12,7 @@ This provider allows users to create, manage, and update Layer0 entities using T
 ## Install
 Download a Layer0 v0.8.4+ [release](/releases).
 The Terraform plugin binary is located in the release zip file as `terraform-provider-layer0`.
-Copy this `terraform-provider-layer0` binary into the same directory as your Terraform binary - and you're done!
+Copy this `terraform-provider-layer0` binary into `~/.terraform.d/plugins/` - and you're done!
```
diemonster: The only problem with this approach is that it assumes you have a `~/.terraform.d/plugins` directory. Older versions of terraform also don't consistently honor this directory for plugins.

PR #634, opened 2019-04-09
4887: Maintenance

tlake:
Layer0 v0.10.10 will be a maintenance release as a result of [JIRA issue OCEMCM-4887](https://jiraims.rm.imshealth.com/browse/OCEMCM-4887). It contains updates to use >= v2.0.0 of the AWS terraform module, and enables DynamoDB tables to scale with load.

Successfully run a complete `l0-setup` and confirm that nothing breaks. Confirm that functionality of `l0` has not been broken. Confirm that DynamoDB tables are able to scale under various loads.

### DIFF
--- a/api/backend/ecs/resource_manager.go
+++ b/api/backend/ecs/resource_manager.go
@@ -57,7 +57,7 @@ func (r *ECSResourceManager) GetProviders(environmentID string) ([]*resource.Res
 func (r *ECSResourceManager) getResourceProvider(ecsEnvironmentID id.ECSEnvironmentID, instance *ecs.ContainerInstance) (*resource.ResourceProvider, bool) {
 	instanceID := pstring(instance.Ec2InstanceId)
 	if pstring(instance.Status) != "ACTIVE" {
-		r.logger.Warnf("Instance %d is not in the 'ACTIVE' state", instanceID)
+		r.logger.Warnf("Instance %s is not in the 'ACTIVE' state", instanceID)
 		return nil, false
 	}
 
--- a/setup/module/api/db.tf
+++ b/setup/module/api/db.tf
@@ -30,7 +30,7 @@ resource "aws_dynamodb_table" "jobs" {
 }
 
 resource "aws_appautoscaling_target" "tags_table_read_target" {
-  max_capacity       = 100
+  max_capacity       = 250
   min_capacity       = 5
   resource_id        = "table/${aws_dynamodb_table.tags.name}"
   scalable_dimension = "dynamodb:table:ReadCapacityUnits"
@@ -54,7 +54,7 @@ resource "aws_appautoscaling_policy" "tags_table_read_policy" {
 }
 
 resource "aws_appautoscaling_target" "jobs_table_read_target" {
-  max_capacity       = 100
+  max_capacity       = 250
   min_capacity       = 5
   resource_id        = "table/${aws_dynamodb_table.jobs.name}"
   scalable_dimension = "dynamodb:table:ReadCapacityUnits"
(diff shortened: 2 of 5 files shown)


api/backend/ecs/resource_manager.go
```diff
@@ -57,7 +57,7 @@ func (r *ECSResourceManager) GetProviders(environmentID string) ([]*resource.Res
 func (r *ECSResourceManager) getResourceProvider(ecsEnvironmentID id.ECSEnvironmentID, instance *ecs.ContainerInstance) (*resource.ResourceProvider, bool) {
 	instanceID := pstring(instance.Ec2InstanceId)
 	if pstring(instance.Status) != "ACTIVE" {
-		r.logger.Warnf("Instance %d is not in the 'ACTIVE' state", instanceID)
+		r.logger.Warnf("Instance %s is not in the 'ACTIVE' state", instanceID)
```
diemonster: Is this just a linter fix? What's the intention of this change?

setup/module/api/db.tf
```diff
@@ -1,6 +1,6 @@
 resource "aws_dynamodb_table" "tags" {
   name           = "l0-${var.name}-tags"
-  read_capacity  = 50
+  read_capacity  = 5
```
diemonster: Do we know the cost, in terms of time and latency, as to when autoscaling takes effect? I can see the argument for a lower initial value so that autoscaling gets to do the heavy lifting (and the potential cost savings therein), but we may want to consider keeping this value at `50` if it was deemed to be reasonable to begin with.

api/backend/ecs/resource_manager.go
tlake: I was getting errors when these things mismatched. I'm on Go 1.12; are you? Maybe you could try playing around with it?

```
package main

import "fmt"

func main() {
    fmt.Printf("percent-s and %s\n", "string")
    fmt.Printf("percent-s and %s\n", 2)
    fmt.Printf("percent-d and %d\n", "string")
    fmt.Printf("percent-d and %d\n", 2)
}
```

setup/module/api/db.tf
tlake: Here's an image of read capacity for the tags table. The deploy I used was a 5-min sleep, and the three spike correspond to three different attempts at hammering the api.
1. `for x in {1..100} ; do nohup l0 task create test-env task-$x test-dpl:latest & done`
2. `for x in {1..200} ; do nohup l0 task create test-env task-$x test-dpl:latest & done`
3. `for y in {1..10} ; do for x in {1..100} ; do nohup l0 task create test-env task-$x test-dpl:latest & done ; sleep 20 ; done`

Screen Shot 2019-04-10 at 10 33 18 AM

api/backend/ecs/resource_manager.go
diemonster: yep I get compile errors as well. Did Go 1.12 change its' behavior with `%d`? I guess so?

setup/module/api/db.tf
diemonster: @Cryowatt have you observed were the read capacity is tending to average when autoscaling is enabled?

Cryowatt: Most of our services are duck-curved, so the capacity trends with when our customers are awake. It seems that during low hours we're down at 5 read units (probably a good minimum for auto-scaling). During busy hours our dev environment hits about 200, but tends to move up and down a lot. I think setting the min low and letting the autoscaling handle things should be fine.

diemonster: Thanks @Cryowatt. Your comments suggest that we should up our `max_capacity` read units count to over 200 as well.

api/backend/ecs/resource_manager.go
tlake: Yeah, that's what it seems like to me.

PR #643, opened 2019-08-14
642: Cross-Zone Load Balancing Management

### LINKED ISSUE #642: Configure Cross-Zone Load Balancing for the Classic Load Balancer
### Expected behavior
Either hard-code Cross-Zone Load Balancing to enabled, or allow it to be specified.

### Actual behavior
Defaults to disabled as that is the AWS API default.
> the default for cross-zone load balancing depends on how you create the load balancer. With the API or CLI, cross-zone load balancing is disabled by default. With the AWS Management Console, the option to enable cross-zone load balancing is selected by default. After you create a Classic Load Balancer, you can enable or disable cross-zone load balancing at any time
https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/enable-disable-crosszone-lb.html

tlake:
Introduces a new feature to load balancers that allows configuration of cross-zone load balancing.

Spin up a new Layer0 dev API and:
- use dev client to create load balancers both with and without the `--disable-cross-zone` flags
- use dev client to test usage of the new load balancer subcommand "`cross-zone`":
    - with no args to view
    - with `--enable` flag
    - with `--disable` flag
- use dev Terraform provider to test behavior of the load balancer's `cross_zone` field in a Terraform configuration:
    - omit the field
    - `cross_zone = true`
    - `cross_zone = false`

**Checklist**
- [x] Unit tests

closes #642

_corresponding issue in JIRA: https://jiraims.rm.imshealth.com/browse/OCEMCM-5776_


sesh-kebab:
Nice work and the quick turn around time! @zpatrick would be impressed.

Initial pass over the code looks fine. But doing a more thorough review pass and QA your RC, will take a bit more time given the current workload. Just wanted to let you know this hasn't been forgotten about.

api/logic/load_balancer_logic_test.go
```diff
@@ -285,3 +286,22 @@ func TestUpdateLoadBalancerIdleTimeout(t *testing.T) {
 
 	testutils.AssertEqual(t, received.IdleTimeout, idleTimeout)
 }
+
+func TestUpdateLoadBalancerCrossZone(t *testing.T) {
```
diemonster: is there an error condition that we can test, too?

PR #658, opened 2020-10-14
OCEMCM-9485

yjwx18:
https://jiraims.rm.imshealth.com/browse/OCEMCM-9485


mb-nx (approved):
Change looks ok to keep the src base in-synch with prod-chunnel.

PR #659, opened 2020-12-07
ecs service ARN update ,fixed OCEMCM-9665.

yjwx18:
Include a fix to OCEMCM-9665, update the new ARN format of ECS services

### DIFF
--- a/common/aws/ecs/ecs.go
+++ b/common/aws/ecs/ecs.go
@@ -769,8 +769,8 @@ func (this *ECS) ListClusterServiceNames(clusterName, prefix string) ([]string, 
 	fn := func(output *ecs.ListServicesOutput, lastPage bool) bool {
 		for _, serviceARN := range output.ServiceArns {
 			// sample service ARN:
-			// arn:aws:ecs:us-west-2:856306994068:service/l0-tlakedev-guestbo80d9d
-			serviceName := strings.Split(aws.StringValue(serviceARN), "/")[1]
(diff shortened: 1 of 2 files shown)


mb-nx (approved):
Reviewed the changes covered..
#1 New ARN format change, https://aws.amazon.com/blogs/compute/migrating-your-amazon-ecs-deployment-to-the-new-arn-and-resource-id-format-2/
#2 Using the latest AMI Image

common/aws/ecs/ecs.go
```diff
@@ -769,8 +769,8 @@ func (this *ECS) ListClusterServiceNames(clusterName, prefix string) ([]string,
 	fn := func(output *ecs.ListServicesOutput, lastPage bool) bool {
 		for _, serviceARN := range output.ServiceArns {
 			// sample service ARN:
-			// arn:aws:ecs:us-west-2:856306994068:service/l0-tlakedev-guestbo80d9d
-			serviceName := strings.Split(aws.StringValue(serviceARN), "/")[1]
```
sesh-kebab, line 773: Any particular reason why we made this only forwards compatible? Something like the below should've handled both new and old formats. Primary concern is that this is one additional thing we have to remember to check if scaling isn't working correctly (i.e., whether the layer0 user or account is still opted in).

```golang
temp := strings.Split(aws.StringValue(serviceARN), "/")
serviceName := temp[len(temp)-1]
```

PR #661, opened 2021-03-09
OCEMCM-11036

mb-nx:
Introduce an optional docker-registry configuration for layer0-api, where a custom registry could be configured during the deploy, which will be used to pull the layer0 images. (dockerhub is used as the default docker registry location)

Layer0-api is running OK (using both default and custom docker-registry)
Layer0-jobs are pulling the container images from the intended docker-registry

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,4 +32,4 @@ stresstest:
 systemtest:
 	$(MAKE) -C tests/system test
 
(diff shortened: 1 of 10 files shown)


sesh-kebab (approved):
This is great. Thanks @mb-nx 
🚢 it.

PR #662, opened 2021-03-31
Ocemcm 11302

yjwx18:
1: Updated the layer0 ec2 image to Linux2 latest version as suggested by aws
2: Update the root size from 8gb to 30gb
3: Updated the default image from t2 to t3. 
4: Added the gitlab pipeline trigger for image sync to d.ims.io

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,8 +33,4 @@ deploy:
     tags: true
     
 after_deploy:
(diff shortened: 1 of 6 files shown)


yjwx18:
We don't need this pull request anymore as single drive approach works

